PR #70, opened 2014-10-25
Use platform-specific iterators instead

### LINKED ISSUE #28: Modify events iterators behavior
`wait_events` and `poll_events` should return *permanent* iterators, ie. iterators that can be reused later.
`wait_events` should not block but instead return an iterator that blocks when `next` is called.

tomaka:
- [x] win32
- [x] x11
- [x] cocoa
- [x] android

Close #28

### DIFF
--- a/src/android/mod.rs
+++ b/src/android/mod.rs
@@ -74,6 +74,52 @@ unsafe impl Send for HeadlessContext {}
 #[cfg(feature = "headless")]
 unsafe impl Sync for HeadlessContext {}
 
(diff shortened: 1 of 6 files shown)


tomaka:
Accepting PRs for cocoa and android!

tomaka:
@glennw Let me know when I can merge this.

glennw:
@tomaka I think it's fine to merge now - if anything breaks once merged let me know and I'll fit it asap.

PR #180, opened 2015-01-02
Change the way that event iterators are represented.

TyOverby:
The bulk of this commit is changing instances of Vec to RingBuf which is
optimized for the push_back() / pop_front() strategy that is used
internaly in the event system.

The glutin custom iterators are now just wrappers around the RingBuf
iterator type.  This will bring the running time of iterator traversal from
O(n^2) to O(n) because shifting-on-delete won't be performed.

### DIFF
--- a/src/android/mod.rs
+++ b/src/android/mod.rs
@@ -7,6 +7,8 @@ use events::ElementState::{Pressed, Released};
 use events::Event::{MouseInput, MouseMoved};
 use events::MouseButton::LeftMouseButton;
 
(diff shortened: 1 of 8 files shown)


TyOverby:
This change also fixes the build for me on nightly.

tomaka:
By "the events system", you mean Piston? Glutin doesn't use unshift/pop_front, and apps that use glutin shouldn't do it either.

tomaka:
Thanks

PR #182, opened 2015-01-04
Fixed win32 at least

Connorcpu:


Connorcpu:
Oh, I see we ended up making all of the same changes, never mind then

PR #185, opened 2015-01-05
Depend on git sources of upstream gl, for this git sourced version

mikedilger:
...of glutin:

When packaging, the upstreams can be cargo.io upstreams, but where development is rapid, they probably should remain git upstreams.

Pre rust-1.0, this probably isn't a critical distinction, because rust itself is pushing the breaking changes.  But once rust settles, cargo pacakges will be longer term and depend on other older stable cargo packages, and git code will need to keep up with it's dependencies a bit faster.

... even if I'm wrong about that, this change was necessary for me today to get glutin to compile, because gl-rs on cargo.io is out of date, but it's git sources are fine.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,10 +17,10 @@ window = []
 headless = []
 
 [dependencies.gl_common]
-gl_common = "*"
+git = "https://github.com/bjz/gl-rs"
 
-[build-dependencies]
-gl_generator = "*"
+[build-dependencies.gl_generator]
+git = "https://github.com/bjz/gl-rs"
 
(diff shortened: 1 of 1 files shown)


tomaka:
I usually download the nightlies, fix everything, and publish to crates.io quickly when the nightlies are published.

But recently the regex/serialize crates have been uploaded on crates.io as well, and when they break it prevents me from doing this. To make things worse they are only updated when I'm about to go to sleep.

mikedilger:
Ok.  I dont' know what the "right" way is.   Was just a suggestion.

PR #187, opened 2015-01-08
Update for Rustc

tomaka:

PR #188, opened 2015-01-08
Update to latest Rust nightly

cybergeek94:
Fixes breakages caused by changes to C string handling in Rust. Compiles and tests successfully as of the opening of this PR.


tomaka:
Merged into #187, thanks!

PR #192, opened 2015-01-13
Add mouse cursor support

bjwbell:
Add a new api, window.set_cursor, for setting the cursor. The enum MouseCursor lists the possible cursors.

Only X11 is implemented. On OSX, Android, & Win32 the window.set_cursor function
either does nothing or calls the "unimplemented!" macro.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "glutin"
-version = "0.0.4-pre"
+version = "0.0.5-pre"
 authors = ["tomaka <pierre.krieger1708@gmail.com>"]
 description = "Cross-plaform OpenGL context provider. Important: the crates.io only supports Windows and Linux for the moment."
 keywords = ["windowing", "opengl"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "glutin"
-version = "0.0.5-pre"
+version = "0.0.4-pre"
 authors = ["tomaka <pierre.krieger1708@gmail.com>"]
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -559,6 +615,11 @@ impl Window {
     pub fn set_window_resize_callback(&mut self, callback: Option<fn(uint, uint)>) {
         self.window.set_window_resize_callback(callback);
     }
+
+    /// Modifies the mouse cursor of the window.
```
tomaka: Could you say that this has no effect on Android?

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 
 name = "glutin"
-version = "0.0.4-pre"
+version = "0.0.5-pre"
```
tomaka: The crates.io version of glutin is, let's say, *experimental*. For the moment I only publish a new version when things are seriously broken.

src/lib.rs
bjwbell: Done (I'd like to add android support in a future commit but no guarantees).

Cargo.toml
bjwbell: Done. Things shouldn't be seriously broken (I've only tested on my machine).

src/lib.rs
tomaka: But Android has no cursor? :huh:

tomaka:
Thanks!

src/lib.rs
bjwbell: When editing text I thought that was a cursor (?) but I'm not actually very familiar with Android.

PR #193, opened 2015-01-13
Change uint/int to usize/isize

bjwbell:
From https://github.com/rust-lang/rfcs/pull/544 the types uint/int were
renamed to usize/isize.


tomaka:
Thanks!

PR #197, opened 2015-01-13
x11, android, win: [ui]size -> [ui]32

ozkriff:


ozkriff:
NOTE: have not tested win32, only made it pass appveyor.

ozkriff:
By the way, do we really need u32? May be replace it with i32 (+ few "assert!(n>=0)") for consistency?

tomaka:
I think I prefer separating u32 and i32.

ozkriff:
ok :)

ozkriff:
r?

Issue #176, opened 2014-12-28
Setting OpenGL context version flags has no effect on OS X


PixelPirate:
This issue is not about the fact that glutin currently has no support for setting the OpenGL version on OS X. I was actually on my way to add that support to glutin but encountered an issue which I could not wrap my head around, so I opened this issue in hopes to get some help.

As far as I understand, OS X only supports 3 OpenGL versions, as documented [here](https://developer.apple.com/opengl/capabilities/index.html).
Namely 2.1 (Legacy), 3.3 and 4.1.
When creating an OpenGL context on OS X, you seem to always get an 2.1 context, in order to get one of the newer versions, you need to specify that version in the attributes of your NSOpenGLPixelFormat instance.

glutin initializes it's context and NSOpenGLPixelFormat instance [here](https://github.com/tomaka/glutin/blob/master/src/osx/mod.rs#L265).

The code currently looks like this:
~~~rust
let attributes = [
    NSOpenGLPFADoubleBuffer as uint,
    NSOpenGLPFAClosestPolicy as uint,
    NSOpenGLPFAColorSize as uint, 24,
    NSOpenGLPFAAlphaSize as uint, 8,
    NSOpenGLPFADepthSize as uint, 24,
    NSOpenGLPFAStencilSize as uint, 8,
    0
];
~~~

To get another context version, one needs to either add
~~~rust
NSOpenGLPFAOpenGLProfile as uint, NSOpenGLProfileVersion4_1Core as uint
~~~
or
~~~rust
NSOpenGLPFAOpenGLProfile as uint, NSOpenGLProfileVersion3_2Core as uint
~~~
to the attributes array.

PS: The enum definition for NSOpenGLProfileVersion* is missing in the rust bindings, here are the missing enum definitions:
~~~objc
enum {
    NSOpenGLProfileVersionLegacy    NS_ENUM_AVAILABLE_MAC(10_7)     = 0x1000,   /* choose a Legacy/Pre-OpenGL 3.0 Implementation */
    NSOpenGLProfileVersion3_2Core   NS_ENUM_AVAILABLE_MAC(10_7)     = 0x3200,   /* choose an OpenGL 3.2 Core Implementation      */
    NSOpenGLProfileVersion4_1Core   NS_ENUM_AVAILABLE_MAC(10_10)    = 0x4100    /* choose an OpenGL 4.1 Core Implementation      */
};
~~~

[...]

PR #208, opened 2015-01-18
Fix #177 merge conflicts

kazimuth:
I looked at #176 and tried adding `NSOpenGLPFAOpenGLProfile, NSOpenGLProfileVersion3_2Core` in the relevant spot in `src/osx/mod.rs`, and it worked - I got a 4.1 core context when I ran the examples. So, I fixed #177's merge conflicts, and made rustc stop complaining about a couple changes it had. (I also accidentally committed some files, but I got rid of them too. I totally know how to use git.)

I wanted to add a test for this but I wasn't exactly sure what it would look like.


tomaka:
:+1:

PR #209, opened 2015-01-18
Remove unnecessary inner method in OSX window constructor

bjz:
This made things a little hard to understand


tomaka:
xml-rs seems broken with the latest nightlies.

Thanks!

PR #210, opened 2015-01-18
Fix for changes in winapi

tomaka:


tomaka:
Broken because of xml-rs. Force-merging.

PR #211, opened 2015-01-19
Upstream of various fixes and features (details below)

glennw:
Add missing get_api stubs.
Add hidpi_factor() support for retina displays.
Fix Resize and MouseMoved events to handle retina displays.
Fix inverted y position for MouseMoved events on mac.
Fix initial painting on retina display.

### DIFF
--- a/src/android/mod.rs
+++ b/src/android/mod.rs
@@ -63,6 +63,10 @@ impl HeadlessContext {
     pub fn get_proc_address(&self, _addr: &str) -> *const () {
         unimplemented!()
     }
+
+    pub fn get_api(&self) -> ::Api {
+        ::Api::OpenGl
+    }
 }
 
 #[cfg(feature = "headless")]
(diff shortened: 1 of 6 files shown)


src/android/mod.rs
```diff
@@ -63,6 +63,10 @@ impl HeadlessContext {
     pub fn get_proc_address(&self, _addr: &str) -> *const () {
         unimplemented!()
     }
+
+    pub fn get_api(&self) -> ::Api {
+        ::Api::OpenGl
```
ozkriff: ::Api::OpenGlEs?

glennw: Ooops, fixed!

tomaka:
Does that work transparently for the user?
Does the user need to use the hdipi factor on the mouse positions, dimensions, etc. or is that done automatically?

glennw:
Mouse positions, resize events are in device pixels. Generally this is what you want and it all works transparently. Occasionally you might need to use the hidpi_factor() if you want to convert the pixel values to the virtual window size units.

PR #213, opened 2015-01-19
Iterators returned by wait_events and poll_events are now persistent

### LINKED ISSUE #28: Modify events iterators behavior

tomaka:
Close #28


tomaka:
cc @aepsil0n

PR #214, opened 2015-01-19
Move osx delegate stuff into separate type

bjz:
This ensures that the class pair is only allocated once. We can't currently clean it up due to a missing function in the `rust-cocoa` ffi, but I plan to rectify that soon.

cc. @glennw, @DavidPartouche


tomaka:
Well, feel free to merge this yourself if no-one says a word.

PR #217, opened 2015-01-22
Some GPU/driver combinations have glxCreateContextAttribsARB present, bu...

glennw:
...t it fails with an X error. In this case, catch the X error and fall back to the old method of creating a context.

### DIFF
--- a/src/x11/window/mod.rs
+++ b/src/x11/window/mod.rs
@@ -18,7 +18,7 @@ static THREAD_INIT: Once = ONCE_INIT;
 
 fn x_error_callback(_: *mut ffi::Display, event: *mut ffi::XErrorEvent) -> libc::c_int {
     unsafe {
-        println!("x error code={} major={} minor={}!", (*event).error_code, (*event).request_code, (*event).minor_code);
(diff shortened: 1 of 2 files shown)


src/x11/window/mod.rs
```diff
@@ -16,10 +16,18 @@ mod monitor;
 
 static THREAD_INIT: Once = ONCE_INIT;
 
+fn x_error_callback(_: *mut ffi::Display, event: *mut ffi::XErrorEvent) -> libc::c_int {
+    unsafe {
+        println!("x error code={} major={} minor={}!", (*event).error_code, (*event).request_code, (*event).minor_code);
```
tomaka: Could this be prefixed with `[glutin] `?
If someone uses a crate which depends on another crate which depends on another crate which depends on glutin, it might be useful for him to know where this `x error` comes from.

glennw: Yep, makes sense. I'll update the PR when I can.

glennw:
@tomaka Updated the log message

tomaka:
Thanks!

PR #218, opened 2015-01-23
Disable compiler unstable warnings

bjwbell:
To cleanup the compiler output when building, disable
warnings about using unstable crates/features

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,3 +1,4 @@
+#![allow(unstable)]
 extern crate gl_generator;
 extern crate khronos_api;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,6 @@
 #![feature(unsafe_destructor)]
 #![unstable]
+#![allow(unstable)]
 
 //! The purpose of this library is to provide an OpenGL context on as many
 //!  platforms as possible.


tomaka:
I'd prefer a single `#![allow(unstable)]` in the crate root, is that possible?

camjackson:
Implementation aside, is squelching all the unstable warnings a good idea? Surely you want to stay aware of what's potentially changing leading up to Rust 1.0? And for things that aren't going to change, the warnings will go away once the language has stabilised.

bjwbell:
@tomaka I couldn't figure out how to use a single #![unstable]. There's already one (#![unstable]) at https://github.com/tomaka/glutin/blob/master/src/lib.rs#L2 but it didn't prevent the warnings.

@camjackson none of the unstable warnings that I squelched appeared useful to me. For instance the warning about CString at https://github.com/tomaka/glutin/blob/master/src/x11/window/mod.rs#L667 seems useless to me. I know the code works now and if stops working during a rust upgrade there were already so many unstable warnings that I wouldn't notice anything about that one.

tomaka:
> I couldn't figure out how to use a single #![unstable]

`#![unstable]` means that all the functions and struct of glutin are unstable (that's why there's a yellow line everywhere [in the docs](http://tomaka.github.io/glutin/glutin/index.html)) whereas `#![allow(unstable)]` means that glutin can use unstable items from the std, which would disable the warnings. They are two different things.

@camjackson The problem is that stability warnings happen both for functions that will be in Rust 1.0 but may slightly change, and for functions that won't be in Rust 1.0. Once the beta is released we'll remove this `#[allow(unstable)]` and cleanup the warnings, something that isn't possible right now.

bjwbell:
@tomaka Thanks for the clarification. Fixed.

tomaka:
Thanks!

PR #220, opened 2015-01-23
Implement vsync for x11 and add vsync example

### LINKED ISSUE #80: Implement swap interval for x11

tomaka:
Close #80

### DIFF
--- a/src/x11/window/mod.rs
+++ b/src/x11/window/mod.rs
@@ -298,7 +298,7 @@ impl Window {
 
 
         // creating GL context
-        let context = unsafe {
+        let (context, extra_functions) = unsafe {
             let mut attributes = Vec::new();
 
             if builder.gl_version.is_some() {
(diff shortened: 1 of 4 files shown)


tomaka:
**r?**

Since glutin is being used by Servo, I don't think I should merge my own PRs now.

src/x11/window/mod.rs
```diff
@@ -345,9 +345,52 @@ impl Window {
...
+                    }
+
+                    if swap != 1 {
+                        return Err(OsError(format!("Couldn't setup vsync: expected \
+                                                    interval `1` but got `{}`", swap)));
+                    }
+                }
+
+            // GLX_MESA_swap_control is not official
+            /*} else if extra_functions.SwapIntervalMESA.is_loaded() {
```
andrewrk: why is this one commented out?

andrewrk:
There should be a way to disable vsync too. Maybe `with_vsync` could accept a boolean?

Calling `swap_interval(0)` turns vsync off and it is sometimes desirable.

tomaka:
> There should be a way to disable vsync too. Maybe with_vsync could accept a boolean?

I agree, but changing the public API is tricky because you need to make sure that you don't break win32, x11, os/x and android. And the last two aren't tested on travis.

src/x11/window/mod.rs
tomaka: The `glXSwapIntervalMESA` function is not generated by `gl-rs` because `GLX_MESA_swap_control` is apparently not an official extension.

andrewrk: Should we ask `gl-rs` to add it? Why does it matter if it is "official" or not?

tomaka: It's not listed on this page. `gl-rs` automatically parses the constant and functions from this file: https://github.com/capnm/Khronos_XML_API/blob/master/glx.xml
Since `GLX_MESA_swap_control` is not in the file, gl-rs doesn't know about it.

PR #224, opened 2015-01-24
fix for latest rustc

andrewrk:
here's a rust issue to go with it: https://github.com/rust-lang/rust/issues/21576


andrewrk:
Wait, I think there's a better way to solve this.

andrewrk:
ok. I had missed changing some traits into the new version but now that's done and I think it should be merged.

tomaka:
Thanks!

PR #227, opened 2015-01-25
Implement grabbing of the mouse pointer for X11

aepsil0n:
Contains new methods in the Window API that closely mirror the Xlib API. The methods are left unimplemented for other platforms for now.

Some bikeshedding on the API is probably in order. Also is it okay to deal with the other platforms later?

### DIFF
--- a/src/android/mod.rs
+++ b/src/android/mod.rs
@@ -358,6 +358,10 @@ impl Window {
     pub fn set_cursor(&self, _: MouseCursor) {
     }
 
+    pub fn grab_cursor(&self) -> Result<(), String> { Ok(()) }
+
+    pub fn ungrab_cursor(&self) {}
+
     pub fn hidpi_factor(&self) -> f32 {
         1.0
     }
--- a/examples/grabbing.rs
+++ b/examples/grabbing.rs
@@ -1,0 +1,50 @@
+#[cfg(target_os = "android")]
+#[macro_use]
+extern crate android_glue;
+
+extern crate glutin;
+
+use glutin::{Event, ElementState};
+
+mod support;
+
+#[cfg(target_os = "android")]
+android_start!(main);
+
+#[cfg(not(feature = "window"))]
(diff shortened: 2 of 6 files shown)


src/android/mod.rs
```diff
@@ -290,6 +290,14 @@ impl Window {
     pub fn set_cursor(&self, _: MouseCursor) {
     }
 
+    pub fn grab_cursor(&self) {
+        unimplemented!()
```
tomaka: This shouldn't be `unimplemented!()` but just empty, since there's no effect on android (same 5 lines below).

src/lib.rs
```diff
@@ -636,6 +636,16 @@ impl Window {
         self.window.set_cursor(cursor);
     }
 
+    /// Grabs the mouse cursor.
```
tomaka: Please add that it has no effect on Android.

tomaka:
It's ok not to implement something, but related issues should be opened after the PR is merged.

Since X11 directly provides a function that does this, it makes sense to call it. I don't know what OS/X provides?

tomaka:
Some documentation should be added:

- Is the cursor also hidden?
- Is the cursor centered on the window or does it just no leave the window? (important for events)
- Which effect does this have if the cursor was already grabbed?

aepsil0n:
Uhm… Let me look into the Xlib API to make sure.

tomaka:
It's important to document public APIs so that all platforms have the same behavior.

aepsil0n:
Hm… looking at that, maybe it would be better to return a result from `grab_cursor` instead of panicking, as it may fail for various reasons.

aepsil0n:
I rebased this onto the current master. While I haven't successfully tested this yet, do you think the API is okay in this form?

tomaka:
Yeah, I guess it's OK.

aepsil0n:
So, now that glutin is working again I added a working example. Should be ready now.

tomaka:
Thanks!

PR #229, opened 2015-01-26
{true, false}->{YES, NO}

bjz:


bjz:
I am just going to merge this seeing as it is a simple change.

PR #231, opened 2015-01-26
Implement some more Cocoa stuff

bjz:


tomaka:
Yay the rename \o/

For `get_position`, it's supposed to be the distance between the top-left hand corner of the screen and the top-left hand corner of the window. I guess I should clarify this and check if it's correct for win32/linux.

bjz:
I reckon some documentation regarding glutin's coordinate system would be useful. Seeing as there are cross platform differences for high dpi displays, etc.

bjz:
Ok, so those implementations are wrong, but I'll need to add some more methods to rust-cocoa to be able to fix it.

bjz:
(ie. the position implementations)

tomaka:
:shipit:

tomaka:
I opened #223

bjz:
Bad link?

PR #239, opened 2015-01-29
Fix crash on close in some X11 drivers

glennw:
On some X11 driver implementations, calling XCloseDisplay from a different thread causes memory corruption, resulting in a crash. Change window proxy to be a weak reference so that the XCloseDisplay is always called from the main window.


tomaka:
Can you rebase this on master to make sure that travis passes?

glennw:
Rebased and travis passes now.

tomaka:
Thanks

PR #240, opened 2015-01-29
Add circle.yml file to circle-ci

tomaka:


tomaka:
I don't want to bother requesting an access for something that is a hack :(

PR #246, opened 2015-02-01
android: Set egl::{RED,GREEN,BLUE,DEPTH}_SIZE to 8

ozkriff:
Maybe this should be configurable?


tomaka:
It *is* configurable: https://github.com/tomaka/glutin/blob/master/src/lib.rs#L190-193

Could do you:

```rust
let (red, green, blue) = match builder.color_bits.unwrap_or(24) {
    24 => (8, 8, 8),
    16 => (6, 5, 6),
    _ => panic!()
};
```

ozkriff:
Oh, sorry) I will prepare another PR soon.

PR #247, opened 2015-02-01
Reorder the fields of `XErrorEvent` to match the struct layout found in recent Xlib.h

### LINKED ISSUE #244: Layout of `ffi::XErrorEvent` does not match Xlib.h
On the two Ubuntu-based machines I have access to, the package libx11-dev has the following version information:
```
libx11-dev:arm64       2:1.6.2-1ubuntu2
libx11-dev:amd64       2:1.6.2-1ubuntu2
```

On both systems, Xlib.h contains the following declaration for `XErrorEvent`:

```
typedef struct {
        int type;
        Display *display;       /* Display the event was read from */
        XID resourceid;         /* resource id */
        unsigned long serial;   /* serial number of failed request */
        unsigned char error_code;       /* error code of failed request */
        unsigned char request_code;     /* Major op-code of failed request */
        unsigned char minor_code;       /* Minor op-code of failed request */
} XErrorEvent;
```

Here, the `resourceid` field is in the middle of the struct, while `XErrorEvent` in x11/ffi.rs has it as the last field. This causes that error reporting in an application becomes meaningless (the actual problem I'm having right now is in Servo).

First I though to submit a PR to reorder the fields but then found some pages by googling which document `XErrorEvent` the way it is implemented here. So now I'm not sure whether a reordering of fields would cause trouble for someone else out there, and if so, how to detect the layout on a given system?

(Still, I'm happy to submit a PR if you think that's the way to go.)

akiss77:
Fixes #244

### DIFF
--- a/src/x11/ffi.rs
+++ b/src/x11/ffi.rs
@@ -1362,11 +1362,11 @@ pub struct XF86VidModeModeInfo {
 pub struct XErrorEvent {
     pub type_: libc::c_int,
     pub display: *mut Display,
+    pub resourceid: XID,
     pub serial: libc::c_ulong,
     pub error_code: libc::c_char,
     pub request_code: libc::c_char,
     pub minor_code: libc::c_char,
-    pub resourceid: XID,
 }
 
 #[cfg(feature = "headless")]


tomaka:
Thanks!

PR #250, opened 2015-02-03
Add tempory fix for #248

### LINKED ISSUE #248: Mouse movements leading to crash with cocoa
From https://github.com/tomaka/glium/issues/425#issuecomment-72496265:

> Running the window example from Glutin has the same error, keyboard and clicks work fine, while mouse movement crashes with this crash report: http://pastebin.com/VjV4gWzd

bjz:
Fixes #248


tomaka:
The latest nightlies broke xml-rs :(

Thanks for the fix!

PR #252, opened 2015-02-05
Add some more support for X11 keys (partially fixes #251)

### LINKED ISSUE #251: Glutin is eating numbers and other special characters on linux
https://github.com/servo/servo/issues/4839

Servo isn't being notified of key events for characters other than [A-Za-z]. I think Enter, space, arrow keys, and backspace work, but not much else.

This seems to be an issue with glutin (since we don't do platform-specific stuff when it comes to handling keys)

Manishearth:
I wasn't sure if I could add more variants to the virtual key enum so I left it alone. Most of the remaining ones can be handled by a shift key anyway.


Manishearth:
Not sure why the build failed (master seems to be failing on appveyor anyway) :/

tomaka:
Oh, it's an example that is broken with the latest rustc. Let me fix that so you can rebase over master.

tomaka:
Thanks!

Manishearth:
Btw, should adding new keys to virtualkeys be okay?

Eg for the numpad plus, etc that x11 supports (not sure if cocoa does), or the extra parentheses keys (instead of Shift-9), etc.

tomaka:
Ideally we'd need a reference implementations where all virtuals keys are implemented. Then make the other implementations match this.

Manishearth:
Would GLFW be a good idea? (btw, I'm in the gitter room if that's preferable :) )

PR #254, opened 2015-02-05
Rename enum variants of MouseButton

aepsil0n:
With enums that are namespaced by default, it seems reasonable to be less redundant in the variant naming here.


tomaka:
Thanks

PR #268, opened 2015-02-11
Only build test bindings if we compile tests

tomaka:
Merge one or two days after https://github.com/rust-lang/cargo/pull/1288 is merged


tomaka:
No longer relevant, as the Cargo feature has been modified.

PR #269, opened 2015-02-11
Fix poll_events() blocking on win32

tomaka:


tomaka:
Does not pass because of a breaking change in nightly. Tested locally, merging.

PR #271, opened 2015-02-15
Setup deployment on crates.io

tomaka:
Automatically deploy a version on crates.io stripped down from OS/X and Android


TyOverby:
Don't you also need to increment the version number?

tomaka:
I've modified `0.0.4-pre` to `0.0.4` in this PR. The version `0.0.4` has been uploaded.

TyOverby:
Doesn't this change make it so that it will auto-publish?

tomaka:
Yes, it will autopublish as soon as I change the version number in the Cargo.toml

TyOverby:
Oh, I see.  So you make the version number change manually, and then it automatically publishes?

tomaka:
Yes

PR #285, opened 2015-02-20
expose platform-specific window handle (currently Win only)

ryanstew:
Similar scenario to platform_display, I suspect. In this case, we're working with the Oculus SDK, which requires passing them the HWND of the render window on Windows.

I didn't implement for other platforms because I have no easy way to test. Let me know if I should rough in implementations anyway, or just open issues for the unimplemented stubs for other platforms.

### DIFF
--- a/src/android/mod.rs
+++ b/src/android/mod.rs
@@ -329,6 +329,10 @@ impl Window {
         self.display as *mut libc::c_void
     }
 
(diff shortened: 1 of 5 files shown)


tomaka:
It's not a problem if they are not implemented. Thanks.

PR #298, opened 2015-02-23
Make headless rendering work on Mac OS X 10.10

andreasgal:
The pixel format attributes match what gecko uses for offscreen rendering.


tomaka:
Thanks

PR #301, opened 2015-03-01
Use the WGL API to determine extended pixel format, plus fix creation

tomaka:


tomaka:
Creates non-hardware accelerated contexts, need to fix this.

tomaka:
Appveyor fails because as a work-around I chose to filter out all software-accelerated formats, and appveyor doesn't have any hardware-acceleration.

PR #316, opened 2015-03-16
Cocoa fixes: memory leaks, monitor handling, is_current()

ryanstew:
cc @bjz  

Sorry for the slightly multi-purpose PR. They're inter-related (for me, at least), but I can break this apart further if necessary.

Covers the following fixes:

* Makes sure retain/release is happening as alloc'd or retained `NSObject`s are passed around the cocoa backend
* Exposes an enum that gives raw access to the platform-native monitor identifier. For both external library integration, and for the next item.
* Fix with_fullscreen() behavior on Cocoa to correctly select the passed-in monitor, rather than always using `NSScreen mainScreen`.
* Add missing implementation for Window::is_current() in cocoa

### DIFF
--- a/src/cocoa/mod.rs
+++ b/src/cocoa/mod.rs
@@ -456,10 +456,7 @@ impl Window {
                 NSBackingStoreBuffered,
                 NO,
             ));
-
-            if *window == nil {
-                None
-            } else {
+            window.non_nil().map(|window| {
                 let title = IdRef::new(NSString::alloc(nil).init_str(title));
                 window.setTitle_(*title);
                 window.setAcceptsMouseMovedEvents_(YES);
@@ -469,21 +466,19 @@ impl Window {
                 else {
                     window.center();
                 }
-                Some(window)
-            }
+                window
+            })
         }
     }
 
     fn create_view(window: id) -> Option<IdRef> {
         unsafe {
             let view = IdRef::new(NSView::alloc(nil).init());
-            if *view == nil {
-                None
-            } else {
+            view.non_nil().map(|view| {
                 view.setWantsBestResolutionOpenGLSurface_(YES);
                 window.setContentView_(*view);
-                Some(view)
-            }
+                view
+            })
         }
     }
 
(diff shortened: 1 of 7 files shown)


ryanstew:
Hmm, one thing I still have to detangle: in builds with only headless, NativeMonitorID never gets pulled into the crate, so the type isn't available for implementations of MonitorID in the backends. I can just move the definition of that type to somewhere not conditional on headless. But should the winimpl specific monitor stuff even be built in the headless case?

bjz:
Will review tonight.

src/window.rs
```diff
@@ -500,6 +500,19 @@ pub fn get_primary_monitor() -> MonitorID {
     MonitorID(winimpl::get_primary_monitor())
 }
 
+/// Native platform identifier for a monitor. Different platforms use fundamentally different types
+/// to represent a monitor ID.
+pub enum NativeMonitorID {
```
bjz: Should we derive `PartialEq` and `Eq` for this?

src/cocoa/mod.rs
```diff
@@ -445,20 +474,20 @@ impl Window {
         }
     }
 
-    fn create_view(window: id) -> Option<id> {
+    fn create_view(window: id) -> Option<IdRef> {
         unsafe {
-            let view = NSView::alloc(nil).init();
-            if view == nil {
+            let view = IdRef::new(NSView::alloc(nil).init());
+            if *view == nil {
```
bjz: Would be nice if we had an `IdRef::nil()` to compare against so you wouldn't need to deref.

tomaka:
`NativeMonitorID` should be named `NativeMonitorId`.

I know that there's `MonitorID` in the code already, but that should be changed too eventually.

src/cocoa/mod.rs
ryanstew: Great idea. cf630ec has a slightly different approach, but maybe it's too cute and I should fall back to this more basic mechanism. Interested in feedback.

src/lib.rs
```diff
@@ -322,3 +324,20 @@ impl<'a> BuilderAttribs<'a> {
                       .expect("Could not find compliant pixel format")
     }
 }
+
+mod native_monitor {
```
ryanstew: I don't love this, but it is a minimally invasive way to make this available crate-wide regardless of feature config while avoiding the pub use behind the window feature gate. Open to other suggestions.

tomaka:
Why `mod native_monitor` and `pub use native_monitor::NativeMonitorId` instead of declaring `NativeMonitorId` at the root?

ryanstew:
I didn't want the type to be public in builds without the window feature, since it's useless to the public interface in that case. And unless I'm missing something, it's not possible promote a type to public via `pub use` without renaming it or moving it to another module.

tomaka:
@bjz: Feel free to merge

src/lib.rs
bjz: Why is it needed crate wide?

bjz:
I'm still hesitant about the design of `NativeMonitorId`. If you split out the memory leak fixes I can merge those fine.

src/lib.rs
ryanstew: Because each winimpl implementation has a MonitorID implementaiton that must return it. All of these are building regardless of which features are selected for the crate. So conditioning this type definition on the "window" feature causes build issues in the --no-default-features case.

Unless you're asking why I need to make it public, and not just limit it to the cocoa implementation. This is (for my use case) an important part of this change, as I have code outside of glutin that needs access to this identifier, across multiple platforms. This is why it's exposed as an enum rather than just hiding it in cocoa's winimpl as a u32.

ryanstew:
Happy to split out stuff if anyone is anxious to see a particular part of this change in place.

I'll clarify my motivation for this PR: I dove into this code because I need the ability to find a specific monitor and open a fullscreen window on it. Both sides of this were broken: the ability to find the monitor by ID and the ability to correctly open the window on it. Totally open to other ways of doing this, but I'm hoping whatever we settle on allows me to do both things (on Windows, OS X, and Linux).

bjz:
If @tomaka is fine, then I am fine.

PR #321, opened 2015-03-22
Use objc crate for Objective-C runtime functionality

SSheldon:
I've been working on an [objc crate](https://crates.io/crates/objc) for interoperating with the Objective-C runtime. glutin is one of the most prominent crates using Objective-C from Rust, so I'm very interested in getting your opinion of the objc crate and hearing if you feel like it would be useful to glutin.

This pull request demonstrates how using the objc crate in glutin would look. This change requires the cocoa crate to use objc, though, so it'd require servo/rust-cocoa#71 as a prerequisite.

I've tested this with the cocoa change and all the examples still work, but I'm not too familiar with glutin, so let me know if there is more thorough testing I can do.

Let me know if the objc crate looks like it could be useful or if you have any suggestions for me!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ clock_ticks = "*"
 version = "0"
 
 [target.x86_64-apple-darwin.dependencies]
+objc = "0.1"
 glutin_cocoa = "0"
 glutin_core_graphics = "0"
 glutin_core_foundation = "0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,23 +30,10 @@ clock_ticks = "*"
 [target.arm-linux-androideabi.dependencies.android_glue]
 version = "0"
 
-[target.i686-apple-darwin.dependencies.glutin_cocoa]
-version = "0"
-
-[target.x86_64-apple-darwin.dependencies.glutin_cocoa]
-version = "0"
-
-[target.i686-apple-darwin.dependencies.glutin_core_graphics]
-version = "0"
-
-[target.x86_64-apple-darwin.dependencies.glutin_core_graphics]
-version = "0"
-
-[target.i686-apple-darwin.dependencies.glutin_core_foundation]
-version = "0"
-
-[target.x86_64-apple-darwin.dependencies.glutin_core_foundation]
-version = "0"
+[target.x86_64-apple-darwin.dependencies]
+glutin_cocoa = "0"
+glutin_core_graphics = "0"
+glutin_core_foundation = "0"
 
 [target.i686-pc-windows-gnu.dependencies]
 winapi = "0.1"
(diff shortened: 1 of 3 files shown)


ryanstew:
Very nice.

#316 introduced a variant of your Id type, where previously this code was leaking all retained ids. Your wrappers are better, but this change will probably need to be rebased against my changes and the Id type swapped in for my IdRef.

bjz:
Ok, rust-cocoa now uses objc as of servo/rust-cocoa#71. Could we have a rebase?

SSheldon:
Rebased! Unfortunately, the `Id` type in objc does not play nicely with the current cocoa crate, because it derefs to `&Object`/`&mut Object` but all the cocoa methods are defined on `*mut Object`. I'm thinking of the best way to reconcile this, but for now this code continues to use the current `IdRef` type.

bjz:
Hmm, still have a conflict - needs another rebase.

tomaka:
Yeah, that's probably because of me merging #325 :-/

SSheldon:
So close, rebased again!

Also note that a new version of glutin_cocoa will need to be published for this to compile. (I've been testing by forcing it to a local checkout.)

Cargo.toml
```diff
@@ -30,23 +30,11 @@ clock_ticks = "*"
...
-
-[target.x86_64-apple-darwin.dependencies.glutin_core_graphics]
-version = "0"
-
-[target.i686-apple-darwin.dependencies.glutin_core_foundation]
-version = "0"
-
-[target.x86_64-apple-darwin.dependencies.glutin_core_foundation]
-version = "0"
+[target.x86_64-apple-darwin.dependencies]
```
tomaka: So this removes `i686-apple-darwin`.

tomaka:
I've uploaded an updated version of `glutin_cocoa`.

Cargo.toml
SSheldon: `objc` doesn't currently compile on 32 bit (need to make some tweaks to the objc_msgSend calling conventions to support fpret), and it didn't look like some of the existing code supported 32bit either; glutin_cocoa uses f64 for NSPoint and NSSize when they should actually be CGFloat, and this code used to always use 3 as the align of a pointer. Based on that it seemed like 32bit wasn't fully supported.

Should I work on adding 32bit support to objc before merging this? Or should we wait for someone who's interested in 32bit support to to lead the effort? I can't even test it because I don't have a 32bit mac.

SSheldon: (Sorry, I should've mentioned this change more explicitly in the pull request, glad you asked about it.)

bjz:
Thanks @SSheldon!

Cargo.toml
nixpulvis: I just bumped into this took me a little while to realize what the problem was. Is there anything that can be done to improve the error messages so it's not just `cannot find crate \`objc\``?

PR #324, opened 2015-03-24
Win32 monitor handling

ryanstew:
Two issues with Win32 monitor handling:

* The rect top left is being ignored when creating a window, even with a monitor specified. This means the window is sized appropriately, but doesn't appear on the correct monitor. Fixed in 1c20ff8.
* `get_available_monitors()` is enumerating display adapters, but not monitors. This has two implications: the friendly names for monitors are less specific, and the native platform identifier for a monitor is incomplete. Fixed in 1c9c5c0.

This still fetches resolution/position at the adapter level. On all of my test hardware, there's a 1:1 mapping between monitors and adapters. I don't know when this isn't true. An alternative is to use `EnumDisplayMonitors` and `GetMonitorInfo` to extract dimensions and positions, but it's a larger/riskier change.

### DIFF
--- a/src/win32/init.rs
+++ b/src/win32/init.rs
@@ -164,6 +164,12 @@ unsafe fn init(title: Vec<u16>, builder: BuilderAttribs<'static>,
             (None, None)
         };
 
+        let (x, y) = if builder.monitor.is_some() {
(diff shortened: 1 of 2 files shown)


tomaka:
Thanks

PR #328, opened 2015-03-26
Cursor grabbing

tomaka:
- Replaces `grab_cursor` and `ungrab_cursor` by `set_cursor_state`.
- Partially implement this on win32.
- A few minor fixes.

### DIFF
--- a/examples/grabbing.rs
+++ b/examples/grabbing.rs
@@ -42,6 +42,11 @@ fn main() {
                               .ok().expect("could not grab mouse cursor");
                     }
                 },
+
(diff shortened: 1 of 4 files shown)


tomaka:
cc @aepsil0n

aepsil0n:
Nice. I guess it makes sense to have a single method and also to separate the hide/grab states.

tomaka:
I added that last commit, but I need to wait for an update in winapi before it compiles.

tomaka:
The functions have been added to winapi, but it can't be published: https://github.com/rust-lang/cargo/issues/1465

PR #330, opened 2015-03-27
Audit msg_send return types

SSheldon:
I audited the uses of `msg_send!` to make sure the specified return type matched that of the method. Shouldn't make much of a difference, but it never hurts to be correct :smile:


tomaka:
Thanks

PR #349, opened 2015-04-02
updating use statements from recent rust-cocoa module changes

ryanstew:
Builds are a mess right now. As I untangled pending PRs, I noticed this still dangling.

When servo/rust-cocoa#83 is merged, a new version of glutin_cocoa will need to be published. This change updates glutin after servo/rust-cocoa#81 so that publish will be unblocked.


SSheldon:
Thanks @ryanstew!

@tomaka, servo/rust-cocoa#83 was merged so could you publish an update to glutin_cocoa now?

tomaka:
Done

PR #351, opened 2015-04-03
Remove all features for 1.0 beta

tomaka:


tomaka:
Have to replace a `Weak<Window>` by `Arc<Window>` in `WindowProxy`.

Hopefully it's not a problem? @glennw

PR #352, opened 2015-04-03
Update OSX to compile on stable rust

SSheldon:
glutin wasn't compiling on OSX because `clone_from`, an unstable method, was implemented for `IdRef`. Deleting the method fixes this.

Also removed unused feature toggles from build.rs and examples so they will compile on the beta channel.


tomaka:
Ah, thanks!

PR #357, opened 2015-04-03
Fix OS X window closing bug

bobakker:
This is hilarious. I've been trying to find and fix this bug for ages.

This fixes the ```glutin::Event::Closed``` bug on OS X.


TyOverby:
HAHAHA

PR #365, opened 2015-04-08
Using x11-rs for X11 bindings.

Daggerbot:


tomaka:
I think you don't need `mem::transmute` to handle incompatibilities. Just adding `as *const _` or `as *mut _` works and is less error-prone.

src/x11/window/mod.rs
```diff
@@ -69,7 +67,7 @@ impl Drop for XWindow {
         unsafe {
             // we don't call MakeCurrent(0, 0) because we are not sure that the context
             // is still the current one
-            ffi::glx::DestroyContext(self.display, self.context);
+            ffi::glx::DestroyContext(mem::transmute(self.display), self.context);
```
tomaka: Here

src/x11/window/mod.rs
```diff
@@ -124,34 +122,34 @@ impl<'a> Iterator for PollEventsIterator<'a> {
...
                     return None;
                 }
             }
-    
-            match xev.type_ {
+
+            match xev.kind() {
                 ffi::KeymapNotify => {
-                    unsafe { ffi::XRefreshKeyboardMapping(&xev) }
+                    unsafe { ffi::XRefreshKeyboardMapping(mem::transmute(&xev)) }
```
tomaka: Here

src/x11/window/mod.rs
```diff
@@ -383,7 +381,7 @@ impl Window {
         // creating the color map
         let cmap = unsafe {
             let cmap = ffi::XCreateColormap(display, root,
-                visual_infos.visual, ffi::AllocNone);
+                mem::transmute(visual_infos.visual), ffi::AllocNone);
```
tomaka: Here

src/x11/window/mod.rs
```diff
@@ -415,7 +413,7 @@ impl Window {
         let window = unsafe {
             let win = ffi::XCreateWindow(display, root, 0, 0, dimensions.0 as libc::c_uint,
                 dimensions.1 as libc::c_uint, 0, visual_infos.depth, ffi::InputOutput,
-                visual_infos.visual, window_attributes,
+                mem::transmute(visual_infos.visual), window_attributes,
```
tomaka: Here

tomaka:
Bindings for OSMesa are missing.
You should compile with `cargo test --features "headless"` to try it out.

src/lib.rs
```diff
@@ -46,6 +46,10 @@ extern crate cocoa;
 extern crate core_foundation;
 #[cfg(target_os = "macos")]
 extern crate glutin_core_graphics as core_graphics;
+#[cfg(target_os = "linux")]
+extern crate x11;
+#[cfg(target_os = "freebsd")]
+extern crate x11;
```
SSheldon: I believe these can be combined with:
``` rust
#[cfg(any(target_os = "linux", target_os = "freebsd"))]
extern crate x11;
```

PR #367, opened 2015-04-09
Add support for Egl/OpenGL ES on Linux

tomaka:
*However*, this adds a link to `libEGL.so` and we want to avoid that.


tomaka:
Closing for #369

PR #368, opened 2015-04-09
Using `x11-rs` and `osmesa-rs`.

Daggerbot:
Build succeeds with `feature="headless"` enabled and disabled.


src/x11/headless.rs
```diff
@@ -49,7 +49,7 @@ impl HeadlessContext {
         unsafe {
             use std::ffi::CString;
             let c_str = CString::new(addr.as_bytes().to_vec()).unwrap();
-            ffi::OSMesaGetProcAddress(mem::transmute(c_str.as_ptr())) as *const ()
+            mem::transmute(ffi::OSMesaGetProcAddress(mem::transmute(c_str.as_ptr())))
```
tomaka: That's bad. `OSMesaGetProcAddress` should just return a `*const c_void` and not some fancy `Option`.

src/x11/window/mod.rs
```diff
@@ -92,14 +90,14 @@ pub struct WindowProxy {
 impl WindowProxy {
     pub fn wakeup_event_loop(&self) {
         let mut xev = ffi::XClientMessageEvent {
-            type_: ffi::ClientMessage,
+            _type: ffi::ClientMessage,
```
tomaka: Underscore prefixes mean that the variable is unused, that's probably not what we want.

src/lib.rs
```diff
@@ -46,6 +46,10 @@ extern crate cocoa;
 extern crate core_foundation;
 #[cfg(target_os = "macos")]
 extern crate glutin_core_graphics as core_graphics;
+#[cfg(target_os = "linux")]
+extern crate osmesa_sys;
```
tomaka: Must be `#[cfg(all(target_os = "linux", feature = "headless"))]` so that you don't need to link to osmesa if the `headless` feature is disabled (that's the reason why the headless feature exists in the first place).

src/x11/ffi.rs
```diff
@@ -1,10 +1,11 @@
-#![allow(dead_code)]
-#![allow(non_snake_case)]
-#![allow(non_camel_case_types)]
-#![allow(non_upper_case_globals)]
+pub use osmesa_sys::*;
```
tomaka: `#[cfg(feature = "headless")]`

tomaka:
> rustc optimizes Option<T> where T is an extern function (and I think other pointers as well) so that a null pointer is None and a non-null pointer is Some(_). This method is used in winapi-rs as well.

That's an implementation detail. Rust has a `#[repr(C)]` attribute that is applied to types that can communicate with C functions. I'm surprised that this compiles at all.

tomaka:
Why did you close this? :o

PR #369, opened 2015-04-09
Allow creating OpenGL ES contexts on Windows and Linux

tomaka:
~~Based over #367~~


tomaka:
This PR is dead, but I'm leaving it open for now because I'm copy-pasting some parts of the code.

PR #370, opened 2015-04-11
Exposing BuilderAttribs and its members

kvark:
I need access to the attributes before creating a window for gfx-rs. I don't see any danger in providing a read-only reference to these attributes and have their members public.

This is a blocker for https://github.com/gfx-rs/gfx_window_glutin

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -233,27 +233,25 @@ pub struct PixelFormat {
 }
 
 /// Attributes
-// FIXME: remove `pub` (https://github.com/rust-lang/rust/issues/23585)
 #[doc(hidden)]
 pub struct BuilderAttribs<'a> {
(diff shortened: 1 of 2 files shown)


kvark:
Actually, that's not what I need. I need the exact attributes of the window...

kvark:
Looks like I need some more deep modifications to make this work:
  - defaults to be present in `WindowBuilder` right off the bat
  - window creation to actually use the values, such as color/depth/stencil bits

@tomaka do you see it feasible?

tomaka:
You want for example the window to implement `get_format() -> [PixelFormat](https://github.com/tomaka/glutin/blob/master/src/lib.rs#L221-L233)`?

kvark:
@tomaka yes, this would be perfect

PR #371, opened 2015-04-11
Adds get_pixel_format() to Window

tomaka:
cc @kvark

Only implemented on win32. The x11 implementation needs some work (creating a `PixelFormat` struct from a `FBConfig`), the android implementation will probably move to the new `egl` module created in #367, and the cocoa implementation needs to be done by someone that has a mac.

### DIFF
--- a/src/android/mod.rs
+++ b/src/android/mod.rs
@@ -15,6 +15,7 @@ use Api;
 use BuilderAttribs;
 use CursorState;
 use GlRequest;
+use PixelFormat;
 use native_monitor::NativeMonitorId;
 
 pub struct Window {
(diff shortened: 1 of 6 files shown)


tomaka:
@bjz @SSheldon Is it technically possible to implement this on os/x?

SSheldon:
There's an [`NSOpenGLPixelFormat`](https://developer.apple.com/library/prerelease/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSOpenGLPixelFormat_Class/index.html) class that seems like a reasonable place to start. It looks like it can do what you want here, but I'm not very familiar with OpenGL

kvark:
@SSheldon I don't think you need to know much about GL here, just gather the `PixelFormat` data from somewhere where it exists. I'd appreciate if you looked at it on OSX ;)

tomaka:
The question is: is it possible to get infos like number of bits in the depth buffer, srgb or not, etc. from the `NSOpenGLPixelFormat`?
I haven't found anything in my short investigations, but I'm not familiar with these docs.

tomaka:
I'll merge in a few hours if no-one is opposed. :shipit:

kvark:
I see `NSOpenGLPixelFormat::get_values` which does exactly the same thing as X11 one.

PR #372, opened 2015-04-11
Print the pixel format in the `window` example

tomaka:
Based over #371 

Merge only when x11/cocoa implement the function, otherwise the window example will panic.

### DIFF
--- a/examples/window.rs
+++ b/examples/window.rs
@@ -24,6 +24,8 @@ fn main() {
     window.set_window_resize_callback(Some(resize_callback as fn(u32, u32)));
     unsafe { window.make_current() };
 
+    println!("Pixel format of the window: {:?}", window.get_pixel_format());
+
     let context = support::load(&window);
 
     while !window.is_closed() {


kvark:
Thanks @tomaka ! I may need to implement it for X11 myself. But even then, can't rely on it until someone takes care of OSX :/

kvark:
Why aren't you merging this? I'll have it conflicting with mine

tomaka:
I'm waitin for an answer in #371

fkaa:
Cocoa implementation is now merged, how is X11 status?

SSheldon:
I believe support was added for X11 in #375.

tomaka:
Oh I wanted to rebase this, but totally forgot.

fkaa:
@SSheldon while you're at it, do you mind taking a look at https://github.com/servo/rust-core-graphics/pull/35 and https://github.com/servo/rust-cocoa/pull/85? Both are very simple changes :)

SSheldon:
@fkaa alas, I'm not a contributor to those repos so I can't do much for you :) But I agree they look good!

PR #378, opened 2015-04-12
Fixes #358

### LINKED ISSUE #358: Apparently there's still problems with closing windows on OS X
The good news is that the window DOES close now. The bad news is that the program just keeps running. No resources are deallocated, and the program doesn't quit unless you ctrl-c it. I suspect this is because the window_should_close doesn't tell glutin to stop running, but I have no idea how to fix this. Any ideas?

bobakker:
The window closes now, and the process quits now, too. Finally!

### DIFF
--- a/src/cocoa/mod.rs
+++ b/src/cocoa/mod.rs
@@ -32,7 +32,7 @@ use std::sync::Mutex;
 use std::ascii::AsciiExt;
 use std::ops::Deref;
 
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks!

@SSheldon @bjz Nothing against moving `pending_events` inside `DelegateState`?

SSheldon:
Nope, this looks reasonable! Might want to do some refactor now that all the state is in `DelegateState`, but I'm not concerned about that now.

PR #384, opened 2015-04-19
Fix alert sounds on keypress for cocoa

fkaa:
Passing on keypress event to parent would make it think the event was not consumed and emit an alert sound.

### DIFF
--- a/src/api/cocoa/mod.rs
+++ b/src/api/cocoa/mod.rs
@@ -199,9 +199,11 @@ impl<'a> Iterator for PollEventsIterator<'a> {
                 NSDefaultRunLoopMode,
                 YES);
             if event == nil { return None; }
-            NSApp().sendEvent_(event);
 
-            let event = match msg_send![event, type] {
+            let event_type = msg_send![event, type];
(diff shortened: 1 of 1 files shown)


src/cocoa/mod.rs
```diff
@@ -197,17 +197,18 @@ impl<'a> Iterator for PollEventsIterator<'a> {
                 NSDefaultRunLoopMode,
                 YES);
             if event == nil { return None; }
-            NSApp().sendEvent_(event);
```
SSheldon: Instead of duplicating the `sendEvent_`, which seems like it'd be easy to miss in one case and cause a weird bug, we could do:
``` rust
let event_type = msg_send![event, type];
NSApp().sendEvent_(if let NSKeyDown = event_type { nil } else { event });
match event_type { ... }
```

Then it's more obvious we always call `sendEvent_` and in which case we pass it nil.

tomaka:
Needs a rebase :(

`src/cocoa/mod.rs` has moved to `src/api/macos/mod.rs`

tomaka:
Thanks

PR #390, opened 2015-04-23
Set WM_CLASS based on initial window title.

glennw:


Kagami:
I think there is some problem with that code. Tile managers like XMonad, Openbox have ability to apply some properties to windows (e.g. make window floated or tiled) based on their properties.

I created window with some initial title using glutin and I can see title and WM_CLASS were correctly set using `xprop` utility. But I can't force XMonad to make glutin window floated using WM_CLASS value, only match based on title works.

I think it's because glutin sets WM_CLASS property too late (i.e. after window was appeared) and so WM can't match its window.

PR #396, opened 2015-04-24
[WIP] Draft for libgbm support

tomaka:


LongChair:
@tomaka : that PR would have been nice for any embedded device where X11 is not the way to go :)

Any plans to continue this work at some point ? :)

tomaka:
@LongChair The problem of this PR is that I couldn't get it to work. Everything my code did was (I think) correct, but libgbm would return an error every time.

PR #402, opened 2015-04-26
Add libcaca backend

tomaka:


tomaka:
Need to add a Cargo feature for this before merging.

PR #405, opened 2015-04-29
[Breaking change] Fix mouse wheel value on win32 and return a f64 instead of i32 in the MouseWheel event

### LINKED ISSUE #260: 2D mouse wheel support
Currently it supports only vertical scrolling.

GLFW and SDL2 implements this as a relative 2D coordinate.

tomaka:
Close #260

### DIFF
--- a/src/api/cocoa/mod.rs
+++ b/src/api/cocoa/mod.rs
@@ -275,7 +275,7 @@ impl<'a> Iterator for PollEventsIterator<'a> {
                     self.window.delegate.state.pending_events.lock().unwrap().extend(events.into_iter());
                     event
                 },
-                NSScrollWheel           => { Some(MouseWheel(event.scrollingDeltaY() as i32)) },
+                NSScrollWheel           => { Some(MouseWheel(event.scrollingDeltaX() as f64, event.scrollingDeltaY() as f64)) },
                 _                       => { None },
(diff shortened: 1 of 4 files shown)


retep998:
:cactus:

fkaa:
I think while we're at it we might as well add horizontal scrolling (`MouseWheel(f64, f64)`), which will close #260.

tomaka:
Oh, right.

src/api/cocoa/mod.rs
```diff
@@ -275,7 +275,7 @@ impl<'a> Iterator for PollEventsIterator<'a> {
                     self.window.delegate.state.pending_events.lock().unwrap().extend(events.into_iter());
                     event
                 },
-                NSScrollWheel           => { Some(MouseWheel(event.scrollingDeltaY() as i32)) },
+                NSScrollWheel           => { Some(MouseWheel(event.scrollingDeltaX() as f64, event.scrollingDeltaY() as f64)) },
```
tomaka: So I guess that `event.scrollingDeltaX()` probably exists, even though I didn't check.

fkaa: Correct!

tomaka:
Will merge before releasing 0.1.

Issue #362, opened 2015-04-06
"invalid pixel format attribute" on OS X


dylanede:
All of the examples fail to run correctly on my OS X 10.8.5 machine at the moment, logging "invalid pixel format attribute" then failing to the create the OpenGL context and quitting. This may not be a recent change that caused this, since I have not touched glutin or glium since December.

dylanede:
After upgrading to OS X 10.10, the problem no longer occurs. So this looks to be specific to 10.8.3 at least. Not sure whether you want to continue supporting that version.

fkaa:
The only pixel format attributes that were added since December, 2014 seem to be `NSOpenGLPFAOpenGLProfile` (10.7). I don't have access to an earlier version of OS X so testing this seems tricky.

@dylanede did it panic and say `invalid pixel format attribute`? Replacing an enum with some random numbers only gives me this:
```rust
thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: OsError("Couldn\'t create OpenGL context")', /Users/rustbuild/src/rust-buildbot/slave/nightly-dist-rustc-mac/build/src/libcore/result.rs:768
Process didn't exit successfully: `target/debug/examples/vsync` (exit code: 101)
```

dylanede:
@fkaa The message I received was very similar to that, only the previous the line of output would say

    invalid pixel format attribute

Just to be clear, the last time that I used Glutin on OS X 10.8.5 before encountering this problem was in December 2014, but I think the actual date was near the start of December, so if you're looking for relevant changes make sure to include those in December.

fkaa:
Possible perpetrator: `NSOpenGLProfileVersion4_1Core`.

OS X 10.8.5 only supported up to 3.2 Core (`NSOpenGLProfileVersion3_2Core`) and `NSOpenGLProfileVersion4_1Core` is defined separately in rust-cocoa as `0x4100`. This could be a simple case of requesting a context which is not available (`NSOpenGLProfileVersion4_1Core` is default, afaik).

Possible solution: Check against current OS version and cap version accordingly (eg. on 10.8.5, `GlRequest::Latest` would request `NSOpenGLProfileVersion3_2Core` instead)

fkaa:
Should be fixed in #470.

PR #409, opened 2015-04-30
GL core profile flag

kvark:
Would be really nice to have someone testing this on Windows and OSX.

### DIFF
--- a/src/api/glx/mod.rs
+++ b/src/api/glx/mod.rs
@@ -3,6 +3,7 @@
 use BuilderAttribs;
 use CreationError;
 use GlContext;
+use GlProfile;
 use GlRequest;
 use Api;
 use PixelFormat;
@@ -52,14 +53,15 @@ impl Context {
                 },
             }
 
-            if let Some(core) = builder.gl_core {
+            if let Some(profile) = builder.gl_profile {
+                let flag = match profile {
+                    GlProfile::Compatibility =>
+                        ffi::glx_extra::CONTEXT_COMPATIBILITY_PROFILE_BIT_ARB,
+                    GlProfile::Core =>
+                        ffi::glx_extra::CONTEXT_CORE_PROFILE_BIT_ARB,
+                };
                 attributes.push(ffi::glx_extra::CONTEXT_PROFILE_MASK_ARB as libc::c_int);
-                attributes.push(if core {
-                        ffi::glx_extra::CONTEXT_CORE_PROFILE_BIT_ARB
-                    } else {
-                        ffi::glx_extra::CONTEXT_COMPATIBILITY_PROFILE_BIT_ARB
-                    } as libc::c_int
-                );
+                attributes.push(flag as libc::c_int);
             }
 
             if builder.gl_debug {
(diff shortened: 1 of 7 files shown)


tomaka:
Is there really a problem with the design of `Latest`?

I really don't see the point of having the user request a core context if they can be already choose the OpenGL version. I could understand why someone would want a compatibility context (using at the same time `glBegin` and GL 4.5 functions), but that's not what your PR is.

Plus, your PR modifies `with_srgb`, which is unrelated to core/compatibility. Please open another PR for this.

fkaa:
@tomaka Perhaps not related to the design of `Latest`, but right now (at least for cocoa) it requests the highest available version for *latest* hardware, rather than the machine it's running on, which is very specific for OS X (see [comment](https://github.com/tomaka/glutin/issues/362#issuecomment-97728684) in #362)

kvark:
@tomaka 

> Is there really a problem with the design of Latest?

Latest as it is implemented now means the driver can give you any context it wants on GLX/WGL. For example, my linux workstation gives a 2.1 context for some reason.

> I could understand why someone would want a compatibility context (using at the same time glBegin and GL 4.5 functions), but that's not what your PR is.

It is a part of my PR. Having `with_gl_core(false)` forces the compatibility profile.

> Plus, your PR modifies with_srgb, which is unrelated to core/compatibility. Please open another PR for this.

It's such a minor and obvious change that I didn't bother. If you have any concerns about it - tell me. If not - let's not waste our time in the bureaucracy.

tomaka:
> Latest as it is implemented now means the driver can give you any context it wants on GLX/WGL. For example, my linux workstation gives a 2.1 context for some reason.

Your PR wouldn't change anything, because the default context type is `core`.

kvark:
> Your PR wouldn't change anything, because the default context type is core.

Neither will it break anything. Is Glutin a low-level abstraction, or a high-level? Since GLX/WGL/Cocoa explicitly differentiate between Core and Compatibility profiles, and OpenGL (of course) does, why not expose more control of it?

tomaka:
I want to delay the change to sRGB in order to put it in version 0.1, as it's a breaking change.

For the core/compatibility thing, I guess it makes sense to offer the choice for the few people that would want this. However I'd prefer if it was `with_compatibility` instead of `with_core`.

kvark:
@tomaka I see. How about `with_gl_profile` accepting an enum then? Not as ergonomic, but at least clearer.

kvark:
@tomaka AppVeyor hang for some reason, but my changes are in place.

src/api/glx/mod.rs
```diff
@@ -52,6 +53,17 @@ impl Context {
                 },
             }
 
+            if let Some(profile) = builder.gl_profile {
+                let flag = match profile {
+                    GlProfile::Compatibility =>
+                        ffi::glx_extra::CONTEXT_CORE_PROFILE_BIT_ARB,
+                    GlProfile::Core =>
+                        ffi::glx_extra::CONTEXT_COMPATIBILITY_PROFILE_BIT_ARB,
```
tomaka: I think that's wrong

kvark: lol, nice catch!

PR #407, opened 2015-04-29
Rework EGL

tomaka:
Much cleaner implementation.

PR #410, opened 2015-04-30
(WIP) Wayland backend

vberger:
These are the firsts steps of building a wayland backend for linux. This is currently a big work in progress and not destined to be merged right away, I open this PR to have some feedback.

This backend is built on top of #407 and uses my [wayland-client library](https://crates.io/crates/wayland-client). This library does runtime detection of wayland using `dlopen`, so it does not introduces a new link-time dependency.

Currently, it supports the creation of a window with an EGL context, and mouse input (motion of the pointer and click). For some reason, the size of the created window does not seem right, I'm trying to figure out why.

To test it, due to current conflicts in the API, you'll have to replace the module `x11` by the module `wayland` in `src/platform/linux.rs` and disable modules `glx` and `x11` in `src/api/mod.rs`.

### DIFF
--- a/src/api/wayland/mod.rs
+++ b/src/api/wayland/mod.rs
@@ -1,0 +1,378 @@
+#![cfg(target_os = "linux")]
+
+use self::wayland::egl::{EGLSurface, is_egl_available};
+use self::wayland::core::{Display, Registry, Compositor, Shell, ShellSurface,
+                          Seat, Pointer, default_display, WSurface, SurfaceId,
+                          Surface};
+
+use libc;
+use api::dlopen;
+use api::egl::Context as EglContext;
+
+use BuilderAttribs;
+use CreationError;
+use Event;
+use PixelFormat;
+use CursorState;
+use MouseCursor;
+use GlContext;
+
+use std::collections::{VecDeque, HashMap};
+use std::sync::{Arc, Mutex};
+use std::ffi::CString;
+
+extern crate wayland_client as wayland;
+
+struct WaylandContext {
+    pub display: Display,
+    pub registry: Registry,
+    pub compositor: Compositor,
+    pub shell: Shell,
+    pub seat: Seat,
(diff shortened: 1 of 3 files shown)


tomaka:
\o/

src/lib.rs
```diff
@@ -51,6 +51,8 @@ extern crate core_foundation;
 extern crate glutin_core_graphics as core_graphics;
 #[cfg(any(target_os = "linux", target_os = "freebsd"))]
 extern crate x11;
+#[cfg(target_os = "linux")]
+extern crate wayland_client as wayland;
```
tomaka: Please move this inside `src/api/wayland/mod.rs` (the other libs should be moved as well eventually)

src/api/wayland/mod.rs
```diff
@@ -0,0 +1,304 @@
+use wayland::egl::{EGLSurface, is_egl_available};
```
tomaka: `#![cfg(target_os = "linux")]` here

src/api/wayland/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+    pub fn platform_display(&self) -> *mut libc::c_void {
+        unimplemented!()
+    }
+
+    pub fn platform_window(&self) -> *mut libc::c_void {
+        unimplemented!()
+    }
+
+    pub fn get_api(&self) -> ::Api {
+        ::Api::OpenGlEs
```
tomaka: `self.context.get_api()`

src/api/wayland/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    pub fn get_pixel_format(&self) -> PixelFormat {
+        unimplemented!();
```
tomaka: `self.context.get_format().clone()`

glennw:
:fireworks:

PR #413, opened 2015-04-30
Add pixelformat for cocoa and remove individual color components

fkaa:
cc #372


fkaa:
Not that this also needs to be implemented for X11, Windows, etc.

tomaka:
Closing in favor of #414

PR #416, opened 2015-04-30
Adding SWP_NOMOVE flag to prevent the window from moving to 0,0 when setting inner size on Windows 8

datalus:
Hey tomaka, we spoke earlier on irc about this issue. I was able to reproduce it on my Windows 8 box. Adding in the flag SWP_NOMOVE solved the issue for me. I am submitting a pull request for consideration into fixing this issue for Windows 8+ for glutin.


tomaka:
Thanks!

PR #417, opened 2015-05-01
Use the EGL API with Android

tomaka:


ozkriff:
Color picker is ZoC is broken with this commit - https://github.com/ozkriff/zoc/blob/ea6370/src/visualizer/src/picker.rs#L130. Hmmm...

tomaka:
@ozkriff Ah, thanks for testing!

The new EGL module should be cleaner than the old one. However what this changes is that the format of the color buffer, size of the depth buffer, etc. that are chosen by default by glutin are different than with the old implementation.

Maybe there are some rounding errors with your code? You could try adding `with_pixel_format(24, 8)` when creating the window?

ozkriff:
yes, `with_pixel_format(24, 8)` solves the problem :)

tomaka:
Nice, let's merge then.

PR #431, opened 2015-05-05
Copy Linux dependencies to ARM target

krdln:
Add `[target.arm-unknown-linux-gnueabihf.dependencies]` with osmesa-sys and x11. This enables glutin to compile on ARM Linux.


tomaka:
Thanks!

PR #436, opened 2015-05-08
Added mut so glutin can compile on nightly.

Esption:
It's a silly one line commit, without the mut, Glutin fails to compile on nightly 2015-05-07.  It doesn't break beta3, either.


tomaka:
Thanks

PR #439, opened 2015-05-08
Fixed build warnings on Linux

Esption:
Should fix some of https://github.com/tomaka/glutin/issues/428

Fixed build warnings on Linux.

A few of them were with ```#![allow(...)]```s, however.


tomaka:
Thanks!

PR #441, opened 2015-05-09
Continuation of wayland API

vberger:
Still not finished, but adding support for `WindowProxy::wakeup_event_loop()` and `MonitorID`, as well as stubbing `Window::get_position()` and `Window::set_position()`, which are not possible with the current wayland stable protocol.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,15 +50,15 @@ kernel32-sys = "0.1"
 
 [target.i686-unknown-linux-gnu.dependencies]
 osmesa-sys = "0.0.5"
-wayland-client = "0.1.3"
+wayland-client = "0.1.4"
 x11-dl = "=1.0.1"
 
 [target.x86_64-unknown-linux-gnu.dependencies]
 osmesa-sys = "0.0.5"
-wayland-client = "0.1.3"
+wayland-client = "0.1.4"
 x11-dl = "=1.0.1"
 
 [target.arm-unknown-linux-gnueabihf.dependencies]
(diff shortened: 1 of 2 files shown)


src/api/wayland/mod.rs
```diff
@@ -159,22 +162,43 @@ pub struct WindowProxy;
...
+pub struct MonitorID {
+    output: Arc<Output>
+}
 
 pub fn get_available_monitors() -> VecDeque<MonitorID> {
-    VecDeque::new()
+    if let Some(ref ctxt) = *WAYLAND_CONTEXT {
+        ctxt.outputs.iter().map(|o| MonitorID::new(o.clone())).collect()
+    } else {
+        VecDeque::new()
```
tomaka: Shouldn't this be a panic?
And if not, `VecDeque::with_capacity(0)` should be preferred instead of `new()`.

Cargo.toml
```diff
@@ -50,12 +50,12 @@ kernel32-sys = "0.1"
...
+wayland-client = "0.1.4"
 x11-dl = "=1.0.1"
 
 [target.x86_64-unknown-linux-gnu.dependencies]
 osmesa-sys = "0.0.5"
-wayland-client = "0.1.3"
+wayland-client = "0.1.4"
 x11-dl = "=1.0.1"
 
 [target.arm-unknown-linux-gnueabihf.dependencies]
```
tomaka: You forgot arm

vberger: added

PR #448, opened 2015-05-12
Make platform::linux generic over X11 and Wayland

vberger:
This PR makes the linux platform code generic over Wayland and X11.

At runtime, if a wayland context is available (ie: the libraries have been successfully loaded **and** a connexion to the wayland server was successfully opened), it will be used. Otherwise glutin will fallback on X11.

The second commit of this PR actually disables the ability to use Wayland, because the backend is not yet production-ready at all (we are still missing important features like keyboard input), however it is needed to have this code in place to allow compilation of glutin while developing the backend.

Thus, this PR does not change the behavior of glutin *at all*, but a simple revert of the second commit will be enough to enable the Wayland backend once it's ready.

### DIFF
--- a/src/api/glx/mod.rs
+++ b/src/api/glx/mod.rs
@@ -14,6 +14,8 @@ use std::{mem, ptr};
 
 use api::x11::ffi;
 
+use platform::Window as PlatformWindow;
+
 pub struct Context {
     glx: ffi::glx::Glx,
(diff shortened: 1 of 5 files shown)


tomaka:
Ultimately there should be deeper changes, like opening and keeping the X display in `platform/linux`. But that will do the trick for now.

PR #449, opened 2015-05-13
x11: Use 10 pixels as the default scroll wheel tick

### LINKED ISSUE #420: Scrolling speed is too slow
In https://github.com/servo/servo/issues/5660 they decided that glutin should be sending a more reasonable scrolling speed than it is doing. They shouldn't need to do post processing to fix it.

mrobinson:
Scroll wheel ticks are simply mouse button presses, so we need
to pick a delta for them to send to clients.

Fixes #420.

### DIFF
--- a/src/api/x11/mod.rs
+++ b/src/api/x11/mod.rs
@@ -239,16 +239,19 @@ impl<'a> Iterator for PollEventsIterator<'a> {
 
                     let state = if xev.get_type() == ffi::ButtonPress { Pressed } else { Released };
 
+                    // Scroll wheel ticks do not have any inherent delta in XInput, because
+                    // they register as mouse button clicks. We use 10 for the moment.
+                    const SCROLL_TICK_DELTA : f64 = 10.0;
                     let button = match event.button {
                         ffi::Button1 => Some(Left),
                         ffi::Button2 => Some(Middle),
                         ffi::Button3 => Some(Right),
                         ffi::Button4 => {
-                            self.window.pending_events.lock().unwrap().push_back(MouseWheel(0.0, 1.0));
+                            self.window.pending_events.lock().unwrap().push_back(MouseWheel(0.0, SCROLL_TICK_DELTA));
                             None
                         }
                         ffi::Button5 => {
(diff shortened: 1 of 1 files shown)


tomaka:
I really don't see the point in returning `10` or `-10` instead of `1` or `-1`.

mrobinson:
How do you suggest that clients differentiate between a touchpad delta of 1 and a mouse wheel tick? I'm happy to implement whatever design you feel is best, but I need to know how to handle this in Servo properly.

mrobinson:
Also, I insist that the original commit makes at least 1/8 of a sense.

tomaka:
For what I see, [glfw does the same as glutin](https://github.com/glfw/glfw/blob/master/src/x11_window.c#L997-L1004).
I don't think using a touchpad makes a difference.

mrobinson:
I tested this with a touchpad in Linux and confirmed that XEvent was still sending scroll ticks. I think it depends on the driver though, as GTK+ has support for receiving a delta as well (https://developer.gnome.org/gdk3/stable/gdk3-Events.html#gdk-event-get-scroll-deltas). Right now I don't think the client can cleanly figure out what a delta of (0.0, 1.0) even means. Perhaps it is a tiny movement on a Cocoa device or a tick from an X11 device.

The core issue, I think, is that there are two types of wheel events. There are those that should carry a directional tick payload and those that should include a delta. I think the cleanest way to do this would be to have MouseWheel carry an enum like this:

```
enum MouseWheelType {
    ForwardTick,
    BackwardTick,
    Delta(f64, f64),
}
```
The reason I didn't do this originally is that the work-around doesn't break the current API. With this approach though, the client can properly detect and deal with forward and backward ticks.

TyOverby:
@mrobinson What about sideways ticks?

mrobinson:
Oh, that's a good point. They could be handled like this:

enum Direction { Up, Down, Left, Right }
enum MouseWheelType {
    Tick(Direction),
    Delta(f64, f64),
}

TyOverby:
@mrobinson: I think this is a good API design.

PR #451, opened 2015-05-14
x11: Express directional mouse wheel ticks in the API

### LINKED ISSUE #420: Scrolling speed is too slow

mrobinson:
In X11, mouse wheel ticks are just direction button presses. Previously
the code represented that as a scroll delta of 1. It is really up to the
application to process mouse wheel ticks though.

Fixes #420.

### DIFF
--- a/src/api/cocoa/mod.rs
+++ b/src/api/cocoa/mod.rs
@@ -39,6 +39,7 @@ use std::ascii::AsciiExt;
 use std::ops::Deref;
 
 use events::Event::{Awakened, MouseInput, MouseMoved, ReceivedCharacter, KeyboardInput, MouseWheel, Closed};
+use events::MouseWheelMotion;
 use events::ElementState::{Pressed, Released};
 use events::MouseButton;
 use events;
@@ -275,7 +276,7 @@ impl<'a> Iterator for PollEventsIterator<'a> {
                     self.window.delegate.state.pending_events.lock().unwrap().extend(events.into_iter());
                     event
                 },
-                NSScrollWheel           => { Some(MouseWheel(event.scrollingDeltaX() as f64, event.scrollingDeltaY() as f64)) },
+                NSScrollWheel           => { Some(MouseWheel(MouseWheelMotion::Delta(event.scrollingDeltaX() as f64, event.scrollingDeltaY() as f64))) },
                 _                       => { None },
             };
 
--- a/src/api/win32/callback.rs
+++ b/src/api/win32/callback.rs
@@ -111,12 +111,13 @@ pub unsafe extern "system" fn callback(window: winapi::HWND, msg: winapi::UINT,
 
         winapi::WM_MOUSEWHEEL => {
             use events::Event::MouseWheel;
+            use events::MouseWheelMotion;
 
             let value = (wparam >> 16) as i16;
             let value = value as i32;
             let value = value as f64 / winapi::WHEEL_DELTA as f64;
 
-            send_event(window, MouseWheel(0.0, value));
+            send_event(window, MouseWheel(MouseWheelMotion::Delta(0.0, value)));
 
             0
         },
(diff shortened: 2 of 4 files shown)


mrobinson:
@tomaka Here's another potential fix for the issue of mouse wheel ticks on X11. I'm happy to adjust any of the particulars here as well.

TyOverby:
What happened to 

```
enum Direction { Up, Down, Left, Right }
enum MouseWheelType {
    Tick(Direction),
    Delta(f64, f64),
}
```

from the previous PR comments?

mrobinson:
I'm happy to implement that as well, but I wasn't sure if two level of abstraction was the cleanest approach.

tomaka:
I still think this is a bad design. In gtk, mouse evenst have a direction **and** a delta, not either.

Let's say I'm using glutin. The only way for me to process mouse events is this one:

```rust
match event {
    MouseWheel(Delta(x, y)) => submit_mouse_wheel(x, y),
    MouseWheel(TickUp) => submit_mouse_wheel(0, 1),
    MouseWheel(TickDown) => submit_mouse_wheel(0, -1),
}
```

Using any value except `1` and `-1` would make the scrolling different on X11 (where it uses ticks) than on cocoa/windows (where it uses deltas).

Which thus defeats the whole point of this change.

mrobinson:
No, mouse events in GTK+ have either a delta *or* a direction. Quoting from the documentation at https://developer.gnome.org/gdk3/stable/gdk3-Event-Structures.html#GdkEventScroll:

> Generated from button presses for the buttons 4 to 7. Wheel mice are usually configured to generate button press events for buttons 4 and 5 when the wheel is turned. Some GDK backends can also generate “smooth” scroll events, which can be recognized by the GDK_SCROLL_SMOOTH scroll direction. For these, the scroll deltas can be obtained with gdk_event_get_scroll_deltas().

Most mice generate simple button events (what I am calling ticks) when the mouse wheel is turned. Some devices (like the Apple touchpad) are smooth scrolling or continuous, which generates a delta. GTK+ applications must handle both sorts of events. Here's [an example](https://github.com/GNOME/gtk/blob/master/gtk/gtkscrolledwindow.c#L2963) from the GTK+ source code. You can see the code falling back if `gdk_event_get_scroll_deltas()` returns false.

This also seems to be the case in Cocoa, although glutin does not appear to be checking the return value of [`hasPreciseScrollingDeltas`](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSEvent_Class/#//apple_ref/occ/instp/NSEvent/scrollingDeltaY).  Something like this change is likely necessary for the Cocoa platform as well. GLFW simply [multiples the returned value by 0.1](https://github.com/glfw/glfw/blob/master/src/cocoa_window.m#L563), but that isn't what the Cocoa documentation suggests. Instead, the scroll should be by the precise amount returned.

I think the application should decide how to deal with these various scroll wheel types.. glutin could also scale precise / smooth scrolling events or scale X11 events though. Doing nothing means that glutin applications will be broken with some pointer devices. Doing what GLFW does means copying an approach which looks a bit bogus.

tomaka:
What does a `Delta` mean?
Let's say I get a mouse wheel event containing `Delta`, how do I interpret the values in it?

mrobinson:
Apart from the build error, I think my initial stab at this API isn't good enough. Here's my current proposal:

1. Wheel events carry either PreciseScrollDelta(f64, f64) or ScrollDelta(f64, f64).
2. In the case of ScrollDelta, the application scrolls by "lines" as Cocoa calls them. This is the chunky scroll that you would see when using a mouse wheel. The application will need to multiply this by scroll chunk ("line") size.
3. In the case of PreciseScrollDelta, the application will scroll by the number of pixels specified by the delta itself.

I think this approach will give the necessary flexibility for all platforms, but perhaps we can chat about this tomorrow over IRC.

tomaka:
Closing this one (see #483)

PR #453, opened 2015-05-15
Add API for transparency and decorations and add support for win32

tomaka:
~~Blocked on https://github.com/retep998/winapi-rs/pull/95 and https://github.com/retep998/winapi-rs/pull/94~~


tomaka:
Apparently this is possible on X11. It's obviously not possible on Android since there is no compositor.

@fkaa @SSheldon Is something like this possible on OS/X?

PR #454, opened 2015-05-16
Fix GLX creating only 3.0 contexts

### LINKED ISSUE #16: X11 only creates 3.0 contexts
See #2

tomaka:
Fix #16 

Would be nice if someone could test this. I'm currently having some issues with my drivers, and only have OpenGL 2.1.


tomaka:
@vberger Reported getting a 3.3 context \o/

It would be nice if someone else could test it, otherwise I will merge.

Issue #89, opened 2014-10-29
X11 doesn't like creating and destroying a lot of windows


tomaka:
To reproduce, clone https://github.com/tomaka/glium/ and run `cargo test`.

I don't know if it's some kind of race-condition, or just events not being processed correctly.

cmr:
@tomaka I'm curious how piglit gets past this. Or maybe they use separate processes. Worth looking into.

tomaka:
I think they just use a single thread (this is python after all).
If I run glium's tests with `RUST_TEST_THREADS=1`, everything works fine.

cmr:
That seems plausible.

tomaka:
Closing for https://github.com/tomaka/winit/issues/32

PR #455, opened 2015-05-17
Rework the X implementation to use only one X connection

tomaka:
~~Blocked on https://github.com/bjz/gl-rs/pull/340~~


tomaka:
This was intended to be a fix for #89, but this doesn't fix it at all.

tomaka:
Merging anyway, since this makes the code more proper.

PR #456, opened 2015-05-18
Fix handling of numpad keys w.r.t. numlock (derived from fix mentione…

aweinstock314:
…d at http://www.kaffe.org/pipermail/kaffe/2000-April/175201.html).


tomaka:
Thanks!

PR #462, opened 2015-05-22
[WIP] Joystick: platform independent interface

cmr:
This is essentially the same API that SDL2 and GLFW support. It does not
have the limitation of a max number of joysticks. It's possible that we may
want to support a max number of joysticks, because that can avoid an
allocation, but there's not really a significant amount of allocation that
can happen due to this!

I've trawled docs for all the major platforms and this API should be fine.
SDL2 emulates both a poll and event-based interface, by either polling
joystick state during the event loop, or updating a local copy of joystick
state and returning that for the poll.

The way this API will work, is that users are encouraged to call
`scan_joysticks` on startup, when a joystick is detached (noticed by `poll`
returning `None`), and when their game logic wants it ("Player 2 press
start!"), but _not_ every frame because it can be slow. The `Joystick` will
internally contain an `Rc` to any state it needs. `poll` should be
non-blocking and return whatever state it had at the last poll if it would
block (though I am not sure if there are many devices where that would be
an issue).

On some platforms, rather than polling for joystick state, you get events.
AFAICT the only platform that _does not_ support events is Windows with
XInput.

I'm not sure I like tying this to the Window, as on no platform are
joystick events tied directly to a window. Instead, all applications
observe the same state. Plus, scanning for joysticks can be off-main-thread
for most platforms!

There's a fair amount of room for alternative design. This seems easiest
and most conservative.


cmr:
Action item from IRC: make it separate from `Window`, they have no dependency.

Cargo.toml
```diff
@@ -19,6 +19,7 @@ gl_common = "0.0.4"
 lazy_static = "0.1.10"
 libc = "0.1"
 shared_library = "0.1.0"
+pool = { git = "https://github.com/cmr/pool" } # "0.1.3"
```
tomaka: This will need to be fixed before merging.

cmr:
Action item from IRC: expose both event and poll-based interfaces. Emulate them on platforms where they are different (only XInput comes to mind).

tomaka:
All joystick-related stuff could be moved from `src/window.rs` to `src/joystick.rs`

cmr:
Will need to add "ball" support as well, forgot about that. http://www.usb.org/developers/hidpage/Hut1_12v2.pdf has a comprehensive list of all the various kinds of input devices (only some of which Glutin cares about).

cmr:
@tomaka I will ping you when it's ready for another review.

PR #469, opened 2015-05-23
Fix vsync not being disabled by default in cocoa

fkaa:
The apple docs are lying!

### DIFF
--- a/src/api/cocoa/mod.rs
+++ b/src/api/cocoa/mod.rs
@@ -553,10 +553,8 @@ impl Window {
                     };
 
                     cxt.setView_(view);
-                    if builder.vsync {
-                        let value = 1;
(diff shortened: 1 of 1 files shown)


fkaa:
Tested in both another XCode project and in glutin, so [this](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSOpenGLContext_Class/index.html#//apple_ref/c/tdef/NSOpenGLContextParameter) piece of documentation is apparently untrue:

> **NSOpenGLCPSwapInterval**
> * Sets or gets the swap interval.

> The swap interval is represented as one long. If the swap interval is set to **0 (the default)**, the flushBuffer method executes as soon as possible, without regard to the vertical refresh rate of the monitor. If the swap interval is set to 1, the buffers are swapped only during the vertical retrace of the monitor.

> Available in OS X v10.0 and later.

tomaka:
I guess it can't hurt to always set this value anyway.

PR #475, opened 2015-05-24
Add OS version detection on cocoa for selecting latest gl version. Fixes #470

### LINKED ISSUE #470: Detect the version of OS/X and don't try to create OpenGL 4.1 if it's not supported

fkaa:
Will select the latest OpenGL version based on the OS version (not all gfx cards on 10.10 support 4.1, and while this could be queried, it feels like a lot of work for little gain)


fkaa:
@tomaka will need to update the `glutin_cocoa` crate to latest version

PR #476, opened 2015-05-25
Fix set_cursor_state on X11

aepsil0n:
When the state is not changed, no code must be executed. Such no-ops previously hit the unimplemented cursor hiding.


tomaka:
Thanks

PR #477, opened 2015-05-26
[WIP] ios support

rozaliev:
I've just started to work on initial ios support and decided to open PR early to discuss some ios specifics and workarounds. btw code is messy now, excuse me for that, I'll clean it up later.

As turns out ios is very different from other platforms and some hacks are required to make it work. 

First lets start with UIApplicationMain, it creates app object and sets up event loop. As far as I understand it's mandatory to use and there is no nice workaround. It blocks so I had to setjmp out of it to return control to glutin. After that on every poll_events.next() we can call CFRunLoopRunInMode to give os some time to process events, then put them in our own queue in delegate and take them from the queue in the poll iterator.

Then there are different kinds of events. Some of them, like touch events, can be processed asynchronously, but events like applicationWillTerminate must be sync, otherwise app will quit and you'll have no control over it. So you have to setjmp out of it too. SDL does the same thing. (not yet implemented)

The last thing is EAGL, Apple's version of EGL. And it's somewhat confusing, because it requires to initialize render buffer and then pass it when swap buffer is called. I haven't started working on this one yet, so I guess I'll grok it later.
(https://developer.apple.com/library/prerelease/ios/documentation/OpenGLES/Reference/EAGLContext_ClassRef/index.html)

I'll appreciate any feedback :)

btw it compiles, event polling works, but there is only one event for now WillEnterForeground. I plan to make it possible to draw something and refactor, hope to make it mergeable soon.

ping @cmr

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,18 @@ clock_ticks = "0.0.5"
 [target.arm-linux-androideabi.dependencies.android_glue]
 version = "0"
 
+[target.x86_64-apple-ios.dependencies]
+objc = "0.1"
+
+[target.aarch64-apple-ios.dependencies]
+objc = "0.1"
+
+[target.armv7s-apple-ios.dependencies]
+objc = "0.1"
+
+[target.armv7-apple-ios.dependencies]
+objc = "0.1"
+
 [target.x86_64-apple-darwin.dependencies]
 objc = "0.1"
 cocoa = "0"
--- a/src/api/ios/delegate.rs
+++ b/src/api/ios/delegate.rs
@@ -1,0 +1,223 @@
+use libc;
+use std::mem;
+use super::DelegateState;
+use Event;
+use events::{ Touch, TouchPhase };
+
+use objc::runtime::{ Class, Object, Sel, BOOL, YES };
+use objc::declare::{ ClassDecl };
+
+use super::ffi::{
+    longjmp,
+    id,
+    nil,
+    CGRect,
+    CGPoint,
+    UIViewAutoresizingFlexibleWidth,
+    UIViewAutoresizingFlexibleHeight
+ };
+
+use super::jmpbuf;
+
+
+pub fn create_delegate_class() {
+    extern fn did_finish_launching(this: &mut Object, _: Sel, _: id, _: id) -> BOOL {
+        unsafe {
+            let main_screen: id = msg_send![Class::get("UIScreen").unwrap(), mainScreen];
+            let bounds: CGRect = msg_send![main_screen, bounds];
+            let window: id = msg_send![Class::get("UIWindow").unwrap(), alloc];
+            let window: id = msg_send![window, initWithFrame:bounds.clone()];
+
+            let size = (bounds.size.width as u32, bounds.size.height as u32);
+
+            let view_controller: id = msg_send![Class::get("MainViewController").unwrap(), alloc];
+            let view_controller: id = msg_send![view_controller, init];
+
+
+            let class = Class::get("MainView").unwrap();
+            let view:id = msg_send![class, alloc];
+            let view:id = msg_send![view, initForGl:&bounds];
+
+
+            let _: () = msg_send![view_controller, setView:view];
+
+
+            let _: () = msg_send![window, setRootViewController:view_controller];
+
+            let _: () = msg_send![window, addSubview:view];
+            let _: () = msg_send![window, makeKeyAndVisible];
+
+            let state = Box::new(DelegateState::new(window, view_controller, view, size));
+            let state_ptr: *mut DelegateState = mem::transmute(state);
+            this.set_ivar("glutinState", state_ptr as *mut libc::c_void);
+
+
+            let _: () = msg_send![this, performSelector:sel!(postLaunch:) withObject:nil afterDelay:0.0];
+        }
+        YES
+    }
+
+    extern fn post_launch(_: &Object, _: Sel, _: id) {
+        unsafe { longjmp(mem::transmute(&mut jmpbuf),1); }
+    }
+
+    extern fn did_become_active(this: &Object, _: Sel, _: id) {
+        unsafe {
+            let state: *mut libc::c_void = *this.get_ivar("glutinState");
+            let state = &mut *(state as *mut DelegateState);
+            state.events_queue.push_back(Event::Focused(true));
+        }
+    }
+
+    extern fn will_resign_active(this: &Object, _: Sel, _: id) {
+        unsafe {
+            let state: *mut libc::c_void = *this.get_ivar("glutinState");
+            let state = &mut *(state as *mut DelegateState);
+            state.events_queue.push_back(Event::Focused(false));
+        }
+    }
+
+    extern fn will_enter_foreground(this: &Object, _: Sel, _: id) {
+        unsafe {
+            let state: *mut libc::c_void = *this.get_ivar("glutinState");
+            let state = &mut *(state as *mut DelegateState);
+            state.events_queue.push_back(Event::Suspended(false));
+        }
+    }
+
+    extern fn did_enter_background(this: &Object, _: Sel, _: id) {
+        unsafe {
+            let state: *mut libc::c_void = *this.get_ivar("glutinState");
+            let state = &mut *(state as *mut DelegateState);
+            state.events_queue.push_back(Event::Suspended(true));
+        }
+    }
+
+    extern fn will_terminate(this: &Object, _: Sel, _: id) {
+        unsafe {
+            let state: *mut libc::c_void = *this.get_ivar("glutinState");
+            let state = &mut *(state as *mut DelegateState);
+            // push event to the front to garantee that we'll process it
+            // immidiatly after jump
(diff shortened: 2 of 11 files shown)


Cargo.toml
```diff
@@ -30,6 +30,9 @@ clock_ticks = "0.0.5"
 [target.arm-linux-androideabi.dependencies.android_glue]
 version = "0"
 
+[target.x86_64-apple-ios.dependencies]
+objc = "0.1"
```
tomaka: Is there no 32bits iOS? (naive question)

tomaka:
\o/

Cargo.toml
rozaliev: there are 5 different types there, this one is for simulator.
(x86_64-apple-ios, i386-apple-ios, aarch64-apple-ios, armv7s-apple-ios, armv7-apple-ios)

src/api/ios/mod.rs
```diff
@@ -0,0 +1,473 @@
...
+        // self.context.is_current()
+        unimplemented!()
+    }
+
+    fn get_proc_address(&self, addr: &str) -> *const libc::c_void {
+        let addr_c = ::std::ffi::CString::new(addr).unwrap();
+        let path = ::std::ffi::CString::new("/System/Library/Frameworks/OpenGLES.framework/OpenGLES").unwrap();
+        unsafe {
+            let lib = dlopen(path.as_ptr(), RTLD_LAZY | RTLD_GLOBAL);
+            dlsym(lib, addr_c.as_ptr())
```
tomaka: Would be better to cache the library instead of reopening it every time.
Also, I'd prefer `use std::ffi::CString` at the top, rather than absolute paths.

Cargo.toml
fkaa: Apple has pushed pretty hard with 64-bits architectures (there hasn't been any 32 bit apple hardware since ~2010 for OS X and iOS 8 only supports 64 bit)

src/events.rs
```diff
@@ -39,6 +39,10 @@ pub enum Event {
 
     /// The window needs to be redrawn.
     Refresh,
+
+
+    /// IOS event
+    WillEnterForeground,
```
tomaka: What about reusing the `Focus` event instead? Is it different?

tomaka:
> The last thing is EAGL, Apple's version of EGL. And it's somewhat confusing, because it requires to initialize render buffer and then pass it when swap buffer is called.

You mean an OpenGL render buffer?

rozaliev:
> You mean an OpenGL render buffer?
Yes

src/events.rs
rozaliev: I'm not sure, iOS app lifecycle is quite different. App never really quits, but have to unload resources on DidEnterBackground and load them back on EnterForeground.

There going to be more ios specific events, I wonder if they should be wrapped in IOS(IOSEvents) or smth.

rozaliev:
Status update:
Good news: glium's triangle example works
(https://www.dropbox.com/s/sqpndv4ueadprya/Screenshot%202015-05-28%2017.28.47.png?dl=0)

Bad news: textures don't. I think it's something to do with format detection, will investigate. It blocks me from proper testing.

mitchmindtree:
Nice work! Keep it up :smile:

tomaka:
@rozaliev Compatibility with OpenGL ES is on my to-do list. I have failing tests waiting to be fixed.
I'll try to work a bit on this.

tomaka:
@rozaliev Hmm, 2D textures work with my local OpenGL ES 2.0 configuration.
According to @fkaa, XCode should have debugging tools.

tomaka:
@rozaliev Ah, nevermind, doesn't work.

tomaka:
@rozaliev Fixed it in https://github.com/tomaka/glium/pull/916

rozaliev:
@tomaka still getting GL_INVALID_OPERATION after creating texture, but format seems to be ok now, thx

> XCode should have debugging tools.
it does, but to debug ES context it's required to run app on a real ios device, thus app have to be signed and i have my account still being in process of activation. So I'll have to wait a few days before i have proper tools.

rozaliev:
Ah, I forgot that textures width/height must be power of two. Everything works fine. https://www.dropbox.com/s/i17siojxzj2brzt/Screenshot%202015-05-29%2003.21.27.png?dl=0

Will add rest of the stuff and refactor tomorrow :)

rozaliev:
Added app lifecycle events and touches. The only major thing left is orientation change support.

Anyway, I believe anyone interested in testing this backend should probably start. I have my game running with it, but it might not be enough to catch all bugs.

Cargo.toml
```diff
@@ -30,6 +30,18 @@ clock_ticks = "0.0.5"
...
+
+[target.aarch64-apple-ios.dependencies]
+objc = "0.1"
+
+[target.armv7s-apple-ios.dependencies]
+objc = "0.1"
+
+[target.armv7-apple-ios.dependencies]
+objc = "0.1"
```
SSheldon: As a heads up, unfortunately I don't think objc will compile for these targets yet: https://github.com/SSheldon/rust-objc/blob/0.1.5/src/message.rs#L13-29

src/api/ios/delegate.rs
```diff
@@ -0,0 +1,223 @@
...
+            let touches_enum: id = msg_send![touches, objectEnumerator];
+
+            loop {
+                let touch: id = msg_send![touches_enum, nextObject];
+                if touch == nil {
+                    break
+                }
+                let location: CGPoint = msg_send![touch, locationInView:nil];
+                let touch_id = touch as u64;
+                let phase: i32 = msg_send![touch, phase];
```
SSheldon: `UITouchPhase` is typedef'd to `NSInteger`, so this should be an `NSInteger` rather than `i32`, right?

Cargo.toml
rozaliev: Yeah, so there should be smth like

```rust
#[cfg(target_arch="aarch64")]
fn msg_send_fn<R>() -> unsafe extern fn(*mut Object, Sel, ...) -> R {
    unsafe { mem::transmute(runtime::objc_msgSend) }
}
```

I have my fork with this changes and will PR once I have more time.

src/api/ios/delegate.rs
rozaliev: I might have misread the docs. This definition is somewhat confusing

```c
typedef enum {
   UITouchPhaseBegan,
   UITouchPhaseMoved,
   UITouchPhaseStationary,
   UITouchPhaseEnded,
   UITouchPhaseCancelled,
} UITouchPhase;
```

SSheldon: Hmm, odd, the definition I'm seeing in the header file as of iOS8 is:
``` objc
typedef NS_ENUM(NSInteger, UITouchPhase) {
    UITouchPhaseBegan,             // whenever a finger touches the surface.
    UITouchPhaseMoved,             // whenever a finger moves on the surface.
    UITouchPhaseStationary,        // whenever a finger is touching the surface but hasn't moved since the previous event.
    UITouchPhaseEnded,             // whenever a finger leaves the surface.
    UITouchPhaseCancelled,         // whenever a touch doesn't end but we need to stop tracking (e.g. putting device to face)
};
```

rozaliev: Could you point me to the location of this headers plz?

I'm getting mine from docs here: https://developer.apple.com/library/ios/documentation/UIKit/Reference/UITouch_Class/

SSheldon: I just jumped to it from Xcode, looks like the file for me is located at: `/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/UIKit.framework/Headers/UITouch.h`

rozaliev: Thx, same location for me. I'll fix it.

tomaka:
Are still working on this or can I merge it?
Even if it's buggy, it can't hurt to have at least minimal support.

rozaliev:
Unfortunately I've got very busy lately and won't have time to finish it properly, sorry. I'll try to get back to it in ~ month.

Current limitations:
 * it will only work on simulator and aarch64
 * support for armv7 will require addition of msg_send_struct! to rust-objc, otherwise it will crash every time you receive small struct like CGRect (on real device)
 *  There is no way for now to handle applicationDidEnterBackground message.  You can't setjmp out of it, because it have to return and you can't setjmp back. You can't simple return from it and handle msg later, because app is stuck in event loop.
 * There is no support for resizing\rotation
 * There is no support for custom screen size, it will always be native size of the device. 

I've not checked rust-objc lately, so maybe some issues have been fixed.

>Even if it's buggy, it can't hurt to have at least minimal support.
Agree, it may also inspire someone to contribute more features. So I i think that you can merge it.

SSheldon:
@rozaliev, the next version of the objc crate should support ARM and handle stret properly: SSheldon/rust-objc@8ffe610 :)

Still working on a way to test this functionality out on a device. Let me know if anything looks wrong in my intepretation of ARM's calling conventions.

PR #481, opened 2015-06-08
Added Control and Shift key events to win32 platform

tsubaki-mifune:
My OS (Windows 10) did not emit
```rust
winapi::LShift
winapi::RShift
winapi::LControl
winapi::RControl
```
but used the general
```rust
winapi::VK_SHIFT
winapi::VK_CONTROL
```
I added universal Shift / Control key to the VirtualKeyCode enum and added the necessary checks to
emit the KeyboardInput event.

I was only able to test this on Windows 10, thus I only added the new keys to the win32 platform.
But if other platforms also use the "direction invariant" shift and control keys it should be fairly easy to check for them.


tsubaki-mifune:
I think this has been fixed in #560

PR #483, opened 2015-06-13
Specify scroll deltas as either line or pixel-based

robertknight:
This PR implements @mrobinson's proposal at https://github.com/tomaka/glutin/pull/451#issuecomment-102311877 for representing scroll deltas in either lines/chunks/rows or pixels, depending on what is supported by the underlying platform and device. 
Windows and X11 currently report deltas in lines, OS X in pixels. X11 could support pixel deltas in future using the XInput2 API which is what GTK uses for high-res scroll events.

It is up to the app to interpret line-deltas appropriately. In Firefox for example, the maximum height of a frame's main font is used as the line height. Apple's documentation specifies that on devices that don't support 'precise' scrolling deltas, the delta should be interpreted as lines or rows (in a list of items). Windows documentation uses the slightly more vague 'chunks'.

This still needs tests. I'm posting the PR to get feedback on the proposed API and namings.

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -63,7 +63,7 @@ pub enum MouseScrollDelta {
 	/// and vertical directions.
 	///
 	/// Positive values indicate movement forward
-	/// (away from the user) or righwards.
+	/// (away from the user) or rightwards.
 	LineDelta(f32, f32),
 	/// Amount in pixels to scroll in the horizontal and
 	/// vertical direction.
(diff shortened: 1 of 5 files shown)


pcwalton:
+1, thanks for doing this!

tomaka:
Thanks!
So I guess this is the correct approach?

I'll wait for one or two days in case someone has something to say against this design, and if everything is fine I'll merge.

src/events.rs
```diff
@@ -25,11 +25,11 @@ pub enum Event {
     /// The parameter are the (x,y) coords in pixels relative to the top-left corner of the window.
     MouseMoved((i32, i32)),
 
-    /// Returns the horizontal and vertical mouse scrolling.
+    /// A mouse wheel or touchpad scroll occurred. Depending on whether the
```
samlh: Incomplete sentence?

PR #484, opened 2015-06-14
Implement transparency for X11

pinumbernumber:
I saw transparency was added for win32 recently, so here it is for X11 too. Information on transparent GLX windows is scarce and inconsistent- I was expecting to need the `GLX_TRANSPARENT_TYPE` attributes and such, but that didn't work at all- but this solution seems to work, and the `transparent` example now works on Linux.

### DIFF
--- a/src/api/x11/window.rs
+++ b/src/api/x11/window.rs
@@ -339,7 +339,23 @@ impl Window {
             if fb.is_null() {
                 return Err(OsError(format!("glx::ChooseFBConfig failed")));
             }
-            let preferred_fb = *fb;     // TODO: choose more wisely
+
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks!

But why does this need a 32bits depth buffer for the FBConfig? I don't really see the link with transparency.

pinumbernumber:
The `depth` member of `XVisualInfo` has the classical definition of "screen colour depth", as in all channels combined- R8G8B8A8 (TrueColor). Nothing to do with OpenGL depth buffers.

(For some reason setting the GLX transparency attributes doesn't work (it finds no appropriate Visuals), but omitting them and then only selecting 32-bit candidates does.)

PR #485, opened 2015-06-15
Make legacy functions crash on osx core contexts

fkaa:
Default behaviour is to just ignore the functions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ version = "0"
 
 [target.x86_64-apple-darwin.dependencies]
 objc = "0.1"
+cgl = "0"
 cocoa = "0"
 core-foundation = "0"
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -36,6 +36,9 @@ cocoa = "0"
 core-foundation = "0"
 core-graphics = "0"
 
+[target.x86_64-apple-darwin.dependencies.cgl]
+git = "https://github.com/servo/cgl-rs/"
```
tomaka: Dependencies should come from crates.io

fkaa: Completely forgot about that, fixed

tomaka:
And needs a rebase!

tomaka:
Thanks

Issue #52, opened 2014-10-12
Destroying a win32 window doesn't always stop the window thread


tomaka:
...which leads to the program continuing to run even when the `main` function has finished.

ethan-tqa:
here is how to prevent it from happening from within the application that is using glutin:

    while !window.is_closed() {
        let mut lol = window.wait_events();
		
		loop {
			match lol.next() {
				Some(x) => {
				},
				None => { break }
			}
		}

        // draw

        window.swap_buffers();
    }

Notice the loop, even though it does nothing here, it is calling next on the WaitEventsIterator object, which in turn processes the events, including Close. That is the point where is_closed will become true. Suggestion to fix this bug: have a special path to handle Close.

There is another thing that I noticed, but not sure about is that you are waiting for WM_DESTROY message. It appears that glfw uses WM_CLOSE. However i tried both and they are both caught when I close the window by pressing the X button.

tomaka:
Thanks.

In my opinion the best fix is to remove this `is_closed` flag and call `GetWindowPlacement` in the `is_closed()` function.

The difference between "destroy" and "close" is that close only hides the window. You can for example hide the window then show it again later, in which case you must only "close" it. Destroying is permanent. Anyway it doesn't really matter as glutin doesn't provide a way to do that.

ethan-tqa:
I'm afraid using GetWindowPlacement could potentially make the check expensive, as it usually placed in the game loop. The idea of using something like is_closed is good, it's just that it must be set correctly based on events. Another possible solution is to have one function that processes all events, which is to be called every frame, and let user provide a callback function. This is how glfw works. Another issue that has not been addressed is low latency input a.k.a raw input. Processing events per frame might cause input lags which is unacceptable in certain games

francesca64:
I'd be shocked if this is still relevant.

PR #488, opened 2015-06-16
Remove the is_closed function

### LINKED ISSUE #442: is_closed doesn't work on X11
Hi.

I think `window.is_closed` helper doesn't correctly work with Linux & X11: I compiled example from readme (https://github.com/tomaka/glutin#usage) and can't close glutin window as usual. I can correctly catch and process `Event::Closed` event with `poll_events()` though.

I'm using XMonad WM if that matters.

tomaka:
For the moment the behavior of `is_closed` is "returns true if a Closed event has been processed in the past", which is confusing and means that you can't use it without events processing.

I initially wanted to change it to "polls the backend to see if the user closed the window at some point", but this has several drawbacks:
 - It doesn't seem possible on X11.
 - It would require a precise definition of "closed". Should it return true if I call `window.hide()` for example?
 - People may want to use it as the sole mechanism to close their application. But this is not the right thing to do anyway.

So my proposition is to simply remove this function altogether and make events processing mandatory.

fix #442 
cc #52


tomaka:
cc @schulzch @ethan-tqa @Kagami

tomaka:
I also removed the vsync example, which is semi-hacky. It could be fixed to not use `is_closed`, but I don't want people to copy-paste it thinking it is a correct thing to do.

schulzch:
:heart:

PR #489, opened 2015-06-17
glx: Unbind the context if it's the current one

ecoal95:
Sorry if the comment was for a reason I'm not aware of.


tomaka:
Thanks!
The other problem is that a context can be binded in a thread different from the thread where the window is being destroyed, but we have no way to know that.

Anyway, I'm merging this as it can't hurt.

PR #493, opened 2015-06-21
Add dependencies lines for MSVC.

ricky26:
This adds yet another copy of the windows dependency lines - this time for the 64-bit MSVC target.

I've tested this with the first piston example (and I get the spinning red square) - so it seems like this is all that's needed.


tomaka:
Thanks!

PR #510, opened 2015-06-28
Use XInput2 for event handling

robertknight:
This changes the X11 input handling code to use XInput2.

This provides smooth per-pixel scrolling for touchpad devices and will
enable support for touch events and other enhancements in future.

Currently button press/release and motion event handling uses XI2, key press/release handling is still using X11 core events.

This depends on the upstream PR to x11-rs to add XInput2 support: https://github.com/Daggerbot/x11-rs/pull/16

### DIFF
--- a/src/api/x11/input.rs
+++ b/src/api/x11/input.rs
@@ -86,7 +86,9 @@ impl XInputEventHandler {
             ffi::XI_ButtonRelease,
             ffi::XI_Motion,
             ffi::XI_Enter,
-            ffi::XI_Leave
+            ffi::XI_Leave,
+            ffi::XI_FocusIn,
+            ffi::XI_FocusOut
         ];
         for event in events {
             ffi::XISetMask(&mut mask, *event);
@@ -159,7 +161,7 @@ impl XInputEventHandler {
     }
 
     pub fn translate_event(&mut self, cookie: &ffi::XGenericEventCookie) -> Option<Event> {
-        use events::Event::{MouseInput, MouseMoved, MouseWheel};
+        use events::Event::{Focused, MouseInput, MouseMoved, MouseWheel};
         use events::ElementState::{Pressed, Released};
         use events::MouseButton::{Left, Right, Middle};
         use events::MouseScrollDelta::{PixelDelta, LineDelta};
@@ -236,6 +238,8 @@ impl XInputEventHandler {
                 None
             },
             ffi::XI_Leave => None,
+            ffi::XI_FocusIn => Some(Focused(true)),
+            ffi::XI_FocusOut => Some(Focused(false)),
             _ => None
         }
     }
--- a/src/api/x11/input.rs
+++ b/src/api/x11/input.rs
@@ -1,0 +1,337 @@
+use std::sync::Arc;
+
+use libc;
+use std::{mem, ptr};
+use std::ffi::CString;
+use std::slice::from_raw_parts;
+
+use events::Event;
+
+use super::{events, ffi};
+use super::XConnection;
+
+#[derive(Debug)]
+enum AxisType {
+    HorizontalScroll,
+    VerticalScroll,
+    Other
+}
+
+#[derive(Debug)]
+struct Axis {
+    id: i32,
+    device_id: i32,
+    axis_number: i32,
+    axis_type: AxisType
+}
+
+#[derive(Debug)]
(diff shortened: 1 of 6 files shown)


tomaka:
\o/

tomaka:
Your PR for x11 has more chances to get accepted if you just implement xinput2 instead of changing the build system of the library at the same.

robertknight:
I'm happy to split the build system changes for x11-rs out into a separate PR. I'll give it a go and see if that helps make progress.

robertknight:
I split out the build system changes as suggested and XInput2 support has been merged upstream, though not yet published to crates.io

robertknight:
Hmm, AppVeyor failed on x86_64-pc-windows-msvc build when building the winapi-build dependency. That build looks new to me:

````
   Compiling winapi-build v0.1.0
     Running `rustc C:\Users\appveyor\.cargo\registry\src\github.com-0a35038f75765ae4\winapi-build-0.1.0\src\lib.rs --crate-name build --crate-type lib -g -C metadata=5e403cad040f7862 -C extra-filename=-5e403cad040f7862 --out-dir C:\projects\glutin\target\debug\deps --emit=dep-info,link -L dependency=C:\projects\glutin\target\debug\deps -L dependency=C:\projects\glutin\target\debug\deps -Awarnings`
Assertion failed: isa<X>(Val) && "cast<Ty>() argument of incompatible type!", file C:\bot\slave\nightly-dist-rustc-win-msvc-64\build\src\llvm\include\llvm/Support/Casting.h, line 237
````

**Comment by tomaka** [comment]
cc @retep998 :rabbit:

**Comment by retep998** [comment]
Oh come on. An LLVM assertion? Really? :rage:

**Comment by tomaka** [comment]
`if let Some(ev) = self.window.pending_events.lock().unwrap().pop_front()` should be moved inside the `loop` in `PollEventsIterator::next`, otherwise events that you add to the list will only be returned the next time.

File: src/api/x11/window.rs [hunk 1]
```diff
@@ -106,8 +108,64 @@ impl WindowProxy {
...
+        }
+    }
+}
+
 pub struct PollEventsIterator<'a> {
-    window: &'a Window,
+    window: &'a Window
+}
+
+impl<'a> PollEventsIterator<'a> {
```
tomaka: Please remove these two functions. It doesn't make sense for the iterator to implement them.

The first one can be inlined directly where you call it (there are only two calls). The second one should be inlined as well in my opinion, but I'd accept a freestanding function if you really don't want to inline it.

src/api/x11/input.rs
```diff
@@ -0,0 +1,328 @@
...
+    pub fn new(display: &Arc<XConnection>, window: ffi::Window, ic: ffi::XIC) -> XInputEventHandler {
+        // query XInput support
+        let mut opcode: libc::c_int = 0;
+        let mut event: libc::c_int = 0;
+        let mut error: libc::c_int = 0;
+        let xinput_str = CString::new("XInputExtension").unwrap();
+
+        unsafe {
+            if (display.xlib.XQueryExtension)(display.display, xinput_str.as_ptr(), &mut opcode, &mut event, &mut error) == ffi::False {
+                panic!("XInput not available")
```
tomaka: Is there a chance that this actually triggers?

src/api/x11/window.rs
robertknight: The piece of code that `queue_event()` is duplicated quite a few times in PollEventsIterator. I had intended to refactor the other uses to use queue_event() instead.

Coming from class-orientated languages I'm used to using private member/static helpers often to avoid repetition. I haven't written enough Rust to what the idiomatic equivalent is.

tomaka: Please inline.

src/api/x11/input.rs
robertknight: Honestly, I don't know if there are any systems out there with X11 but without XInput that we are likely to care about. XInput2 was introduced in 2009 as a standard part of X11. My inclination would be to panic and if it turns out that there is such a system somewhere refactor accordingly.

robertknight: This would also fail if there is an error connecting to the X server, although in that case setting up the X connection originally would almost certainly have failed.

tomaka:
Oh sorry, I forgot about this PR.

Issue #496, opened 2015-06-21
Consider adding a drag & drop event


thelink2012:
GLFW as of 3.1 provides a callback to queue dropped files and afaik SDL2 also has such event.

mitchmindtree:
Just an update: It looks like this event [has been added](https://github.com/tomaka/glutin/blob/master/src/events.rs#L15), but is only implemented for the `win32` api so far.

mitchmindtree:
I've been looking into how to get this working for cocoa.

In order to receive drag events, cocoa requires implementing the `NSDraggingDestination` protocol. It seems this involves subclassing `NSWindow` and then implementing the [`performDragOperation`](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Protocols/NSDraggingDestination_Protocol/#//apple_ref/occ/intfm/NSDraggingDestination/performDragOperation:) which gets called when a drag is released over the window. We could use this callback to add an event to the `VecDeque` of `pending_events`. I found a simple example of implementing the `NSDraggingDestination` protocol in obj c [here](http://stackoverflow.com/questions/8567348/accepting-dragged-files-on-a-cocoa-application).

Once implemented, we will have to register the window as a dragging destination for the specific types of files that we can want to handle (in our case, all of them). To do this, we have to call the [registerForDraggedTypes](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSWindow_Class/#//apple_ref/occ/instm/NSWindow/registerForDraggedTypes:) method on our window after we have created it and ensured that it is `non_nil`. This will probably look something like this:
```rust
// Register the window as a dragging destination for creating `DroppedFile` events.
msg_send![*window, registerForDraggedTypes:NSArray::arrayWithObject(nil, appkit::NSFilenamesPboardType)];
```

mitchmindtree:
The cocoa API supports handling multiple dropped files at once. When this occurs, we could either:
- emit multiple `DroppedFile` events or
- change `DroppedFile(PathBuf)` to `DroppedFile(Vec<PathBuf>)`

The second option might be useful for applications that need to know when multiple files are dropped at once though would be a breaking change. The first option is probably simplest to implement in the meantime.

thelink2012:
As a side note, Win32 also supports the event of many files at once (and that's what's used in my impl), however I still opted for triggering many events, seems less intrusive imo.

retep998:
[...]

PR #511, opened 2015-06-28
[0.3] DroppedFile event for Win32

thelink2012:
cc #496
**Breaking**, `Event` is not `Copy` anymore.


tomaka:
r+ but I'm going to leave this open for some time until we're ready to publish 0.3

PR #523, opened 2015-07-18
Split creating an EGL context in two parts

tomaka:
In order to properly create an X11 window, we need to know the visual ID of the pixel format chosen by EGL. So we split the EGL creation process in two parts.

`EglContext::new` returns a `EglContextPrototype`, which can be used to retreive the native visual ID. This prototype can then be turned into a real context by passing a window.


tomaka:
I checked on my machine, and this PR is indeed working on windows.

PR #527, opened 2015-07-19
Correct window creation on X11

tomaka:
Follow-up of #523 

Creating a window with X11 now uses the correct visual infos.


tomaka:
Confirmed working on windows.

PR #531, opened 2015-07-19
Add ToWide trait for marshaling wide strings

diaphore:
Nothing fancy but it reads better.


tomaka:
Sorry but I don't think it reads better.
When I see `OsStr::blabla` I know what it does. When I see to `to_wide` I have to go look at the trait or function definition.

In fact two of these calls should be changed to use `LoadLibraryA`.

diaphore:
Fair enough, I was thinking about factorizing those calls. Do you want me to close the PR ?

PR #535, opened 2015-07-22
Only listen for mouse events from master devices

### LINKED ISSUE #533: Mouse click events are sent twice on X11
cc @robertknight 

I don't actually know which events are received twice, but I noticed something wrong in my game after switching to the latest version of glutin.
Haven't investigated yet.

robertknight:
XInput2 has a concept of master and slave devices,
where a slave device is the actual physical device,
attached to a master device representing the cursor or keyboard
focus.

See http://who-t.blogspot.co.uk/2009/05/xi2-recipes-part-1.html

Mouse events were being received from both the master and slave
devices, but we are only interested in events from the master device.

Fixes #533


tomaka:
Thanks!

PR #538, opened 2015-07-22
Add aarch64-unknown-linux-gnu support

akiss77:
* Adding dependencies
* Replacing `i8` with `::libc::c_char` (since `c_char` can be
  unsigned on some platforms, aarch64 is one of them)
* Adding `extern crate libc;` to examples


akiss77:
Adapted from servo/glutin#38

tomaka:
Could you put `CStr::from_ptr(gl.GetString(gl::VERSION) as *const _)` instead and remove the `extern crate libc;` in the examples?

akiss77:
Much cleaner, thanks. Updated.

tomaka:
Thanks!

PR #541, opened 2015-07-25
Fixed docs for `get_inner_size`

bvssvni:
Closes https://github.com/tomaka/glutin/issues/388


tomaka:
I opened #542 as well

PR #546, opened 2015-07-25
Attempt to fix issues with sRGB

tomaka:


tomaka:
This doesn't fix the sRGB issue, but I'm going to merge it anyway.

PR #548, opened 2015-07-25
Add support for creating a pbuffer with EGL

tomaka:


tomaka:
I forgot something: the surface type must include the `pbuffer` bit.

PR #560, opened 2015-08-03
Modified Windows key event handling to support modifier keys.

### LINKED ISSUE #559: Control, shift and alt key events return None virtual key code on Windows
Other keys that do this include the tilde key and the minus and equals keys next to the number row.

Having the modifier keys return None is stopping cef-rs from being able to track modifiers and send key-combo commands (like Ctrl-A, etc.).

It would also be handy to be able to query the state of a key from Glutin, but that should probably be another issue.

dylanede:
This fixes #559 properly, with support for distinguishing left and right modifier keys. The modifications to do this are based on http://stackoverflow.com/a/15977613.

Unlike PR #481, this does not modify the virtual key code enum at all.

### DIFF
--- a/src/api/win32/event.rs
+++ b/src/api/win32/event.rs
@@ -1,8 +1,20 @@
 use events::VirtualKeyCode;
 use winapi;
+use user32;
+use ScanCode;
 
-pub fn vkeycode_to_element(code: winapi::WPARAM) -> Option<VirtualKeyCode> {
-    match code as i32 {
+const MAPVK_VSC_TO_VK_EX: u32 = 3;
+
+pub fn vkeycode_to_element(wparam: winapi::WPARAM, lparam: winapi::LPARAM) -> (ScanCode, Option<VirtualKeyCode>) {
+    let scancode = ((lparam >> 16) & 0xff) as u8;
+    let extended = (lparam & 0x01000000) != 0;
+    let vk = match wparam as i32 {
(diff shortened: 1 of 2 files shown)


src/api/win32/event.rs
```diff
@@ -1,8 +1,24 @@
 use events::VirtualKeyCode;
 use winapi;
+use ScanCode;
 
-pub fn vkeycode_to_element(code: winapi::WPARAM) -> Option<VirtualKeyCode> {
-    match code as i32 {
+#[link(name="user32")]
+extern "stdcall" {
+    fn MapVirtualKeyA(code: i32, map_type: i32) -> i32;
```
tomaka: You should use that function from the `user32` library.
There's already a `extern crate user32;` at the root, so just add `use user32;` and call `user32::MapVirtualKeyA`.

dylanede: MapVirtualKeyA isn't there, as far as I could see. There's a commented out stub for it.

tomaka:
r+ after the function call is fixed.

src/api/win32/event.rs
dylanede: Wait what, it's there now. Don't know what I was looking at.

dylanede: So, just for reference, this PR will wait until MapVirtualKeyA is in the user32-sys published to crates.io.

cybergeek94:
@tomaka @dylanede The latest version of `user32-sys` on Crates has this function now.

dylanede:
I've updated the commit and rebased. Should be good to go.

tomaka:
Thanks!

cybergeek94:
:+1:

PR #563, opened 2015-08-05
Removed dot in error message

bvssvni:
Closes https://github.com/tomaka/glutin/issues/562


tomaka:
That's really some nitpicking, but I'm ok with following the guidelines!

PR #567, opened 2015-08-10
Fix OpenGL profile selection

paulrouget:
This addresses 2 issues:

1. OpenGL 3.0 was not selecting NSOpenGLProfileVersionLegacy
2. because of a Rust bug, match behavior was selecting the wrong profile (this should be address by https://github.com/rust-lang/rust/pull/27050)


tomaka:
I don't think requesting a core profile with a legacy version should return an error.

paulrouget:
What profile to use in his case? `NSOpenGLProfileVersionLegacy`?

paulrouget:
Switched to legacy for 1.*, 2.*, 3.0 and 3.1.

tomaka:
r+ for me. Leaving open for a day or two in case someone more familiar with OS/X has something to say.

PR #580, opened 2015-08-24
Fix smooth-scrolling increment on some X11 devices

mbrubeck:
From the "Smooth Scrolling" section of [XI2Proto.txt][1]:

> One unit of scrolling in either direction is considered to be equivalent to one button event, e.g. for a unit size of 1.0, -2.0 on an valuator type Vertical sends two button press/release events for button 4. Likewise, a button press event for button 7 generates an event on the Horizontal valuator with a value of +1.0. The server may accumulate deltas of less than one unit of scrolling.

[1]: http://www.x.org/releases/X11R7.7/doc/inputproto/XI2proto.txt

This fixes scrolling with my Microsoft mouse in X11 on Debian 8.1.

### DIFF
--- a/src/api/x11/input.rs
+++ b/src/api/x11/input.rs
@@ -21,7 +21,8 @@ struct Axis {
     id: i32,
     device_id: i32,
     axis_number: i32,
-    axis_type: AxisType
+    axis_type: AxisType,
+    scroll_increment: f64,
 }
 
 #[derive(Debug)]
@@ -167,7 +168,7 @@ impl XInputEventHandler {
         use events::Event::{Focused, MouseInput, MouseMoved, MouseWheel};
         use events::ElementState::{Pressed, Released};
         use events::MouseButton::{Left, Right, Middle};
-        use events::MouseScrollDelta::{PixelDelta, LineDelta};
+        use events::MouseScrollDelta::LineDelta;
         use events::{Touch, TouchPhase};
 
         match cookie.evtype {
@@ -221,7 +222,7 @@ impl XInputEventHandler {
                 }
 
                 if scroll_delta.0.abs() > 0.0 || scroll_delta.1.abs() > 0.0 {
-                    Some(MouseWheel(PixelDelta(scroll_delta.0 as f32, scroll_delta.1 as f32)))
+                    Some(MouseWheel(LineDelta(scroll_delta.0 as f32, scroll_delta.1 as f32)))
                 } else {
                     let new_cursor_pos = (event_data.event_x, event_data.event_y);
                     if new_cursor_pos != self.current_state.cursor_pos {
@@ -290,7 +291,8 @@ fn read_input_axis_info(display: &Arc<XConnection>) -> Vec<Axis> {
                             ffi::XIScrollTypeHorizontal => AxisType::HorizontalScroll,
                             ffi::XIScrollTypeVertical => AxisType::VerticalScroll,
                             _ => { unreachable!() }
-                        }
+                        },
+                        scroll_increment: scroll_class.increment,
                     })
                 },
                 _ => {}
@@ -314,7 +316,7 @@ fn calc_scroll_deltas(event: &ffi::XIDeviceEvent,
             prev_axis.axis_number == axis_id
     });
     let delta = match prev_value_pos {
-        Some(idx) => axis_value - prev_axis_values[idx].value,
+        Some(idx) => prev_axis_values[idx].value - axis_value,
         None => 0.0
     };
 
@@ -335,8 +337,8 @@ fn calc_scroll_deltas(event: &ffi::XIDeviceEvent,
         if axis.id == event.sourceid &&
             axis.axis_number == axis_id {
                 match axis.axis_type {
-                    AxisType::HorizontalScroll => scroll_delta.0 = delta,
-                    AxisType::VerticalScroll => scroll_delta.1 = delta
+                    AxisType::HorizontalScroll => scroll_delta.0 = delta / axis.scroll_increment,
+                    AxisType::VerticalScroll => scroll_delta.1 = delta / axis.scroll_increment
                 }
             }
     }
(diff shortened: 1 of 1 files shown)


mbrubeck:
This isn't correct; we need to look at the "increment" field from XIScrollClassInfo.  Updated patch in progress...

mbrubeck:
Updated to work correctly with both my mouse (which has an increment of -1.0) and my trackpad (which has an increment of 114.0).  See the commit messages for details.

tomaka:
cc @robertknight

robertknight:
@mbrubeck - What app are you testing this with?

mbrubeck:
> What app are you testing this with?

Servo.

robertknight:
> Updated to work correctly with both my mouse (which has an increment of -1.0) and
> my trackpad (which has an increment of 114.0). See the commit messages for details.

Does this mean that your mouse uses the 'traditional' scrolling direction (mouse wheel towards you scrolls the screen down) and your touchpad is using 'natural' scrolling (drag finger towards you scrolls the screen up)?

robertknight:
Just tested the example with my own trackpad and indeed this inverts the scrolling direction, which doesn't match the way scrolling works on the rest of my system. In Ubuntu I have the 'natural scrolling' preference turned off, which I _think_ is the default setting, though I haven't verified this.

mbrubeck:
> Does this mean that your mouse uses the 'traditional' scrolling direction (mouse wheel towards you scrolls the screen down) and your touchpad is using 'natural' scrolling (drag finger towards you scrolls the screen up)?

My mouse and trackpad are both using the ‘traditional’ scrolling direction (moving the wheel or fingers towards me scrolls down, i.e., the same direction as pressing the “down” arrow key). With this patch, Servo works in the traditional direction, matching the other (non-Rust) apps on my system.  I'm running the MATE desktop (Gnome 2 fork) on Debian, which does not seem to offer a setting to invert the scrolling direction.

> Just tested the example with my own trackpad and indeed this inverts the scrolling direction, which doesn't match the way scrolling works on the rest of my system.

Interesting. Perhaps Servo is accidentally inverting the direction, and it went unnoticed because `increment` was ignored (causing the direction to be flipped twice on devices with negative `increment`).

src/api/x11/input.rs
```diff
@@ -270,7 +271,7 @@ fn read_input_axis_info(display: &Arc<XConnection>) -> Vec<Axis> {
     // only get events from the master devices which are 'attached'
```
robertknight: This comment is now out of sync with the code below.

src/api/x11/input.rs
```diff
@@ -167,7 +168,7 @@ impl XInputEventHandler {
         use events::Event::{Focused, MouseInput, MouseMoved, MouseWheel};
         use events::ElementState::{Pressed, Released};
         use events::MouseButton::{Left, Right, Middle};
-        use events::MouseScrollDelta::{PixelDelta, LineDelta};
```
robertknight: Of the three platforms, Windows and Linux are now using LineDelta and OS X is still using PixelDelta. I wonder if PixelDelta is still appropriate for OS X or whether it should be replaced there as well?

mbrubeck:
> Perhaps Servo is accidentally inverting the direction, and it went unnoticed because increment was ignored (causing the direction to be flipped twice on devices with negative increment).

This doesn't seem to be a problem in Servo; Servo behaves consistently with the platform expectations on MacOS X, for example, both when "natural" scrolling is enabled and disabled.

mbrubeck:
@robertknight, do you know if your test program behaves correctly on Mac OS X? Does it treat a positive *y* delta as "scroll down" or "scroll up"?  Servo treats a positive delta as "scroll up", and this is consistent with what Glutin sends on Mac.

robertknight:
> But without this patch, Glutin discards both events because read_input_axis_info
> does not return any matching axis for either event:

On my system which has only a trackpad, I do get the trackpad scroll edges reported when querying only the master devices. It looks like on systems with multiple inputs XIQueryDevice only associates the scrolling axis from one of them with the master pointer. In that case it makes sense to query all devices as you are doing in your patch. 

>  Servo treats a positive delta as "scroll up", and this is consistent with what Glutin sends on Mac.

Makes sense. IIRC in CoreGraphics coordinate system Y increases going up. On Windows, the WM_MOUSEWHEEL event behaves the same way. XInput2 uses the opposite direction (as shown by `xinput2 --test-xi2`), so makes sense that it needs to be inverted.

On that basis the only change needed in this PR is to adjust or remove the comment above the `XIQueryDevice` call argument change.

robertknight:
What was confusing me earlier is that the mouse and trackpad and scroll increments with different signs. When I attach a basic USB mouse to my laptop I see the same behavior as @mbrubeck . When I have no mouse attached, XIQueryDevice reports the trackpad's vertical/horizontal scroll axis' when querying only the master device. When I do have a mouse attached, XIQueryDevice reports only the USB mouse's vertical/horizontal scroll axis when querying only the master device.

For the mouse, rotating the scroll wheel forwards (away from me) results in the axis values reported by XInput increasing. For the trackpad, moving my finger forwards (away from me) results in the axis values reported by XInput decreasing. Since they have scroll increments with opposite signs, that counteracts the difference and they both behave the same way.

src/api/x11/input.rs
robertknight: Re-reading the Cocoa documentation for `[NSEvent scrollingDeltaY]` having `PixelDelta` still makes sense, since that matches Apple's guidance about how the app should behave depending on the value of `[NSEvent hasPreciseScrollingDeltas]`.

mbrubeck:
Fixed the comment.

mbrubeck:
@tomaka I believe this is ready to merge now.

tomaka:
Thanks @mbrubeck and @robertknight !

robertknight:
Thanks @tomaka ! . Looking forward to seeing this land in Servo. Firefox became much more pleasant to use when smooth scrolling was enabled under Linux recently.

PR #584, opened 2015-08-27
Fix android build.

glennw:


src/api/egl/mod.rs
```diff
@@ -41,6 +41,26 @@ pub struct Context {
...
 }
 
+#[cfg(target_os = "android")]
+fn is_platform_display_loaded(egl: &ffi::egl::Egl) -> bool {
+    true
+}
+
+#[cfg(target_os = "android")]
+fn is_platform_display_ext_loaded(egl: &ffi::egl::Egl) -> bool {
+    false
```
tomaka: Did you check if `GetPlatformDisplay` and `GetPlaformDisplayEXT` are/aren't available?
It seems unusual to me that the former would be supported and not the latter.

glennw: No - I don't have a device available today, so I used the safest option (since the ext version appears to never be used on android currently).

I can check tomorrow on a real device though and let you know.

glennw:
@tomaka Verified this runs the examples on a N7 2012.

tomaka:
Thanks

PR #585, opened 2015-08-28
Use HTTPS for doc hyperlink

frewsxcv:


tomaka:
What's the benefit of doing this?

frewsxcv:
Users get to their destination without threat of eavesdropping, MITM attacks, or content injection

tomaka:
I don't see the point, but why not.

PR #588, opened 2015-08-29
Make it work on DragonFly/FreeBSD

mneumann:
Tested on DragonFly BSD. Should work similarily on FreeBSD.


tomaka:
That's a lot of code duplication!

Is there a special reason why is wayland disabled? The .so is loaded at runtime, so in my opinion you could just use the same code as Linux.

mneumann:
I tried to avoid depending on the wayland libraries not knowing that they are portable. I have another pull request (https://github.com/tomaka/glutin/pull/589) which gets rid of all this duplicated code and squashes both commits together.

PR #590, opened 2015-08-30
No longer filter out non-conformant formats

### LINKED ISSUE #575: Enabling multisampling returns error when it shouldn't.
Only recently, when I use `with_multisampling` in my glutin/glium project I get the error `Couldn\'t find any pixel format that matches the criterias.'

This never used to happen before, and in fact, multisampling is supported (and worked) on the computer that I'm testing on!

tomaka:
Fix #575 

cc @TyOverby this should fix your problem


TyOverby:
Do you want me to patch this manually and see if it works, or is this scheme going to get merged?

PR #600, opened 2015-09-17
[cocoa] add resizable mask to decoration-less windows

paulrouget:
This is required to make window resizable and have a shadow.

Without and with the resizable mask:

screen shot 2015-09-17 at 09 04 08


paulrouget:
@tomaka I think the failure is unrelated. Can we merge this PR?

fkaa:
@paulrouget it is unrelated, yeah.

Issue #601, opened 2015-09-19
Panic XNotSupported running on Raspbian (Raspberry Pi)


pentagolo:
I cross-compiled the application

``` rust
#[macro_use]
extern crate glium;
fn main() {
    let display = glium::glutin::WindowBuilder::new().build_glium();
}
```

with the Cargo.toml

```
[dependencies]
glium = "*"
```

using a cross-compiler as explained at https://github.com/Ogeon/rust-on-raspberry-pi.
Notice that i do not call unwrap on the result, so the variable display is of type `std::result::Result<_, _>`.

On raspbian i started X, opened a terminal and started the application. The error

```
thread '<main>' panicked at 'callled `Result::unwrap()` on an `Err` value: XNotSupported', src/libcore/result.rs:731
```

was reported.

There is a function calling unwrap inside the glium::glutin lib thats causing the panic. I think this should be reported as an error-result instead.

But what can i do to make it run on the Raspberry Pi?

tomaka:
The changes in #602 will report an error instead of panicking and will indicate precisely what the problem is (instead of just `XNotSupported`).

tomaka:
The other part of the issue (the fact that it fails to load) is probably due to #554

tomaka:
Closing for https://github.com/tomaka/winit/issues/49

PR #602, opened 2015-09-20
[Breaking change] Better error handling in Linux's api dispatch

tomaka:
cc #601


tomaka:
This is a breaking change. It removes the `Clone`, `PartialEq` and `Eq` trait implementations from `CreationError`.

Issue #517, opened 2015-07-05
Allow the user to manually choose the pixel format from a list


tomaka:
Let's say that you want to use multisampling only if the color bits are equal to 32, but you would also accept a format with neither multisampling nor 32 color bits as a fallback.
This is currently not possible with glutin.

In addition to this, the `strict` system is unclear and a bit weird. The whole system is unprecise. You don't really know what it's doing under the hood.

It would be beneficial if the user had a way to manually choose from the list of available pixel formats.

tomaka:
I have a critical question: is it possible to enumerate the list of supported pixel formats on OS/X?

cc @fkaa @SSheldon

SSheldon:
@tomaka, sorry, I don't know anything about OpenGL! Did find this though:

> If the system can't find at least one pixel format that satisfies the constraints specified by the attribute array, it returns NULL for the pixel format object. In this case, your application should have an alternative that ensures it can obtain a valid object.

> One alternative is to set up your attribute array with the least restrictive attribute first and the most restrictive attribute last. Then, it is fairly easy to adjust the attribute list and make another request for a pixel format object.

From the [OpenGL Programming Guide for Mac](https://developer.apple.com/library/mac/documentation/GraphicsImaging/Conceptual/OpenGL-MacProgGuide/opengl_pixelformats/opengl_pixelformats.html), section "Ensuring a Valid Pixel Format Object".

tomaka:
I'm going to close this as "not possible".

Issue #516, opened 2015-07-05
Make the BuilderAttribs struct public


tomaka:
The `with_blabla` API seems a bit too magical to me.
I think it would be beneficial to show the user that it simply modifies some fields in the attributes, and that the user can manually modify them as well. It would also make it possible to expose the default values in a clean fashion.

However the struct would need some cleanup before being made public (it's not just adding `pub` in front of it). For example the `headless` member should disappear. I also don't want it to be possible to have an invalid state.

tomaka:
Done.

PR #603, opened 2015-09-20
Separate the builder attributes into multiple substates

tomaka:
This PR removes the `BuilderAttribs` struct and replaces it with `GlAttributes`, `WindowAttributes` and `PixelFormatRequirements`.

The purpose of this change is to avoid passing windowing-related attributes to WGL/GLX/EGL, and get a better isolation.

Small breaking change: the `HeadlessRendererBuilder` now takes a lifetime parameter to store the context to share lists with.

cc #516 
~~cc #517~~

### DIFF
--- a/src/api/android/mod.rs
+++ b/src/api/android/mod.rs
@@ -14,12 +14,14 @@ use events::MouseButton;
 use std::collections::VecDeque;
 
 use Api;
-use BuilderAttribs;
 use ContextError;
 use CursorState;
+use GlAttributes;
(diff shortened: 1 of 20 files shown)


tomaka:
Ah, I guess what I'm trying to do is not possible because of that cancer that is OS/X.

tomaka:
cc @glennw 

I'm not sure if this compiles for android (the changes were made blindly). I'm going to land this anyway and the compilation will eventually be fixed in master.

PR #605, opened 2015-09-20
Make all xlib libraries optional except xlib itself

### LINKED ISSUE #554: Support X11 systems that don't have extension libraries installed
libXcursor, libXxfvm86, libXi, etc. are not necessarly available everywhere.

### LINKED ISSUE #601: Panic XNotSupported running on Raspbian (Raspberry Pi)

tomaka:
Close #554 
Fix #601 

cc @robertknight (I'm touching your xinput2 code)

### DIFF
--- a/src/api/x11/input.rs
+++ b/src/api/x11/input.rs
@@ -51,57 +51,59 @@ pub struct XInputEventHandler {
 
 impl XInputEventHandler {
     pub fn new(display: &Arc<XConnection>, window: ffi::Window, ic: ffi::XIC) -> XInputEventHandler {
-        // query XInput support
-        let mut opcode: libc::c_int = 0;
-        let mut event: libc::c_int = 0;
-        let mut error: libc::c_int = 0;
-        let xinput_str = CString::new("XInputExtension").unwrap();
+        if let Some(xinput2) = display.xinput2.as_ref() {
+            // query XInput support
+            let mut opcode: libc::c_int = 0;
+            let mut event: libc::c_int = 0;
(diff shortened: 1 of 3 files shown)


robertknight:
As it stands, if XInput2 is not available you'll have keyboard input but no mouse input, since the mouse input code was replaced with logic that uses XInput.

I believe it is possible to have both XInput and X.org's core mouse handling live in the same app - if XInput registers mouse event handlers then mouse input results in XInput events, otherwise the app gets X.org core input mouse events. This would just be a matter of restoring the original mouse input code alongside the XInput code.

The downside is that you'd then have additional code for X11 which might not be exercised that often.

tomaka:
I just tried and there's no error on the raspi anymore. However context creation fails.

PR #607, opened 2015-09-21
Add #[inline] attributes

tomaka:


ozkriff:
Does this actually improve performance?

tomaka:
@ozkriff In theory yes. Rust doesn't inline functions between crates unless they are template or you put a `#[inline]` attribute.

PR #609, opened 2015-09-21
make glutin compile on all iOS targets (again)

TimNN:
This make glutin compile (again) for all iOS targets. I have not yet gotten around to actually compiling a test application, that will probably have to wait for next weekend.


tomaka:
Thanks!

PR #611, opened 2015-09-22
fix up win32 support

vvuk:
A few fixes for win32, needed for servo...

- user32-sys 0.1.2 is needed for MapVirtualKeyA
- wakeup_event_loop is implemented using a custom event
- two unimplemented!() calls were removed. the cursor setting one could be implemented pretty straightforwardly, and I'll probably get to it soon.. but that one in particular causes a crash with servo since it tries to set the cursor.


src/api/win32/callback.rs
```diff
@@ -281,6 +282,12 @@ pub unsafe extern "system" fn callback(window: winapi::HWND, msg: winapi::UINT,
             0
         },
 
+        x if x == wakeup_msg => {
```
tomaka: Can this be `if x == *super::WAKEUP_MSG_ID` to avoid the `wakeup_msg` local variable?

src/api/win32/mod.rs
```diff
@@ -218,7 +229,10 @@ impl Window {
     }
 
     pub fn set_cursor(&self, _cursor: MouseCursor) {
-        unimplemented!()
+        // XXX Ignore cursor changes for now on Windows, but don't crash; this
+        // should be implemented using SetCursor and LoadCursor after mapping _cursor
+        // to Windows cursors.
+        //unimplemented!()
```
tomaka: I don't like this change. It just fails silently and then users spend time trying to understand why things don't work.

vvuk:
Removed the local variable, and only removed the unimplemented!() from platform_display.  I agree about the cursor change; we can temporarily keep the patch to remove the unimplemented!() in the servo branch for now until we write a proper set_cursor().

tomaka:
Thanks!

I'll merge when CI passes.

PR #613, opened 2015-09-23
Fix Window.set_inner_size() on win32

mvdnes:
Previously, the function would actually set the *outer* size of the window instead of the inner size.
We fix this by first letting windows calculate the outer size based upon the specified inner size.

### DIFF
--- a/src/api/win32/mod.rs
+++ b/src/api/win32/mod.rs
@@ -182,8 +182,17 @@ impl Window {
         use libc;
 
         unsafe {
-            user32::SetWindowPos(self.window.0, ptr::null_mut(), 0, 0, x as libc::c_int,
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks!

mvdnes:
I see that you have published a new version (0.3.7) to crates.io. However, this fix seems not to be included. Have you published it without having the latest master locally?

tomaka:
@mvdnes Yes, master contains breaking changes already, so this change will be in version 0.4.0.
The version 0.3.7 was an urgent fix because of a problem of semver in gl_generator. I took version 0.3.6, applied the fix, and published.

mvdnes:
A ok no problem! It was just a bit confusing since there are no indicators in the git repo (that I am aware of).

PR #626, opened 2015-10-03
Remove the window feature entirely

tomaka:


vberger:
Just out of curiosity, what's the reason for this ?

tomaka:
The `window` and `headless` features existed so that it's not mandatory to link to xlib and osmesa. Now that they are loaded with dlopen, they are no longer necessary.

PR #636, opened 2015-10-11
ensure server has GLX_ARB_create_context extension

wartman4404:
Somehow, my machine (running Ubuntu 12.04) only has GLX_ARB_create_context on the client, which made all the examples fail with BadRequest when they tried to create a window.

### DIFF
--- a/src/api/glx/mod.rs
+++ b/src/api/glx/mod.rs
@@ -260,7 +260,7 @@ fn create_context(glx: &ffi::glx::Glx, extra_functions: &ffi::glx_extra::Glx, ex
                   -> Result<ffi::GLXContext, CreationError>
 {
     unsafe {
-        let context = if extra_functions.CreateContextAttribsARB.is_loaded() {
+        let context = if extensions.split(' ').find(|&i| i == "GLX_ARB_create_context").is_some() {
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks!

Could you remove `extra_functions.CreateContextAttribsARB.is_loaded()` entirely, and fix the alignment so that everything at the level of the `if` ([like here for example](https://github.com/tomaka/glium/blob/1eeb32209e19cd07d27bf6a0fdb34c1977df9408/src/buffer/alloc.rs#L1165-L1167))?

wartman4404:
It fits (barely) on one line if `extra_functions.CreateContextAttribsARB.is_loaded()` is removed.  But, I thought that call checked whether it was available on the client.  Is it okay to only check the server?

tomaka:
I don't know how extensions work when it comes to client and server, but the specs guarantee that the function is available if the list of extensions contains `GLX_ARB_create_context`.

On the other hand I have sometimes noticed that OpenGL implementations had empty functions which they didn't report in their extensions list. Checking whether a function is loaded is not sufficient.

wartman4404:
I see, thanks!  I didn't know how extensions worked in the specs.  Will this do?

tomaka:
Yeah, thanks

PR #638, opened 2015-10-14
git submodule was never commited

oli-obk:
the .gitmodules file was updated, but if the actual submodule isn't commited, `git submodule init` does nothing.

found the solution through stackoverflow: http://stackoverflow.com/a/9049703/1103681


tomaka:
Ah in fact it's the contrary. I forgot to remove `.gitmodules`.

oli-obk:
so the android tutorial at https://github.com/tomaka/glutin#android is outdated?

tomaka:
What is described still works.

However it doesn't really make sense to teach how to compile glutin itself for Android. You should learn how to compile your own project for android, and know that glutin supports Android.

oli-obk:
> You should learn how to compile your own project for android, and know that glutin supports Android.

Yea ^^ working on that...

Issue #543, opened 2015-07-25
Add a way of closing the "fullscreen" example


bvssvni:
It seems there is no way to close the "fullscreen" example without force quit on OSX.

Pressing Esc to exit would be nice.

tomaka:
Closing for https://github.com/tomaka/winit/issues/47

PR #645, opened 2015-10-25
Improve fullscreen example

adambadawy:
Two semantic changes here:
* The prompt to choose a monitor wasn't being flushed, and I fixed that.
* I added the ability to exit the example using <kbd>Esc</kbd>, as requested in issue #543.


tomaka:
Thanks

PR #646, opened 2015-10-26
Fix fullscreen on X11

adambadawy:
Two semantic changes:
* Use `vdisplay` again instead of `vsyncstart`
* Focus newly created windows

I know there was some reason you were using `vsyncstart` instead of `vdisplay`, but I couldn't figure it out from Googling around. (I have no idea what I'm doing.) It is of note, however, that Rust's X11 bindings shouldn't be missing `XF86VidModeModeInfo::hskew` anymore. (Daggerbot/x11-rs#19)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,39 +85,39 @@ osmesa-sys = "0.0.5"
 wayland-client = { version = "0.2.1", features = ["egl", "dlopen"] }
 wayland-kbd = "0.2.0"
 wayland-window = "0.1.0"
-x11-dl = "~2.0"
+x11-dl = "~2.2"
 
 [target.x86_64-unknown-linux-gnu.dependencies]
 osmesa-sys = "0.0.5"
 wayland-client = { version = "0.2.1", features = ["egl", "dlopen"] }
 wayland-kbd = "0.2.0"
 wayland-window = "0.1.0"
-x11-dl = "~2.0"
+x11-dl = "~2.2"
 
 [target.arm-unknown-linux-gnueabihf.dependencies]
 osmesa-sys = "0.0.5"
 wayland-client = { version = "0.2.1", features = ["egl", "dlopen"] }
 wayland-kbd = "0.2.0"
 wayland-window = "0.1.0"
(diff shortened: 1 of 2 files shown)


adambadawy:
After further testing, I found an issue with this. Not quite sure what to call it, but if I switch windows (e.g., via <kbd>Alt</kbd> + <kbd>Tab</kbd>):
* The Glutin fullscreen window doesn't appear in the window switching menu (Linux Mint 17.1 / Cinnamon)
* The Glutin fullscreen window doesn't go behind the newly focused window
* There is no way to switch back to the Glutin window

(You can probably still end the program by focusing your terminal and hitting <kbd>Ctrl</kbd> + <kbd>C</kbd>.)

adambadawy:
The error code is printed when trying to create a GL 3.2 context here:
https://github.com/tomaka/glutin/blob/master/src/api/glx/mod.rs#L164

The way that code is written, it seems like it should be fine that that's failing (since we go on to try creating a GL 3.1 context, which succeeds on my system.) Still investigating.

adambadawy:
Ok, so now I know what all of the code _I_ added does, and I've written some comments for my added lines.

Admittedly, I don't know how all of `Window::new` works, but there are definitely either (A) problems with my code or (B) parts of the code we can take out now. I'd appreciate some help with figuring this out.

As a place to start, we have an unused variable (`mode_to_switch_to`) and the blob I commented out (because I thought it was causing problems by bypassing the window manager).

adambadawy:
Ok, so I've checked the majority of the code everything's close to being in working order. There's one last caveat as far as I can tell: when a video mode other than the current desktop mode is used, the video mode isn't switched back upon exiting the program.

I've done quite a bit in the way of debugging this and I can't figure out why the call to `XF86VidModeSwitchToMode` in `XWindow::drop` is seemingly effectless. It _is_ being called and returning `1`. I know `1` is sometimes an error code in C, but I couldn't find docs on the return value of that function. For what it's worth, it's yielding the same value as the first, apparently successful call to `XF86VidModeSwitchToMode`.

PR #652, opened 2015-11-04
[Breaking change] Remove public exports of gl_common and libc

### LINKED ISSUE #650: Upgrade libc to 0.2
https://users.rust-lang.org/t/the-libc-crate-is-now-at-v0-2-0/3509

tomaka:
cc https://github.com/bjz/gl-rs/pull/359

Close #650

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,11 +13,11 @@ build = "build.rs"
 [dependencies]
 gl_common = "0.1.0"
 lazy_static = "0.1.10"
-libc = "0.1"
+libc = "0.2"
 shared_library = "0.1.0"
 
 [build-dependencies]
-gl_generator = "0.1.0"
+gl_generator = "0.2.0"
 khronos_api = "0.0.8"
 
 [dev-dependencies]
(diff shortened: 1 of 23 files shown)


SSheldon:
Removing `c_void` seems good, but I will mention that `*const ()` is not a C type, so you'll get a warning if you try to use it with an extern function.

I'm not sure if that's a concern here; if it it, the easiest way around it seems to be to make a no-variant enum and use that instead of `()`, like I did in SSheldon/rust-block@a1c5ef6.

tomaka:
After checking crates.io, it turns out that absolutely nobody uses that `load` thing, so I'm tempted to publish a minor version after this change instead of a major version.

mitchmindtree:
:pray:

mitchmindtree:
:tada:

SimonSapin:
Out of curiosity, why `()` rather than `std::os::raw::c_void`?

tomaka:
@SimonSapin Because I didn't know that existed -_-

PR #655, opened 2015-11-08
Lock objc to 0.1.7. Fix remaining libc pointer errors with casts. Fix warnings in cursor example.

mitchmindtree:
This fixes #654

PR #658, opened 2015-11-09
Update to libc 0.2, remove unnecessary gl_common crate

mitchmindtree:
I thought I'd open up another PR branch (an alternative to #655) that attempts to update to libc 0.2 and removes the gl_common dependency.

This build succeeds locally on OSX.

@tomaka I don't have a thorough understanding on what gl_common's original role was, however I noticed that you recently removed it from gl-rs, so I thought it might no longer have a place here - let me know if I was wrong and this should be done another way!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,11 +13,11 @@ build = "build.rs"
 [dependencies]
 gl_common = "0.1.0"
 lazy_static = "0.1.10"
-libc = "0.1"
+libc = "0.2"
 shared_library = "0.1.0"
 
 [build-dependencies]
-gl_generator = "0.1.0"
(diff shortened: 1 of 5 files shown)


tomaka:
Conflicts a bit with #652 

gl_common was meant to provide a way to replace `load_with(|s| window.get_proc_address(s))` with 
`load(window)`.

mitchmindtree:
Ahh, wasn't sure if you were still going ahead with #652 - perhaps I should close this and you can go ahead and do this within #652? It seems the libc update and gl_common fixes would need to happen in the same PR.

tomaka:
I merged your stuff in #652

PR #665, opened 2015-11-30
Fix compilation for 32 bit targets

antrik:
Regression introduced in 47df0e9eaa7f1231a07f86fc44426936c7a2589a

Casting fullscreen_atom (which is the result from XInternAtom, i.e.
c_ulong) as i64 is obviously wrong -- the whole point of types such as
c_ulong is that long in C does *not* always have the same bit size...

Cast it as c_long instead.

While this is the most straightforward fix, I'm not sure it's the best
one: perhaps the x11 crate should offer a set_ulong() method along with
set_long(), which could be used here instead of the cast?

### DIFF
--- a/src/api/x11/window.rs
+++ b/src/api/x11/window.rs
@@ -8,6 +8,7 @@ use std::cell::Cell;
 use std::sync::atomic::AtomicBool;
 use std::collections::VecDeque;
 use std::sync::{Arc, Mutex};
+use std::os::raw::c_long;
 
 use Api;
 use ContextError;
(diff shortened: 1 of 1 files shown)


src/api/x11/window.rs
```diff
@@ -535,11 +535,13 @@ impl Window {
                 message_type: state_atom, // the _NET_WM_STATE atom is sent to change the state of a window
                 format: 32,               // view `data` as `c_long`s
                 data: {
+                    use std::os::raw::c_long;
```
tomaka: Please move that line to the top of the module.

antrik:
I pushed an updated patch. Is it fine now?

tomaka:
Thanks!

PR #666, opened 2015-12-01
Update some libraries

SimonSapin:


bjz:
@SimonSapin build failiure? :/

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "glutin"
-version = "0.4.2"
+version = "0.4.3"
```
bjz: Not sure what @tomaka's policy regarding API changes, but this alters some public type signatures. Should it be a bump to v0.5.0?

adambadawy: Which types were changed? I only see fully-qualified names changed to imports.

SimonSapin:
Travis is now green, and I’ve push a blind attempt at fixing AppVeyor. (I don’t have a Windows dev machine)

SimonSapin:
Could someone on Windows take it from here?

Cargo.toml
bjz: They have moved from the `libc` definitions to those in `std::raw`.

adambadawy: Oops. I read too quickly.

tomaka: No public API has been changed as far as I can see.

tomaka:
I finished this in #667

ozkriff:
:metal:

Cargo.toml
bjz: Oh were those in private modules? My bad.

PR #668, opened 2015-12-05
Update the emscripten port of glutin

tomaka:
Unfortunately I'm not sure what the target triplet will end up being, so maybe not merge that immediately.


tomaka:
Turns out the triplet will be `asmjs-unknown-emscripten`, so this PR needs an update.

tomaka:
Waiting for https://github.com/rust-lang/rust/pull/30629

PR #673, opened 2015-12-11
Fix fullscreen on OS X

ahihi:
OS X fullscreen wasn't working, does now.


tomaka:
Thanks!

Since I never touched the cocoa code, I was hoping that someone else to review this.
Anyway, here we go.

PR #674, opened 2015-12-14
[WIP] Wayland backend rewrite

vberger:
This is still a WIP, don't merge just yet, I mainly still need to fix all the `unimplemented!()` in the outputs.

I still open it for some early feedback.

The diff should not be read as a whole, but rather on a per-commit basis, as it is a complete overhaul of `api/wayland`.


vberger:
Not sure about the appveyor fail... It does not seem related to my changes ?

tomaka:
Appveyor fails because of https://github.com/rust-lang/rust/issues/30229

vberger:
Okay, I *think* with this there should be a mostly functional wayland backend. At least for the classic uses.

PR #675, opened 2015-12-14
x11: Improve xlib error reporting

ecoal95:
This will show the X error string when an error happens, which should help to diagnose errors.

I re-opened `xlib` since:

 * This is not expected to be called a lot, so performance should not matter *that much*.
 * I didn't know what the best way to restructure of the current code will be.

I'll be happy to re-organize it if you wish.

### DIFF
--- a/src/api/x11/xdisplay.rs
+++ b/src/api/x11/xdisplay.rs
@@ -23,24 +23,18 @@ pub struct XConnection {
 unsafe impl Send for XConnection {}
 unsafe impl Sync for XConnection {}
 
+pub type XErrorHandler = Option<unsafe extern fn(*mut ffi::Display, *mut ffi::XErrorEvent) -> libc::c_int>;
+
 impl XConnection {
-    pub fn new() -> Result<XConnection, XNotSupported> {
+    pub fn new(error_handler: XErrorHandler) -> Result<XConnection, XNotSupported> {
         // opening the libraries
         let xlib = try!(ffi::Xlib::open());
         let xcursor = try!(ffi::Xcursor::open());
         let xf86vmode = try!(ffi::Xf86vmode::open());
         let xinput2 = try!(ffi::XInput2::open());
 
-        unsafe extern "C" fn x_error_callback(_: *mut ffi::Display, event: *mut ffi::XErrorEvent)
-                                              -> libc::c_int
-        {
-            println!("[glutin] x error code={} major={} minor={}!", (*event).error_code,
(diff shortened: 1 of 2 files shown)


vberger:
Calling `unwrap()` in `extern "C"` code is generally a bad idea, as panicking into C code is undefined behavior.

Probably matching the result of `Xlib::open()` and falling back to the old error message if it failed would be a better setup, even if `Xlib::open()` here is very unlikely to fail.

ecoal95:
I unwrapped there since it since a failure there was *really* unlikely to happen (it had been successfully opened previously).

Agreed though, it's easy to prevent. I updated the PR accordingly :)

tomaka:
From experience I have noticed that the `XGetErrorText` function is beyond useless. For example if you call it with the error code 52, it will return something like `"error 52"`.
I remember trying various error codes just to see if the function was working at all, and **all** error codes returned `"error #n"`.

ecoal95:
In my case it does report something kind of useful.

For example, on my machine with this patch `cargo run --example=multiwindow` yields:

```
$ cargo run --example=multiwindow
[glutin] x error code=8 major=42 minor=0: BadMatch (invalid parameter attributes)!
OpenGL version 3.3 (Core Profile) Mesa 11.2.0-devel (git-8cc372b 2015-12-10 wily-oibaf-ppa)
OpenGL version 3.3 (Core Profile) Mesa 11.2.0-devel (git-8cc372b 2015-12-10 wily-oibaf-ppa)
OpenGL version 3.3 (Core Profile) Mesa 11.2.0-devel (git-8cc372b 2015-12-10 wily-oibaf-ppa)
```

(I'm investigating what causes that error too)

ecoal95:
The errors I see are on `SetInputFocus`, so they're not dangerous I think (it just fails to switch the windows focus), am I right?.

tomaka:
I think that the fact that you reopen xlib is too hacky to be merged.
A better solution, I think, would be to move the callback function to [api_dispatch.rs](https://github.com/tomaka/glutin/blob/master/src/platform/linux/api_dispatch.rs) and read the existing xlib from the `BACKEND` variable.
The point to the callback function would then be passed as a parameter to `XConnection::new`.

ecoal95:
Agreed, just did it.

tomaka:
Thanks!

I'll leave this open in case someone has something to say about this design.
Otherwise I'll merge in ~24 hours. Please ping me if I forget.

PR #682, opened 2015-12-26
Fixed setting cursor and cursor state

### LINKED ISSUE #355: Cursors are broken on win32
For example putting your cursor on the border of the window doesn't bring the resizing arrows.

This is probably due to the interception of `WM_SETCURSOR`.

regnierof:
Also fixed the broken example 'grabbing'

Closes #355

### DIFF
--- a/examples/grabbing.rs
+++ b/examples/grabbing.rs
@@ -18,8 +18,8 @@ fn main() {
 
     let context = support::load(&window);
     let mut grabbed = false;
-    
-    for event in window.poll_events() {
+
+    for event in window.wait_events() {
         match event {
             Event::KeyboardInput(ElementState::Pressed, _, _) => {
                 if grabbed {
@@ -33,6 +33,8 @@ fn main() {
                 }
             },
 
+            Event::Closed => break,
+
             a @ Event::MouseMoved(_) => {
                 println!("{:?}", a);
             },
@@ -44,4 +46,3 @@ fn main() {
         let _ = window.swap_buffers();
     }
 }
-
--- a/src/api/win32/mod.rs
+++ b/src/api/win32/mod.rs
@@ -42,9 +42,13 @@ lazy_static! {
     static ref WAKEUP_MSG_ID: u32 = unsafe { user32::RegisterWindowMessageA("Glutin::EventID".as_ptr() as *const i8) };
 }
 
+/// Cursor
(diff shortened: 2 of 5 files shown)


examples/grabbing.rs
```diff
@@ -18,8 +18,8 @@ fn main() {
 
     let context = support::load(&window);
     let mut grabbed = false;
-    
-    for event in window.poll_events() {
+
+    for event in window.wait_events() {
```
tomaka: What's the purpose of this change? Using `wait_events` means that the window won't be redrawn at each frame.

src/api/win32/mod.rs
```diff
@@ -261,41 +265,55 @@ impl Window {
...
+            MouseCursor::EwResize | MouseCursor::ColResize => winapi::IDC_SIZEWE,
+            MouseCursor::NsResize | MouseCursor::RowResize => winapi::IDC_SIZENS,
+            MouseCursor::Wait | MouseCursor::Progress => winapi::IDC_WAIT,
+            MouseCursor::Help => winapi::IDC_HELP,
+            _ => winapi::IDC_ARROW, // use arrow for the missing cases.
+        };
+
+        unsafe {
+            let mut cur = self.window_state.lock().unwrap();
+            cur.cursor = mem::transmute(cursor_id);
```
tomaka: Could you use the right type for `cur.cursor` and remove this transmute? (I guess `*const winapi::wchar_t`).
Transmutes are the last resort because it's very error-prone.

tomaka:
Thanks!
It looks good other than the two small nitpicks.

examples/grabbing.rs
regnierof: I changed it because when I call poll_events() on my windows machine, the window disappears after being on the screen for a fraction of a second.

I thought it was a bug, so I changed it to wait_events(), which makes the window stick on the screen.

src/api/win32/mod.rs
regnierof: Ah, the transmute was a silly work around. I was supposed to change it before I sent the PR but I forgot about it.

examples/grabbing.rs
tomaka: Ahhh you're right. I didn't notice that there was no `loop {}` around the thing.
Yeah let's fix that with `wait_events`.

PR #683, opened 2015-12-26
win fix: min/max dimensions now correctly use client area size

### LINKED ISSUE #681: resizability & min/max dimensions
There didn't seem to be a resizable flag so I tried setting the min and max dimension to the same values as the window dimensions. This didn't seem to give the desired effect. 
with_min_dimensions() & with_max_dimensions() set the dimension _including_ the window border. I don't think this is the desired behavior for most users. 

platform: windows 7

```
    let window = glutin::WindowBuilder::new()
        .with_title("Title".to_string())
        .with_dimensions(720, 480) // set the size excluding the border
        .with_min_dimensions(720, 480) // sets the size including the border
        .with_max_dimensions(720, 480) // sets the size including the border
        .build_glium()
        .unwrap();

```

SenneH:
Fixed #681.

### DIFF
--- a/src/api/android/mod.rs
+++ b/src/api/android/mod.rs
@@ -123,7 +123,7 @@ impl Window {
         // not implemented
         assert!(win_attribs.min_dimensions.is_none());
         assert!(win_attribs.max_dimensions.is_none());
-        assert!(window.resizable);
+        assert!(win_attribs.resizable);
 
         let opengl = opengl.clone().map_sharing(|w| &w.context);
 
--- a/src/api/cocoa/mod.rs
+++ b/src/api/cocoa/mod.rs
@@ -272,7 +272,7 @@ impl Window {
         // not implemented
         assert!(win_attribs.min_dimensions.is_none());
         assert!(win_attribs.max_dimensions.is_none());
-        assert!(window.resizable);
+        assert!(win_attribs.resizable);
 
         match opengl.robustness {
             Robustness::RobustNoResetNotification | Robustness::RobustLoseContextOnReset => {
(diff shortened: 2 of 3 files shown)


Sergesosio:
I also fixed the issue in #355.

tomaka:
Thanks!

However issue #355 is already being fixed by PR #682.

Also, since you added the `resizable` flag, please add an `assert!(window.resizable);` in `api/glx/window.rs`, `api/android` and `api/cocoa`.

Sergesosio:
#682 doesn't fix #355. It only adds functionality to change the cursor icon to another. The issue about the cursor not changing into to resize icon when hovering the border of the window is still there. 
I could make it so that it correctly uses the default behavior if no cursor is set?

tomaka:
Looks good!

However why do you need to add a `reset_cursor` function instead of just setting the value to `None` if the user calls `set_cursor(Default)`?

Also you need to add `assert!(window.resizable);` before I merge.

Sergesosio:
You're right, having a ```resert_cursor``` function is unnecessary. ```set_cursor``` should now accept an ```Option<>```

As for ```assert!(window.resizable);```, I added it in ```api/android``` and ```api/cocoa```. But I don't have the file ```api/glx/window.rs```

tomaka:
I meant `api/x11/window.rs` instead of `glx`.

For the cursor, please don't change it to an `Option`. There a `MouseCursor::Default` variant that is supposed to mean "the default cursor".

francesca64:
It looks like `resizable` is the only part of this that would still be relevant, and I doubt it would be worth the effort to try to rebase this for winit.

francesca64:
and now `resizable` is being covered over in tomaka/winit#540

Issue #664, opened 2015-11-25
unnecessary parentheses warning on OSX


frewsxcv:
```
src/api/cocoa/mod.rs:369:34: 369:44 warning: unnecessary parentheses around `for` head expression, #[warn(unused_parens)] on by default
src/api/cocoa/mod.rs:369                         for i in (0..count) {
                                                          ^~~~~~~~~~
```

fkaa:
Fixed in #684

PR #684, opened 2015-12-28
Implement set_cursor_position on OSX with Cocoa

AnthIste:
The implementation uses [CGWarpMouseCursorPosition](https://developer.apple.com/library/mac/documentation/GraphicsImaging/Reference/Quartz_Services_Ref/index.html#//apple_ref/c/func/CGWarpMouseCursorPosition) that accepts global screen coordinates. These are therefore offset by the current window coordinates.

### DIFF
--- a/src/api/cocoa/mod.rs
+++ b/src/api/cocoa/mod.rs
@@ -34,7 +34,7 @@ use core_foundation::base::TCFType;
 use core_foundation::string::CFString;
 use core_foundation::bundle::{CFBundleGetBundleWithIdentifier, CFBundleGetFunctionPointerForName};
 
-use core_graphics::display::{CGAssociateMouseAndMouseCursorPosition, CGMainDisplayID, CGDisplayPixelsHigh};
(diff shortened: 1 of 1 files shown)


fkaa:
Looks good to me. Have you noticed any delay as mentioned in the original issue? Also seems to fix #664

AnthIste:
I haven't personally noticed any delay, but I haven't done extensive testing. I see on [this stackoverflow answer](http://stackoverflow.com/a/17559012/761648) a workaround is to call `CGAssociateMouseAndMouseCursorPosition(true)` immediately after warping the cursor. [This post](https://groups.google.com/forum/#!msg/wx-dev/dkKkq2v929U/XoOMu4yPniUJ) also suggests that it is a valid (if undocumented) fix.

I'll update the code to include this and test locally. I can't tell if it fixes any delay because I am not experiencing any, but I will be able to see if it causes problems.

AnthIste:
Adding in the extra call actually made a huge difference in cursor responsiveness :)

I based this change on the current stable crates.io version but I think it should be OK to merge into master. When is the next crates.io release scheduled? I need this functionality so I am currently referencing my own local bugfix version.

tomaka:
@AnthIste Just submit a PR which bumps the version in the Cargo.toml

PR #685, opened 2015-12-29
Publish 0.4.5

AnthIste:
I added a `0.4.5` tag to this commit but this will probably get lost during the merge seeing as the tag is against my branch commit and not the merge commit?


tomaka:
You don't have the rights to push tags in theory.

PR #686, opened 2015-12-29
Add a useless call to `glXQueryVersion()` before doing anything in the GLX port to work around a bug in the VirtualBox OpenGL driver.

pcwalton:


tomaka:
Wow!

Thanks

PR #687, opened 2015-12-29
Support for side buttons on mouse

Aceeri:
Adds XButton support on windows.


tomaka:
Thanks!

I don't even know what a XButton is!

PR #688, opened 2015-12-30
Adds ability to use icons on windows

Aceeri:
Modifies the tray and title bar icons as long as the path to the file is found. Should probably add some warnings if the file is not found.

### DIFF
--- a/src/api/win32/init.rs
+++ b/src/api/win32/init.rs
@@ -239,19 +239,16 @@ unsafe fn init(title: Vec<u16>, window: &WindowAttributes, pf_reqs: &PixelFormat
         rx
     };
 
-    match window.icon {
-        Some(ref icon) => {
-            let list: Vec<u16> = icon.as_os_str().encode_wide().chain(Some(0)).collect();
-            let h_icon = user32::LoadImageW(ptr::null_mut(), list.as_ptr(), winapi::IMAGE_ICON, 0, 0, winapi::LR_DEFAULTSIZE | winapi::LR_LOADFROMFILE);
+    if let Some(ref icon) = window.icon {
+        let list: Vec<u16> = icon.as_os_str().encode_wide().chain(Some(0)).collect();
+        let h_icon = user32::LoadImageW(ptr::null_mut(), list.as_ptr(), winapi::IMAGE_ICON, 0, 0, winapi::LR_DEFAULTSIZE | winapi::LR_LOADFROMFILE);
 
-            if h_icon.is_null() {
-                return Err(CreationError::FileNotFound);
-            }
+        if h_icon.is_null() {
+            return Err(CreationError::FileNotFound);
+        }
 
-            user32::PostMessageW(real_window.0, winapi::WM_SETICON, 0, h_icon as winapi::LPARAM);
-            user32::PostMessageW(real_window.0, winapi::WM_SETICON, 1, h_icon as winapi::LPARAM);
-        },
-        None => { }
+        user32::PostMessageW(real_window.0, winapi::WM_SETICON, 0, h_icon as winapi::LPARAM);
+        user32::PostMessageW(real_window.0, winapi::WM_SETICON, 1, h_icon as winapi::LPARAM);
     }
 
     // building the struct
--- a/src/api/win32/init.rs
+++ b/src/api/win32/init.rs
@@ -20,7 +20,7 @@ use GlRequest;
 use PixelFormatRequirements;
 use WindowAttributes;
 
-use std::ffi::{OsStr};
+use std::ffi::{OsStr, CString};
 use std::os::windows::ffi::OsStrExt;
 use std::sync::mpsc::channel;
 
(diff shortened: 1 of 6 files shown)


tomaka:
That's not really a good design in my opinion.
The `icon` member should be an `Option<String>`, and an error should be returned if the icon is not found.

Also, you should put `if icon.is_some() { unimplemented!() }` in the Android, Cocoa and X11 codes.

src/api/win32/init.rs
```diff
@@ -237,6 +237,15 @@ unsafe fn init(title: Vec<u16>, window: &WindowAttributes, pf_reqs: &PixelFormat
         rx
     };
 
+    let path = CString::new(window.icon.clone()).unwrap().as_ptr();
+    let h_icon = user32::LoadImageA(ptr::null_mut(), path, winapi::IMAGE_ICON, 0, 0, winapi::LR_DEFAULTSIZE | winapi::LR_LOADFROMFILE);
```
retep998: Grooooossss. You're using the A version of things. That means it'll completely disregard unicode!
Use `LoadImageW` with `OsStrExt::encode_wide`.

src/api/win32/init.rs
```diff
@@ -237,6 +237,15 @@ unsafe fn init(title: Vec<u16>, window: &WindowAttributes, pf_reqs: &PixelFormat
...
+
+    // Won't accept an isize for some reason
+    #[cfg(target_pointer_width="32")] type Handle = i32;
+    #[cfg(target_pointer_width="64")] type Handle = i64;
+    user32::PostMessageW(real_window.0, winapi::WM_SETICON, 0, h_icon as Handle);
+    user32::PostMessageW(real_window.0, winapi::WM_SETICON, 1, h_icon as Handle);
```
retep998: No need to create a new type alias here, just cast it to `LPARAM` since that is the type it takes.

src/window.rs
```diff
@@ -78,6 +78,13 @@ impl<'a> WindowBuilder<'a> {
         self
     }
 
+    /// Sets the path to the icon of the window.
```
retep998: If this is a path, shouldn't it be taking an actual `Path`? Or perhaps a `T: AsRef<Path>`. Gotta remember that not all paths are proper unicode.

Aceeri:
Whoops, didn't know the difference between the W and A types sorry.

@tomaka: Should the error just be a straight up Err ("")? I noticed earlier that you were adding some XError type or something.

tomaka:
Just add a new entry in the `CreationError` enum.

src/api/win32/init.rs
```diff
@@ -94,6 +94,8 @@ unsafe fn init(title: Vec<u16>, window: &WindowAttributes, pf_reqs: &PixelFormat
                opengl: &GlAttributes<RawContext>, egl: Option<Egl>)
                -> Result<Window, CreationError>
 {
+    #[cfg(windows)] use std::os::windows::ffi::OsStrExt;
```
tomaka: This is unnecessary as the trait is already imported at the top.

src/lib.rs
```diff
@@ -143,6 +144,7 @@ pub enum CreationError {
     RobustnessNotSupported,
     OpenGlVersionNotSupported,
     NoAvailablePixelFormat,
+    FileNotFound,
```
tomaka: I'd prefer `IconFileNotFound` so that the user knows why the error happens.

src/window.rs
```diff
@@ -78,6 +79,13 @@ impl<'a> WindowBuilder<'a> {
         self
     }
 
+    /// Sets the path to the icon of the window. Currently only works with .ico file types, other file types will result in a FileNotFound creation error.
```
tomaka: I'm a bit reluctant with this because you can't ask a Linux user that it must pass an .ico file.

src/lib.rs
Aceeri: What if instead of creating another CreationError, we include a variable to the CreationError for extensibility? So then we don't have to create a new CreationError whenever you need to load a new file.

src/window.rs
Aceeri: Yeah... this might be a bit difficult, windows uses .ico files, while macos uses .icns files, x11 uses some sort of Cardinal ARGB object, and I'm not sure what wayland does.

vberger: I don't think wayland support setting such icons at all (for now at least).

But when (if ?) it does, it'll most like be the responsibility of the client (aka glutin) to load the image to a in-memory buffer and provide it to the wayland compositor, like all other wayland surfaces.

fkaa: @Aceeri It should be noted that setting the icon on OS X is done entirely in the distribution structure. In `Program.app/Contents/` there would be a `Ìnfo.plist` file containing some data about the bundle. The following entry would set the icon for the application:
````XML
<key>CFBundleIconFile</key>
<string>iconfile</string>
```
where `iconfile` will point to `iconfile.icns` in the `Resources` folder.

tomaka: I'm going to add a API for platform-specific configuration. Setting the icon will be put in there.

Aceeri: I'm thinking this may be better handled once we have a simpler way of linking resources to an application.

PR #690, opened 2015-12-30
Fix EGL for wayland.

vberger:
These are the changes I managed to get EGL context initialization to work on wayland on my computer.

I don't really know what all that implies, it was mostly try and retry.

Your thoughts @tomaka ?

### DIFF
--- a/src/api/egl/mod.rs
+++ b/src/api/egl/mod.rs
@@ -466,10 +466,10 @@ unsafe fn choose_fbconfig(egl: &ffi::egl::Egl, display: ffi::egl::types::EGLDisp
             out.push(ffi::egl::RGB_BUFFER as c_int);
         }
 
-        if egl_version >= &(1, 3) {
-            out.push(ffi::egl::SURFACE_TYPE as c_int);
-            out.push((ffi::egl::WINDOW_BIT | ffi::egl::PBUFFER_BIT) as c_int);
-        }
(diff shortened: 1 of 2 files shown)


tomaka:
Do you get an error with `EGL_SURFACE_TYPE`?

vberger:
If I don't comment out the `SURFACE_TYPE` block, `eglChooseConfig` triggers a `EGL_BAD_ATTRIBUTE`.

No idea what's causing this, maybe a bug in my version of Mesa? Could be nice if someone else coult test as well.

tomaka:
Well, I noticed an error in that `SURFACE_TYPE`, `PBUFFER_BIT` and `WINDOW_BIT` are supported even by EGL 1.0. So that `if` could be removed.

tomaka:
@vberger What happens if you leave only `WINDOW_BIT`?

vberger:
@tomaka cannot test it right now, will do it later in the afternoon.

However, if I remember correctly what I read in the EGL spec, having only `WINDOW_BIT` is the default value for this attribute.

tomaka:
> having only WINDOW_BIT is the default value for this attribute.

That's why I'm asking what happens.

vberger:
By removing `PBUFFER_BIT` it works as well.

tomaka:
Please add a comment in the code explaining the situation, then I'll merge.

PR #694, opened 2016-01-02
Updating the GLX API to use the latest GL version

rhoot:
Previously when using GlRequest::Latest, it would first try to use 3.2, then 3.1, and finally 1.0. Meaning any other version was unavailable (including newer versions).

Instead, by not specifying a specific version, whichever version is actually latest will be used. This is also what glfw does (see https://github.com/glfw/glfw/blob/master/src/glx_context.c#L428).

The specific issue that prompted this change is that the proprietary nvidia drivers do not seem to expose any extensions for OpenGL 3.2 for whatever reason, but it does for the latest version (4.5).

### DIFF
--- a/src/api/glx/mod.rs
+++ b/src/api/glx/mod.rs
@@ -177,26 +177,10 @@ impl<'a> ContextPrototype<'a> {
         // creating GL context
         let context = match self.opengl.version {
             GlRequest::Latest => {
-                if let Ok(ctxt) = create_context(&self.glx, &extra_functions, &self.extensions, (3, 2),
-                                                 self.opengl.profile, self.opengl.debug,
-                                                 self.opengl.robustness, share,
-                                                 self.display, self.fb_config, &self.visual_infos)
-                {
-                    ctxt
-                } else if let Ok(ctxt) = create_context(&self.glx, &extra_functions, &self.extensions,
-                                                        (3, 1), self.opengl.profile,
-                                                        self.opengl.debug,
-                                                        self.opengl.robustness, share, self.display,
-                                                        self.fb_config, &self.visual_infos)
-                {
-                    ctxt
-
-                } else {
-                    try!(create_context(&self.glx, &extra_functions, &self.extensions, (1, 0),
-                                        self.opengl.profile, self.opengl.debug,
-                                        self.opengl.robustness,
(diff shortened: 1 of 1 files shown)


tomaka:
That's a tricky question. I did some researchs, and apparently the problem is that nVidia drivers are non-compliant. Your PR would fix the behavior nVidia drivers but would break it for compliant drivers.

https://github.com/mosra/magnum/issues/80#issuecomment-145958456

In my opinion a better fix would be to try each version one by one (4.5, then 4.4, 4.3, and so on).

rhoot:
Hm, it seems you may be correct. Trying all the versions sounds like a pain though.

Maybe a better fix is just to do what both glfw and waffle does? If 1.0 was requested to not specify it (since it's the default anyway), and to cause `GlRequest::Latest` to request 1.0?

rhoot:
Actually, the effect would be the same as with my current change come to think of it. And I'm not sure how it would fail with compliant drivers. It just causes `GLX_CONTEXT_MAJOR_VERSION_ARB` and `GLX_CONTEXT_MINOR_VERSION_ARB` to be non-specified when using `GlRequest::Latest`, which would cause them to default to 1.0.

I guess the question is if there are any drivers where that's a problem.

tomaka:
If you don't specify the version, MESA will give you 2.1 or 3.1 and not the latest version (which is the right behavior when you don't support compatibility).

rhoot:
Actually, the issue seems to be on my end. I mistakenly thought the library I was using supported more recent OpenGL versions than it did. I'll just fix this on my end by specifying the version it supports. Thanks for the insight nonetheless.

For reference, versions requested vs versions handed back by the nvidia drivers:

* 4.5: 4.5
* 4.4: 4.4
* 4.3: 4.3
* 4.2: 4.2
* 4.1: 4.1
* 4.0: 4.0
* 3.3: 3.3
* 3.2: 3.2
* 3.1: 3.1
* 3.0: 4.5
* 2.1: 4.5
* 2.0: 4.5
* 1.5: 4.5
* 1.4: 4.5
* 1.3: 4.5
* 1.2: 4.5
* 1.1: 4.5
* 1.0: 4.5

tomaka:
I opened #695 because that's still a problem.

PR #698, opened 2016-01-05
Support HiDPI on Linux by querying the Xft DPI through XSETTINGS.

pcwalton:
The Xft DPI is set when the user adjusts the scale in the Ubuntu
"Displays" preference pane.

This depends on `rust-xsettings`, which in turn uses the C library
`libxsettings-client-0`. Unfortunately, this library is not installed by
default on most Linux distributions. To avoid unnecessary dependencies,
this commit places support for HiDPI on Linux behind the `use-xsettings`
Cargo feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,10 @@ repository = "https://github.com/tomaka/glutin"
 documentation = "https://tomaka.github.io/glutin/"
 build = "build.rs"
 
+[features]
+default = []
+use-xsettings = ["xsettings"]
+
 [dependencies]
 lazy_static = "0.1.10"
 libc = "0.2"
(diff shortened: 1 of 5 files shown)


pcwalton:
I'm happy to adjust how the feature works if you'd like.

Cargo.toml
```diff
@@ -86,6 +90,10 @@ wayland-kbd = "0.3.3"
 wayland-window = "0.2.2"
 x11-dl = "~2.3"
 
+[target.i686-unknown-linux-gnu.dependencies.xsettings]
```
SimonSapin: Nit: could you format this as `xsettings = {…}` in the existing `[target.….dependencies]` sections?

tomaka:
I've been trying to use more and more dynamic loading in glutin. X11 is currently loaded with `dlopen`, because systems that have wayland installed and not X11 should be able to run glutin apps as well. And #605 is an old PR that was supposed to make all the additional X libraries optional, notably because raspbian (debian for the raspberry pi) doesn't have them.

I don't know what the state of art about this is, but if you distribute pre-compiled binaries you don't want to distribute a X11 version, a wayland version, a X11-with-xcursor version, and so on..

It would be nicer if `x-settings` followed the same principle instead of using Cargo features.

ishitatsuyuki:
@pcwalton If you're still interested in this, this should now be added to winit. There's a non-standard, actual data based HiDPI detection currently.

francesca64:
Relevant winit issues: tomaka/winit#105 tomaka/winit#169

PR #700, opened 2016-01-06
x11: Prevent leak on XIQueryDevice

ecoal95:
Per http://linux.die.net/man/3/xiquerydevice:

> XIFreeDeviceInfo frees the information returned by
> XIQueryDevice.

Was discovered while investigating servo leaks.


vberger:
The travis failure is due to https://github.com/rust-lang/rust/issues/30713

tomaka:
Thanks!

PR #704, opened 2016-01-14
Fix double buffering on egl and restore default (fixing everything else)

dten:
Restore the default that was changed by https://github.com/tomaka/glutin/pull/690 (changing defaults isn't nice)

fix double buffer requirement check

double_buffering is forced on at line 603 and so this should fail if single buffer was requested

### DIFF
--- a/src/api/egl/mod.rs
+++ b/src/api/egl/mod.rs
@@ -538,7 +538,7 @@ unsafe fn choose_fbconfig(egl: &ffi::egl::Egl, display: ffi::egl::types::EGLDisp
             out.push(stencil as c_int);
         }
 
-        if let Some(true) = reqs.double_buffer {
+        if let Some(false) = reqs.double_buffer {
             return Err(CreationError::NoAvailablePixelFormat);
         }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -461,7 +461,7 @@ impl Default for PixelFormatRequirements {
             alpha_bits: Some(8),
             depth_bits: Some(24),
             stencil_bits: Some(8),
-            double_buffer: None,
+            double_buffer: Some(true),
             multisampling: None,
             stereoscopy: false,
             srgb: false,


tomaka:
Passing `None` in the requirements means "use the implementation default", which in practice means double-buffering for windows and single-buffering for pbuffers. That's exactly what we want.

There is indeed an error in the code, but it's that glutin should return an error for both `Some(true)` and `Some(false)`, as enforcing double-buffering or single-buffering is unimplemented.

dten:
fine but isn't changing the default quite a serious change? changing the default without implementing some way to change it back to what works sucks majorly.

with it set to None, glium just draws black on windows if the glutin dependency targets the commit from merging #690, exposing the setting in WindowBuilder and setting it so Some(true) results in things drawing again but breaks android because of the false Some(true) = fail check

tomaka:
> fine but isn't changing the default quite a serious change? changing the default without implementing some way to change it back to what works sucks majorly.

Well, it's not supposed to not work.

You're describing two problems:

 1 - Windows shouldn't create a single-buffered window by default
 2 - Glium shouldn't draw black even if the window is single buffered

I don't think we should fix these by reverting to wrong code.

dten:
the other change is not implemented to broken though...
the change to `src/api/egl/mod.rs` seems only right in the current state, you can ask for anything except Some(false)
also who says Windows shouldn't be single buffered. The docs for double_buffer say default is Some(true)

dten:
So my reasoning for this change is that changing the default is bad. Everything out there has already been able to assume that double buffering is the default and right now there's no way to disable that if you use a WindowBuilder. Hence Glium breaks, the default changed underneath it. Bad thing to do.

However the default was changed because Some(true) is erroneously being thrown out as bad despite that being the reality. That is the real fix, not changing the default.

tomaka:
It has only been since #677, merged less than a month ago, that the user is allowed to specify whether double buffering or simple buffering is required. Before this PR you couldn't even choose and you would just get an implementation-defined behavior (double buffering for windows, single buffering for pbuffers).

In #677 the default value for `double_buffer` was set to `Some(true)`. This was definitely a mistake, since pbuffers aren't supposed to be double-buffered. This change may have broken some stuff. Changing that to `None` restores the behavior as it was one month ago, when you weren't able to choose this.
If changing the value to `None` breaks Windows, then that means there was another problem in #667 that was only revealed when switching to `None`.

About rejecting `Some(true)`, as I said above both `Some(true)` and `Some(false)` should be rejected for now, until the EGL implementation is complete.

dten:
I see the wgl version only sets double buffer if requested. If i do a PR that makes it double unless false is set will that be ok ?

tomaka:
Sure!

dten:
I just had a look at changing that around https://github.com/tomaka/glutin/blob/master/src/api/wgl/mod.rs#L386 and https://github.com/tomaka/glutin/blob/master/src/api/wgl/mod.rs#L544 but it doesn't feel right either =/

tomaka:
I think the problem is that these functions (ChoosePixelFormat and wglChoosePixelFormat) can return multiple formats, but glutin just takes the first one. Maybe it's better if glutin took the last one instead, as they are often sorted from worse to best.

For the record, before #677 the formats were enumerated manually and picked manually.

PR #705, opened 2016-01-14
Prefer double buffer on wgl if unspecified

dten:
According to https://msdn.microsoft.com/en-us/library/windows/desktop/dd318284(v=vs.85).aspx if double buffer is unavailable then a single buffer will be returned

should this also be done for https://github.com/tomaka/glutin/blob/master/src/api/wgl/mod.rs#L544


tomaka:
Yes, please also change line 544. But also add a small comment explaining that in theory we shouldn't be doing that.

The change at line 544 is the most important, as `ChoosePixelFormat` is only used as a fallback.

dten:
For me this makes my glium project alive on all platforms again :+1: happy days

tomaka:
Note that glium is not supposed to display a black screen for single buffering, I guess it's another bug.

PR #706, opened 2016-01-16
x11: Poll the window until it is really visible

### LINKED ISSUE #697: Benign errors are now fatal
Before updating glutin to a version that includes #679, Servo would print

```
[glutin] x error code=8 major=42 minor=0!
```

and would then carry on with no apparent rendering issue. After updating, Servo panics instead:

```
thread 'Constellation' panicked at 'Failed to call XSendEvent after wakeup: XError { description: "BadMatch (invalid parameter attributes)", error_code: 8, request_code: 42, minor_code: 0 }', ../src/libcore/result.rs:760
```

… making it unusable: https://github.com/servo/servo/pull/9115

(Interestingly, the PR for updating https://github.com/servo/servo/pull/9096 landed, so maybe that particular error does not occur in CI servers? But it does occur on my machine and at least one other Servo contributor’s.)

* What’s the underlying cause of this error and should a fix be made so that it doesn’t happen in the first place?
* Should this particular error really be fatal? (Causing a panic through using `Result::except` in #679) It seemed safe to ignore in the previous behavior.
* Should *any* X11 error be fatal?

ecoal95:
Fixes #697

It seems that `XSync` doesn't really makes the window viewable.

This feels hacky, other option to do it could be using `XIfEvent` or
similar to listen to `MapNotify` events, but we'll have a loop still.

In practice, this lasts between two and thre iterations on my machine,
which is something not noticeable.

### DIFF
--- a/src/api/x11/window.rs
+++ b/src/api/x11/window.rs
@@ -9,6 +9,8 @@ use std::sync::atomic::AtomicBool;
 use std::collections::VecDeque;
 use std::sync::{Arc, Mutex};
 use std::os::raw::c_long;
+use std::thread;
+use std::time::Duration;
 
 use Api;
 use ContextError;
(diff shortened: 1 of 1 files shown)


tomaka:
Well, I'm going to trust you on this one.

PR #709, opened 2016-01-17
support utf8 window titles via _NET_WM_NAME standard

gdamjan:
based on this freerdp patch
https://github.com/FreeRDP/FreeRDP/commit/9767f7f042a58aae876e0ad3b2e7bde356c8fda9

thanks to emiliocobos on irc


tomaka:
:shipit:

gdamjan:
for reference
http://standards.freedesktop.org/wm-spec/1.3/ar01s05.html

PR #710, opened 2016-01-19
Allow it to work under Xvfb.

notriddle:
Part of servo/servo#8641


tomaka:
I'm hesitating. It's definitely not a good change, as it ignores errors on purpose, but on the other hand fullscreen support is so crappy that it needs a rewrite anyway.

notriddle:
Any way to get into full screen mode without using that extension? We could
use that as a fallback instead...

tomaka:
I'll just merge this.

PR #714, opened 2016-01-26
Replace `i8` with `libc::c_char`

mmatyas:
On ARM/AArch64 platforms, `char` is traditionally unsigned, using `i8` causes compile errors. We should use `libc::c_char` instead.


tomaka:
Thanks

PR #718, opened 2016-02-02
osx: scroll delta should take hidpi factor into account

paulrouget:
Fixes https://github.com/servo/servo/issues/9502


tomaka:
These scrolling issues are a bit over my head. I'm going to trust you and merge blindly.

PR #729, opened 2016-02-26
Stop using `NSFullSizeContentViewWindowMask` to get borderless windows, because that disables the swap interval.

pcwalton:
This hack is very simple but a little evil. We get the superview of the
content view, which is the `NSThemeFrame`, and install an OpenGL context
into it.  `NSThemeFrame` is a private class, but we never speak its
name, we get to it with public APIs, and we only call public `NSView`
APIs on it, so this seems OK in terms of being supported by Apple going
forward.

Additionally, we remove the standard window buttons so that they
disappear and the user can't click them. This can also be done using
public APIs.

The reason for using this hack as opposed to
`NSFullSizeContentViewWindowMask` is that the latter forces the window
to be Core Animation-backed, which results in us rendering to an off
screen surface. Not only does this inject another compositor into the
system, but it also results in us rendering to an off-screen surface,
disabling the swap interval.

This depends on a `cocoa-rs` upgrade to add a binding to the `-[NSView
superview]` method.

Note that the top edge of the window is not rounded, although the shadow
is. Applications that wish to use this mode will need to round the top
edge themselves if they wish. (This is how Cocoa internally works:
rounding is done by the app, not the window manager.)

This should fix servo/servo#9431.

### DIFF
--- a/src/api/cocoa/mod.rs
+++ b/src/api/cocoa/mod.rs
@@ -295,7 +295,7 @@ impl Window {
             Some(window) => window,
             None         => { return Err(OsError(format!("Couldn't create NSWindow"))); },
         };
-        let view = match Window::create_view(*window) {
+        let view = match Window::get_or_create_view(*window, win_attribs.decorations) {
             Some(view) => view,
             None       => { return Err(OsError(format!("Couldn't create NSView"))); },
         };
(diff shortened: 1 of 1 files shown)


src/api/cocoa/mod.rs
```diff
@@ -416,8 +416,7 @@ impl Window {
                 NSClosableWindowMask as NSUInteger |
                 NSMiniaturizableWindowMask as NSUInteger |
                 NSResizableWindowMask as NSUInteger |
-                NSTitledWindowMask as NSUInteger |
-                NSFullSizeContentViewWindowMask as NSUInteger
+                NSTitledWindowMask as NSUInteger,
```
tomaka: Travis complains that this coma shouldn't be there.

tomaka:
I'm not qualified to review this! If someone who knows cocoa wants to have a look at it, that would be nice.
Otherwise I'm just going to be merge in a few days.

paulrouget:
This has been reviewed by someone who knows cocoa here: https://github.com/servo/glutin/pull/68#issuecomment-188277073

ZeGentzy:
No longer relevant.

PR #731, opened 2016-02-29
cocoa: Post Awakened events at the end of the queue to avoid floods.

pcwalton:
Closes servo/webrender#179.

r? @paulrouget


paulrouget:
Thanks.

PR #732, opened 2016-03-02
Fix this_platform_is_not_supported on OpenBSD.

zkirill:
I followed the steps of @mneumann.

Now runs on snapshot OpenBSD 5.9 with X11, rustc 1.6.0 and cargo 0.9.0 from https://github.com/jasperla/openbsd-wip.


tomaka:
Thanks!

PR #733, opened 2016-03-02
Fix memory leak while creating NSEvent and swapping buffers (Fixes #514)

### LINKED ISSUE #514: Memory usage increasing over time in OS X
In OS X glutin program memory usage slow increasing over time. (example [1](https://github.com/tomaka/glium/blob/master/examples/teapot.rs) [2](https://github.com/PistonDevelopers/piston-examples/blob/master/src/sprite.rs))
especially when shaking the mouse memory usage will increase rapidly.

[reddit post](https://www.reddit.com/r/rust_gamedev/comments/3c2us6/newbie_quesion_this_is_a_memory_leak/)

fkaa:


paulrouget:
@tomaka what do you think?

tomaka:
I have no opinion on that. I don't know how it works!

pcwalton:
Looks good to me.

@tomaka @paulrouget For future reference: In Objective-C, objects can be memory-managed in two ways: (a) the reference count is managed explicitly by the runtime and the object is destroyed when it reaches 0; (b) the object is placed in a per-thread arena (`NSAutoreleasePool`) and the arena is emptied at the end of each turn of the event loop. Ordinarily, Cocoa manages this for you, because the built-in `NSRunLoop` that `NSApplicationMain()` pumps (I believe it's `NSApplicationMain()`; might be some other function) automatically sets up an autorelease pool and flushes it after every event. But if you're interacting with Objective-C objects from another thread, or you're pumping the event loop manually, you have to set up and manage an autorelease pool yourself. If you don't, objects that were allocated via method (b) will leak. Glutin isn't doing this, so lots of objects are leaking; this patch fixes this.

paulrouget:
r+

PR #734, opened 2016-03-03
Add phases to wheel events.

pcwalton:
Needed for apps to support overscroll on the Mac.

r? @paulrouget

### DIFF
--- a/src/api/cocoa/mod.rs
+++ b/src/api/cocoa/mod.rs
@@ -42,10 +42,10 @@ use std::sync::Mutex;
 use std::ascii::AsciiExt;
 use std::ops::Deref;
 
-use events::Event::{Awakened, MouseInput, MouseMoved, ReceivedCharacter, KeyboardInput, MouseWheel, Closed, Focused, TouchpadPressure};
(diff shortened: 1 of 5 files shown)


paulrouget:
`TouchPhase` not imported everywhere (at least windows, don't know about linux).

tomaka:
Even more bloating :'(

metajack:
@tomaka Not sure what you mean by bloat. Aren't users of the library going to want to use all the events at their disposal?

metajack:
Superceded by #735

tomaka:
@metajack The major problem here is that it's totally undocumented.
Also this commit broke Emscripten and Android, and probably iOS.
Also this commit is a breaking change and will require bumping the major version.

When I started glutin I wanted to write clean code, and in the last month or so this has been a total failure.

metajack:
Sorry I would have addressed those in the updated patch had I known or had it failed tests. We can certainly do a better job of adding documentation in the future.

On the bright side, this has enabled some great features in Browser.html / Servo.

PR #744, opened 2016-03-19
Enable more Windows keys

adamncasey:
I started looking into this after the period key wasn't working in Servo. Some keys were enabled the other day but more are still missing.

This should enable most required keys on Windows. It also changes one mapping VK_OEM_PLUS from Add to Equals. I believe Equals is the correct binding of this key.

Mappings taken from: https://msdn.microsoft.com/en-gb/library/windows/desktop/dd375731%28v=vs.85%29.aspx and applied with some guesswork to the Glutin names.

I'm not 100% sure on whether the Grave key is correct - I only have wikipedia to go on for what a US keyboard looks like. 

Does anyone know how libraries such as glutin should handle keyboards with more keys? For example most non-US keyboards (that I've seen) has an extra #~ key. I couldn't find a Glutin key code for this.


src/api/win32/event.rs
```diff
@@ -143,44 +143,38 @@ pub fn vkeycode_to_element(wparam: winapi::WPARAM, lparam: winapi::LPARAM) -> (S
...
+        winapi::VK_VOLUME_UP => Some(VirtualKeyCode::VolumeUp),
+        winapi::VK_MEDIA_NEXT_TRACK => Some(VirtualKeyCode::NextTrack),
+        winapi::VK_MEDIA_PREV_TRACK => Some(VirtualKeyCode::PrevTrack),
+        winapi::VK_MEDIA_STOP => Some(VirtualKeyCode::MediaStop),
+        winapi::VK_MEDIA_PLAY_PAUSE => Some(VirtualKeyCode::PlayPause),
+        winapi::VK_LAUNCH_MAIL => Some(VirtualKeyCode::Mail),
+        winapi::VK_LAUNCH_MEDIA_SELECT => Some(VirtualKeyCode::MediaSelect),
+        /*winapi::VK_LAUNCH_APP1 => Some(VirtualKeyCode::Launch_app1),
+        winapi::VK_LAUNCH_APP2 => Some(VirtualKeyCode::Launch_app2),*/
+        winapi::VK_OEM_1 => Some(VirtualKeyCode::Semicolon),
```
tomaka: The link says that this can vary between keyboards, so this souldn't be set to `Semicolon`.

src/api/win32/event.rs
```diff
@@ -143,44 +143,38 @@ pub fn vkeycode_to_element(wparam: winapi::WPARAM, lparam: winapi::LPARAM) -> (S
...
-        winapi::VK_OEM_7 => Some(VirtualKeyCode::Oem_7),
-        winapi::VK_OEM_8 => Some(VirtualKeyCode::Oem_8),
-        winapi::VK_OEM_102 => Some(VirtualKeyCode::Oem_102),
-        winapi::VK_PROCESSKEY => Some(VirtualKeyCode::Processkey),
+        winapi::VK_OEM_2 => Some(VirtualKeyCode::Slash),
+        winapi::VK_OEM_3 => Some(VirtualKeyCode::Apostrophe),
+        winapi::VK_OEM_4 => Some(VirtualKeyCode::LBracket),
+        winapi::VK_OEM_5 => Some(VirtualKeyCode::Backslash),
+        winapi::VK_OEM_6 => Some(VirtualKeyCode::RBracket),
+        winapi::VK_OEM_7 => Some(VirtualKeyCode::Grave),
```
tomaka: Same for all these ^

tomaka:
The "virtual key code" indicates the meaning of the key. It is perfectly fine to have keyboard events whose virtual key code is `None`.

adamncasey:
Hmm OK. That seems reasonable.

Servo at the moment ignores all key events which don't have a virtual key code. But that isn't glutin's problem.

I'll remove the incorrect virtual key codes and squash.

tomaka:
Thanks!

I'm a bit dubious about that change from Plus to Equals. That seems weird.

adamncasey:
Sorry, github email reply didn't come through:

I see your point. My thinking was that Add lines up better to the Numpad Add input (It is listed along with Divide, Multiply, Subtract in the VK_ codes as well.

I just tested in servo (not a perfect citizen of glutin KeyboardInput(?), but perhaps representative of an application using glutin): VK_OEM_PLUS mapping to Add produces no output in a textbox, but setting it to VirtualKeyCode::Equals gives the correct output on US/UK keyboards.

adamncasey:
According to an OSX header file referenced here: http://stackoverflow.com/a/16125341/655811 , the mac virtual key codes will be incorrect in non-US keyboard layouts as well.

Given that the consequences would be the same as the Mac keyboard binds, I think it would be better to leave the windows OEM keys in. It prevents quite a surprising difference between platforms when porting glutin-based programs.

tomaka:
> I just tested in servo (not a perfect citizen of glutin KeyboardInput(?), but perhaps representative of an application using glutin): VK_OEM_PLUS mapping to Add produces no output in a textbox, but setting it to VirtualKeyCode::Equals gives the correct output on US/UK keyboards.

You're not supposed to use the virtual key code for keyboard input. Plus that could simply be a bug in Servo.

> Given that the consequences would be the same as the Mac keyboard binds, I think it would be better to leave the windows OEM keys in. It prevents quite a surprising difference between platforms when porting glutin-based programs.

It's the opposite, mac virtual key codes should be removed as well.

adamncasey:
You're right, my initial ideas behind needing this were incorrect: the way servo handles keyboard input isn't right. I'll see if the current keyboard work will fix that :)

It looks like the OS X and Windows interpretation of of "Virtual Key Code" is different to the one mentioned above though - both documentations mention that the codes aren't the actual printed keycap value. 

I don't think it makes sense to take out the OS X key codes

adamncasey:
It could be worth seeing what other libraries similar to Glutin to here. My interpretation of the similarities and differences between OS keyboard input systems isn't particularly relevent.

tomaka:
Could you run the glutin example named "window" and check that pressing '+' or '=' on your numpad produces the correct events?

adamncasey:
With the changes in this PR applied, I pressed the following numpad keys in order: '+' '-' '*' '/' '<Enter>'. My numpad doesn't have a '=' key.
Followed by '+/=' next to backspace (with & without shift).

````
KeyboardInput(Pressed, 78, Some(Add))
ReceivedCharacter('+')
KeyboardInput(Released, 78, Some(Add))
KeyboardInput(Pressed, 74, Some(Subtract))
ReceivedCharacter('-')
KeyboardInput(Released, 74, Some(Subtract))
KeyboardInput(Pressed, 55, Some(Multiply))
ReceivedCharacter('*')
KeyboardInput(Released, 55, Some(Multiply))
KeyboardInput(Pressed, 53, Some(Divide))
ReceivedCharacter('/')
KeyboardInput(Released, 53, Some(Divide))
KeyboardInput(Pressed, 28, Some(Return))
ReceivedCharacter('\r')
KeyboardInput(Released, 28, Some(Return))
KeyboardInput(Pressed, 54, Some(RShift))
KeyboardInput(Pressed, 13, Some(Equals))
ReceivedCharacter('+')
KeyboardInput(Released, 54, Some(RShift))
KeyboardInput(Released, 13, Some(Equals))
KeyboardInput(Pressed, 13, Some(Equals))
ReceivedCharacter('=')
KeyboardInput(Released, 13, Some(Equals))
````

tomaka:
So that didn't trigger `VK_OEM_PLUS` I guess.

adamncasey:
Only the '+=' key triggers that from what I have witnessed. There are some MSDN comments & reddit posts online which confirm this as well, but the only official documentation I can see just states it's the '+' key.

SDL also does the same thing as this PR. https://github.com/spurious/SDL-mirror/blob/c4420fc9de77b59f0fb510dad1994c3b249c55b9/src/video/winrt/SDL_winrtkeyboard.cpp#L236

tomaka:
Oh, the SDL code is a good argument.
Well, let's merge this then.

PR #745, opened 2016-03-20
Update objc dependency to 0.2

SSheldon:
I just published version 0.2 of the objc crate, changelog here: https://github.com/SSheldon/rust-objc/blob/master/CHANGELOG.md#020

Main differences are gnustep support, (optionally) more type checking, removed the libc dependency, and fixed a bug in the space reserved for ivars when declaring classes.

By removing the libc dependency, objc [can't break glutin anymore](https://github.com/SSheldon/rust-objc/commit/fbb62f89e77266500b08e7cc331926a67a7784d9#commitcomment-14262093) by having a mismatch between libc versions :) 

This requires servo/cocoa-rs#120 to be merged first, which also means the Travis checks will fail until then.

Testing: verified glutin still compiles and tests pass, verified the window example looks good.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,24 +23,24 @@ khronos_api = "1.0"
 version = "0.1"
 
 [target.i386-apple-ios.dependencies]
-objc = "0.1.8"
+objc = "0.2"
 
 [target.x86_64-apple-ios.dependencies]
-objc = "0.1.8"
+objc = "0.2"
 
 [target.aarch64-apple-ios.dependencies]
-objc = "0.1.8"
+objc = "0.2"
 
 [target.armv7s-apple-ios.dependencies]
-objc = "0.1.8"
+objc = "0.2"
 
 [target.armv7-apple-ios.dependencies]
-objc = "0.1.8"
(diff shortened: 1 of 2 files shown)


fkaa:
Seems to refer to a `0.3` version of `cocoa` which doesn't exist?

Ms2ger:
Cocoa has been published.

tomaka:
Thanks!

Do `cocoa` and `objc` have to be tied to a specific version? If so, it can still break :-/

SSheldon:
Yeah, it can still break, but as long as `cocoa` and `objc` are good about their versioning, it can only break by someone making a change to `glutin`, it won't suddenly happen to users. Being good about versioning can be hard, though, as demonstrated by the `libc` issue I caused with my bump to `objc` :P

To get rid of the requirement that `cocoa` and `objc` be on specific version together, we'd have to stop using the `msg_send!` macro on types from the `cocoa` crate. Or `cocoa` would have to export its own version of the `msg_send!` macro.

Using types from other libraries is still pretty hard :/

PR #746, opened 2016-03-21
Implement CursorState::Hide for x11

Thinkofname:
I used http://stackoverflow.com/a/664528 for a reference on how to implement this.


tomaka:
Thanks

PR #750, opened 2016-03-25
Add support for X11 window size constraints

BonsaiDen:
Size hints are only being set for non-fullscreen windows, if
`max_dimensions` are set they'll override the normal `dimensions`
since X11 will not automatically resize the window after setting
the size hints.

`PSize` hint is currently set along with the `min/max` hints for
good measure, might be superfluous though.


src/api/x11/window.rs
```diff
@@ -310,11 +310,7 @@ impl Window {
                pf_reqs: &PixelFormatRequirements, opengl: &GlAttributes<&Window>)
                -> Result<Window, CreationError>
     {
-        let dimensions = window_attrs.dimensions.unwrap_or((800, 600));
-
-        // not implemented
-        assert!(window_attrs.min_dimensions.is_none());
-        assert!(window_attrs.max_dimensions.is_none());
+        let dimensions = window_attrs.max_dimensions.unwrap_or(window_attrs.dimensions.unwrap_or((800, 600)));
```
tomaka: Doesn't that mean that the window will always have the max dimensions when you create it, instead of the dimensions you specified?

tomaka:
> since X11 will not automatically resize the window after setting the size hints.

Did you mean X11 *will* automatically resize the window after setting the size hints?

BonsaiDen:
No, it will definitely not resize the window for you. If your window is bigger than `max_dimensions` when you set those, it will stay bigger. Same goes for `min_dimensions`, I guess this stems from the fact that these things are really just "hints" and are only checked once you actually start resizing the window.

As for defaulting to the `max_dimensions` I do see the problem and why it might not be the desired behaviour, I think the correct logic here would be to limit the `dimensions` to the respective max/min values if those are present.

BonsaiDen:
I've updated the code to correctly limited with the initially requested dimensions according the specified min/max values.

tomaka:
Thanks

PR #755, opened 2016-03-30
glx: Support non-conformant multisampling fbconfigs.

### LINKED ISSUE #754: GLX: `CONFIG_CAVEAT`  should be set to `NON_CONFORMANT_CONFIG` when multisampling is requested
Multisampling FBConfigs are non-conformant and currently only `NONE` is set when `reqs.hardware_accelerated` which causes the `FBConfig` to return `NoAvailablePixelFormat`.

Hacking the `NON_CONFORMANT_CONFIG` into `glx/mod.rs` correctly creates a framebuffer with multisampling enabled:

```       
if let Some(hardware_accelerated) = reqs.hardware_accelerated {
    out.push(ffi::glx::CONFIG_CAVEAT as c_int);
    out.push(if hardware_accelerated {
        ffi::glx::NON_CONFORMANT_CONFIG as c_int
        //ffi::glx::NONE as c_int
    } else {
        ffi::glx::SLOW_CONFIG as c_int
    });
}
```

I'll try to whip up a PR which checks for multisampling tomorrow.

BonsaiDen:
Fixes #754.

### DIFF
--- a/src/api/glx/mod.rs
+++ b/src/api/glx/mod.rs
@@ -360,6 +360,7 @@ unsafe fn choose_fbconfig(glx: &ffi::glx::Glx, extensions: &str, xlib: &ffi::Xli
                           -> Result<(ffi::glx::types::GLXFBConfig, PixelFormat), ()>
 {
     let descriptor = {
+        let mut glx_non_conformant = false;
         let mut out: Vec<c_int> = Vec::with_capacity(37);
 
         out.push(ffi::glx::X_RENDERABLE as c_int);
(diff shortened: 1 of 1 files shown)


tomaka:
I was thinking of always putting `NON_CONFORMANT`, not just for multisampling.

BonsaiDen:
My initial idea was to limit the scope of  `NON_CONFORMANT` to situations where it is explicitly needed, but that might of course cause more issues in the future and users might not go the extra mile to actually figure why they can't get a non-conformant fb config to work.

I personally would be fine with putting `NON_CONFORMANT` in all the time, we can still refine it in case we run into issues in the future.

tomaka:
I'm merging this since I'd like to publish a new version of glutin with the fix.
I'm in favor of using NON_CONFORMANT all the time, but I don't have a strong opinion on this.

PR #760, opened 2016-04-10
Fixed context creation on Windows

fika04:
The WGL context was never created and the function CreateContextAttribsARB attempted to point to one. This should fix the issue. Also the ShareLists part of the code is completely redundant as `share` can only ever really be null.

Closes tomaka/glium#1450


tomaka:
I don't think that's the right fix. As far as I know `CreateContextAttribsARB` doesn't need a current context. The fix probably works for you because you end up using the regular `CreateContext`.

PR #763, opened 2016-04-13
glx: Always set CONFIG_CAVEAT to DONT_CARE

### LINKED ISSUE #754: GLX: `CONFIG_CAVEAT`  should be set to `NON_CONFORMANT_CONFIG` when multisampling is requested

Bobo1239:
Fixes #754

### DIFF
--- a/src/api/glx/mod.rs
+++ b/src/api/glx/mod.rs
@@ -360,7 +360,6 @@ unsafe fn choose_fbconfig(glx: &ffi::glx::Glx, extensions: &str, xlib: &ffi::Xli
                           -> Result<(ffi::glx::types::GLXFBConfig, PixelFormat), ()>
 {
     let descriptor = {
-        let mut glx_non_conformant = false;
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks

PR #766, opened 2016-04-16
Improve X11 cursor symbols handling

### LINKED ISSUE #765: Can't set cursor to hand on X11
When I try to run the cursor example, and press some key multiple times, it will change the cursor, but it will panic when trying the hand cursor.

This is probably the cause for servo bug https://github.com/servo/servo/issues/10475

```
$ RUST_BACKTRACE=1 cargo run --example cursor
     Running `target/debug/examples/cursor`
OpenGL version 3.3 (Core Profile) Mesa 11.0.2
Setting cursor to "Default"
Setting cursor to "Crosshair"
Setting cursor to "Hand"
thread '<main>' panicked at 'Failed to call XPeekEvent: XError { description: "BadCursor (invalid Cursor parameter)", error_code: 6, request_code: 95, minor_code: 0 }', ../src/libcore/result.rs:746
stack backtrace:
   1:     0x7f0e71a69230 - sys::backtrace::tracing::imp::write::h3675b4f0ca767761Xcv
   2:     0x7f0e71a6bacb - panicking::default_handler::_$u7b$$u7b$closure$u7d$$u7d$::closure.44519
   3:     0x7f0e71a6b738 - panicking::default_handler::h18faf4fbd296d909lSz
   4:     0x7f0e71a628cc - sys_common::unwind::begin_unwind_inner::hfb5d07d6e405c6bbg1t
   5:     0x7f0e71a62d58 - sys_common::unwind::begin_unwind_fmt::h8b491a76ae84af35m0t
   6:     0x7f0e71a687e1 - rust_begin_unwind
   7:     0x7f0e71a9aeef - panicking::panic_fmt::h98b8cbb286f5298alcM
   8:     0x7f0e7196a071 - result::unwrap_failed::h2422879806685704086
                        at ../src/libcore/macros.rs:29
   9:     0x7f0e71969e6a - result::Result<T, E>::expect::h4742976310235235929
                        at ../src/libcore/result.rs:708
  10:     0x7f0e7196e4ac - api::x11::window::WaitEventsIterator<'a>.Iterator::next::h377d65c98f632d97vng
                        at src/api/x11/window.rs:285
  11:     0x7f0e71916358 - platform::platform::api_dispatch::WaitEventsIterator<'a>.Iterator::next::h8230b9405dbeb438ozh
                        at src/platform/linux/api_dispatch.rs:169
  12:     0x7f0e719162fb - window::WaitEventsIterator<'a>.Iterator::next::hb10ac20a53b3f4c3cEk
                        at src/window.rs:589
  13:     0x7f0e7190da5f - main::h27e09f06a3a5c622Ape
                        at examples/cursor.rs:23
  14:     0x7f0e71a6b394 - sys_common::unwind::try::try_fn::h14622312129452522850
  15:     0x7f0e71a6876b - __rust_try
  16:     0x7f0e71a6ae2b - rt::lang_start::h0ba42f7a8c46a626rKz
  17:     0x7f0e71916b29 - main
  18:     0x7f0e70ceda3f - __libc_start_main
  19:     0x7f0e7190d7b8 - _start
  20:                0x0 - <unknown>
Process didn't exit successfully: `target/debug/examples/cursor` (exit code: 101)
```

est31:
1. Make it non failing. Setting to an unknown cursor won't panic anymore,
   but use the default cursor instead.

2. Introduce a system where multiple special cursors are tried
   in order to work with different themes and desktop environments.
   This way we get less often into the situation where we have to use a
   default cursor.

3. Also set names for some cursors that previously only had a placeholder.

Fixes #765.
Will fix https://github.com/servo/servo/issues/10475 as well.

### DIFF
--- a/src/api/x11/window.rs
+++ b/src/api/x11/window.rs
@@ -841,50 +841,76 @@ impl Window {
     pub fn set_cursor(&self, cursor: MouseCursor) {
         unsafe {
             use std::ffi::CString;
-            let cursor_name = match cursor {
-                MouseCursor::Alias => "link",
-                MouseCursor::Arrow => "arrow",
-                MouseCursor::Cell => "plus",
-                MouseCursor::Copy => "copy",
-                MouseCursor::Crosshair => "crosshair",
-                MouseCursor::Default => "left_ptr",
-                MouseCursor::Grabbing => "grabbing",
-                MouseCursor::Hand | MouseCursor::Grab => "hand",
-                MouseCursor::Help => "question_arrow",
-                MouseCursor::Move => "move",
-                MouseCursor::NoDrop => "circle",
-                MouseCursor::NotAllowed => "crossed_circle",
-                MouseCursor::Progress => "left_ptr_watch",
+            let load = |name :&str| {
+                let c_string = CString::new(name.as_bytes().to_vec()).unwrap();
+                return (self.x.display.xcursor.XcursorLibraryLoadCursor)(self.x.display.display, c_string.as_ptr());
+            };
+            let loadn = |names :&[&str]| {
+                for name in names.iter() {
+                    let xcursor = load(*name);
+                    if xcursor != 0 {
+                        return xcursor;
+                    }
+                }
+                return 0;
+            };
+            // Try multiple names in some cases where the name
+            // differs on the desktop environments or themes.
+            //
+            // Try the better looking (or more suiting) names first.
+            let mut xcursor = match cursor {
+                MouseCursor::Alias => load("link"),
+                MouseCursor::Arrow => load("arrow"),
+                MouseCursor::Cell => load("plus"),
+                MouseCursor::Copy => load("copy"),
+                MouseCursor::Crosshair => load("crosshair"),
+                MouseCursor::Default => load("left_ptr"),
+                MouseCursor::Hand => load("hand1"),
+                MouseCursor::Help => load("question_arrow"),
+                MouseCursor::Move => load("move"),
+                MouseCursor::Grab => loadn(&["openhand", "grab"]),
+                MouseCursor::Grabbing => loadn(&["closedhand", "grabbing"]),
+                MouseCursor::Progress => load("left_ptr_watch"),
+                MouseCursor::AllScroll => load("all-scroll"),
+                MouseCursor::ContextMenu => load("context-menu"),
+
+                MouseCursor::NoDrop => loadn(&["no-drop", "circle"]),
+                MouseCursor::NotAllowed => load("crossed_circle"),
+
 
                 /// Resize cursors
-                MouseCursor::EResize => "right_side",
-                MouseCursor::NResize => "top_side",
-                MouseCursor::NeResize => "top_right_corner",
-                MouseCursor::NwResize => "top_left_corner",
-                MouseCursor::SResize => "bottom_side",
-                MouseCursor::SeResize => "bottom_right_corner",
-                MouseCursor::SwResize => "bottom_left_corner",
-                MouseCursor::WResize => "left_side",
-                MouseCursor::EwResize | MouseCursor::ColResize => "h_double_arrow",
-                MouseCursor::NsResize | MouseCursor::RowResize => "v_double_arrow",
-                MouseCursor::NwseResize => "bd_double_arrow",
-                MouseCursor::NeswResize => "fd_double_arrow",
+                MouseCursor::EResize => load("right_side"),
+                MouseCursor::NResize => load("top_side"),
+                MouseCursor::NeResize => load("top_right_corner"),
+                MouseCursor::NwResize => load("top_left_corner"),
+                MouseCursor::SResize => load("bottom_side"),
+                MouseCursor::SeResize => load("bottom_right_corner"),
+                MouseCursor::SwResize => load("bottom_left_corner"),
+                MouseCursor::WResize => load("left_side"),
+                MouseCursor::EwResize => load("h_double_arrow"),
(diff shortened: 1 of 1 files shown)


src/api/x11/window.rs
```diff
@@ -841,46 +841,80 @@ impl Window {
...
-                MouseCursor::Hand | MouseCursor::Grab => "hand",
-                MouseCursor::Help => "question_arrow",
-                MouseCursor::Move => "move",
-                MouseCursor::NoDrop => "circle",
-                MouseCursor::NotAllowed => "crossed_circle",
-                MouseCursor::Progress => "left_ptr_watch",
+            let load = |name :&str| {
+                let c_string = CString::new(name.as_bytes().to_vec()).unwrap();
+                return (self.x.display.xcursor.XcursorLibraryLoadCursor)(self.x.display.display, c_string.as_ptr());
+                self.x.display.check_errors().expect("Failed to call XcursorLibraryLoadCursor");
```
jdm: Isn't this unreachable code, given the `return` on the previous line?

est31: Thanks for noting, fixed.

tomaka:
Oh I'm really not qualified to review that :-/

src/api/x11/window.rs
```diff
@@ -841,46 +841,81 @@ impl Window {
...
-                MouseCursor::Crosshair => "crosshair",
-                MouseCursor::Default => "left_ptr",
-                MouseCursor::Grabbing => "grabbing",
-                MouseCursor::Hand | MouseCursor::Grab => "hand",
-                MouseCursor::Help => "question_arrow",
-                MouseCursor::Move => "move",
-                MouseCursor::NoDrop => "circle",
-                MouseCursor::NotAllowed => "crossed_circle",
-                MouseCursor::Progress => "left_ptr_watch",
+            let load = |name :&str| {
```
emilio: Why not:

```rust
let load = |names: &[&str]| {
    for name in names.iter() { ... }
};
```

src/api/x11/window.rs
```diff
@@ -841,46 +841,81 @@ impl Window {
...
+                let c_string = CString::new(name.as_bytes().to_vec()).unwrap();
+                let xcursor = (self.x.display.xcursor.XcursorLibraryLoadCursor)(self.x.display.display, c_string.as_ptr());
+                self.x.display.check_errors().expect("Failed to call XcursorLibraryLoadCursor");
```
emilio: I don't see any reference to generic x errors in[1], so this line could be removed:

[1]: http://linux.die.net/man/3/xcursorlibraryloadcursor

emilio:
@tomaka: This was raised while investigating https://github.com/servo/servo/issues/10475.

It seems KDE and GNOME use different names for cursors, so loading a cursor might fail depending on the name.

Also, the way to check for the correct load of a cursor is not the generic `check_errors()`, but checking the return value (which is 0 if the cursor failed to load).

What caused the X error was calling `SetCursor` with a null cursor.

src/api/x11/window.rs
est31: Removing load2 and replacing it with a generic loadn is a good idea, but I'll keep `load` imo as it allows to be called in a more compact way.

est31:
@tomaka Additionally to what @emilio said, I have found multiple panic causing cursor symbols similar to #765 (not a full list but `NwseResize`, `NeswResize`, `Grabbing` all caused a panic), I've fixed all of them.

That `XcursorLibraryLoadCursor` returns `0` if it couldn't find a cursor was found out by me with trial and error, I don't have any docs for this.

My main source was the contents of a cursors theme directory on my computer (for me it was `/usr/share/icons/breeze_cursors/cursors/`).

est31:
I've noted a small mistake with how I wrote the loadn function: at the end i panicked, but this was wrong. Now I return 0 like it should be.

I've squashed it all into one commit again.

src/api/x11/window.rs
```diff
@@ -841,46 +841,81 @@ impl Window {
...
-            let c_string = CString::new(cursor_name.as_bytes().to_vec()).unwrap();
-            let xcursor = (self.x.display.xcursor.XcursorLibraryLoadCursor)(self.x.display.display, c_string.as_ptr());
-            self.x.display.check_errors().expect("Failed to call XcursorLibraryLoadCursor");
+
+            // Fall back to default cursor if no cursor was found.
+            // If default wasn't found either, panic.
+            if xcursor == 0 {
+                xcursor = load("left_ptr");
+                if xcursor == 0 {
+                    panic!(concat!("Couldn't set cursor to \"{:?}\", ",
```
emilio: There's no need to panic here: according to [the docs][1], calling XDefineCursor with `None` (that is, `0`) value will use the parent window cursor or the default one.

That means I was wrong about when the error was being generated (it was generated in [XFreeCursor][2]).

So we should remove this two `if` statements, and just add:

```rust
if xcursor != {
    (self.x.display.xlib.XFreeCursor)(self.x.display.display, xcursor);
}
```

And then *after it*, if we want to check errors (that should never happen but anyways):

```rust
self.x.display.check_errors().expect("Failed to set or free the cursor");
```

There should be no need here for `XFlush()`, but maybe @tomaka knows why it was there in the first place.


[1]: https://tronche.com/gui/x/xlib/window/XDefineCursor.html
[2]: https://tronche.com/gui/x/xlib/pixmap-and-cursor/XFreeCursor.html

est31: Indeed, XFreeCursor is the troublemaker.

emilio:
This looks good to me!

Also, I have a patch over this one to add hiding the cursor :)

r? @tomaka

PR #767, opened 2016-04-17
x11: Add cursor hiding

emilio:


emilio:
cc @est31

tomaka:
That may be a weird request, but please don't fix the warnings. I'm trying to keep the winit repo in sync, and you're easing my life if you avoid merge conflicts if possible.

est31:
Perhaps also use the new function from `set_cursor_state`? It has a functionality to hide the cursor as well.

emilio:
Heh, no problem, removed the last commit :-)

emilio:
@est31 you're right, let me find a good way to do it :)

src/api/x11/window.rs
```diff
@@ -975,18 +982,13 @@ impl Window {
         match state {
             Normal => Ok(()),
             Hide => {
-                let data = &[0, 0, 0, 0, 0, 0, 0, 0];
                 unsafe {
-                    let mut black = ffi::XColor {
-                        red: 0, green: 0, blue: 0,
-                        pad: 0, pixel: 0, flags: 0,
-                    };
-                    let bitmap = (self.x.display.xlib.XCreateBitmapFromData)(self.x.display.display, self.x.window, data.as_ptr(), 8, 8);
```
emilio: Also, I think this line could segfault pretty easily (has only 8 bytes, but the image was 8x8).

Not totally sure though, I don't know well enough the XPM format.

emilio:
Refactored!

re-r? @tomaka

est31:
Cursor hiding from the cursors example works for me now, and the patch looks good.

tomaka:
I think we should eventually split that file into several modules to make it easier to understand, but that's out of scope of that PR.

Thanks!

PR #768, opened 2016-04-23
Get cursor pos

trolleyman:
This PR adds a new function, `get_cursor_pos` to `Window`. It is meant to call the underlying system and request the cursor position relative to the window origin. I made this because currently, on Windows at least, it is impossible through glutin to get the current cursor position.

I tried to implement all systems, however I cannot test if this works everywhere, only on Windows.

### DIFF
--- a/src/api/cocoa/mod.rs
+++ b/src/api/cocoa/mod.rs
@@ -727,11 +727,8 @@ impl Window {
     #[inline]
     pub fn get_cursor_position(&self) -> Result<(i32, i32), ()> {
         unsafe {
-            // Very unsafe, but I think it's fine as mouseLocation doesn't touch the pointer.
-            // https://github.com/servo/cocoa-rs/blob/master/src/appkit.rs#L2070-L2072
-            // TODO: Check for errors.
-            let point = NSEvent::mouseLocation(ptr::null_mut() as id);
-            Ok(self.view.convertPoint_fromView_(point, ptr::null_mut() as id))
+            let point = self.window.mouseLocationOutsideOfEventStream();
+            Ok((point.x, point.y))
         }
     }
 }
--- a/src/api/cocoa/mod.rs
+++ b/src/api/cocoa/mod.rs
@@ -27,7 +27,6 @@ use cocoa::foundation::{NSAutoreleasePool, NSDate, NSDefaultRunLoopMode, NSPoint
 use cocoa::appkit;
 use cocoa::appkit::*;
 use cocoa::appkit::NSEventSubtype::*;
-use cocoa::foundation::NSPoint;
 
 use core_foundation::base::TCFType;
 use core_foundation::string::CFString;
@@ -728,7 +727,7 @@ impl Window {
     pub fn get_cursor_position(&self) -> Result<(i32, i32), ()> {
         unsafe {
             let point = self.window.mouseLocationOutsideOfEventStream();
-            Ok((point.x, point.y))
+            Ok((point.x as i32, point.y as i32))
         }
     }
 }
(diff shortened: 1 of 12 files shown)


fkaa:
Previous discussion (https://github.com/tomaka/glutin/issues/753) mentions that wayland has no way of directly querying the mouse location, which would mean it would have to be tracked by glutin/winit.

src/api/cocoa/mod.rs
```diff
@@ -721,6 +723,17 @@ impl Window {
...
         Ok(())
     }
+
+    #[inline]
+    pub fn get_cursor_position(&self) -> Result<(i32, i32), ()> {
+        unsafe {
+            // Very unsafe, but I think it's fine as mouseLocation doesn't touch the pointer.
+            // https://github.com/servo/cocoa-rs/blob/master/src/appkit.rs#L2070-L2072
+            // TODO: Check for errors.
+            let point = NSEvent::mouseLocation(ptr::null_mut() as id);
```
fkaa: Should use [[NSWindow mouseLocationOutsideOfEventStream]](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSWindow_Class/index.html#//apple_ref/occ/instm/NSWindow/mouseLocationOutsideOfEventStream) to get the mouse location relative to the window base coordinate, rather than the screen. Might be the same for other implementations, but not really too sure.

fkaa:
Also see [Travis](https://travis-ci.org/tomaka/glutin/builds/125244437) for errors on the OSX/Linux builds.

trolleyman:
Ok, should all be fixed now

trolleyman:
Maybe not :p

vberger:
I'd like some thought given about the prototype of the added function:

Current proposal is `pub fn get_cursor_position(&self) -> Result<(i32, i32), ()>`

I guess the returned numbers are location of the pointer, relative to window coordinates. Now what is the semantics of the error case ?

To be explicit about my question: in wayland, you cannot know where the cursor is if it isn't in the window, as you only receive update events for its location when it is on your window. Even if the glutin app has several windows, knowing the position of the cursor in one window tells nothing about its location relative to other windows, as you cannot know the location of a window regarding the others.

Thus, in wayland, the `Err(..)` case would trigger quite often for non-fullscreen windows.

Wouldn't using an `Option<(i32, i32)>` be more accurate, with a clear documentation that depending on the platform, it is not always possible to query the location of the pointer if the pointer is not on the window, and that in such cases the function returns `None`?

trolleyman:
That's probably a good idea. Should the `unimplemented!` functions return `None` as well?

trolleyman:
Also, I can't seem to find a "get cursor position" type function in the wayland api. Is this because we have to process all the events to get the most recent position? Will there have to be a timeout until the function returns None, or will it just return the most recent position?

tomaka:
Using a `Result` means that we can differentiate between "the cursor is outside of the window", "the window has been closed", "there's no cursor" (for Android/iOS), and "an error happened".

I also think that the function should return "the cursor is outside of the window" for all platforms when the cursor is outside of the window, instead of returning negative coords.

vberger:
@trolleyman Yeah, as said previously by @fkaa , wayland do not provide you any way to query the cursor position, it's mandatory to process the events as they come.

vberger:
Also, I completely agree with @tomaka regarding the return type, makes a lot of sense this way :+1:

trolleyman:
So the return type should be `Result<Option<(u32, u32)>, ()>` to allow differentiation between all these cases?

tomaka:
No, it should be explicit: `Result<(u32, u32), GetCursorError>`

Where `GetCursorError` is an enum of each potential problem.

est31:
To add to @fkaa 's comment, wayland doesn't even have the notion of only one cursor, but it supports multiple cursors a time: https://www.youtube.com/watch?v=WO2L_ihO_rI

PR #769, opened 2016-04-28
Allow OSX WindowBuilder to specify 'activation behavior'.

frewsxcv:


Cargo.toml
```diff
@@ -39,7 +39,7 @@ objc = "0.2"
 [target.x86_64-apple-darwin.dependencies]
 objc = "0.2"
 cgl = "0.1"
-cocoa = "0.3"
+cocoa = "0.3.2"
```
frewsxcv: Had to bump cocoa for https://github.com/servo/cocoa-rs/pull/122. Needed the activation policy enum to implement `Copy`.

tomaka:
Thanks.

The `NSApplicationActivationPolicyRegular` type shouldn't be public, or we will need to bump glutin whenever cocoa gets bumped. It should either be a custom enum, or maybe just a bool.

tomaka:
Also `platform_specific` shouldn't be public in `WindowBuilder`, or the user could just write `builder.platform_specific.activation_policy = ...` and then wonder why it doesn't work on other platforms.

The fact that it has to go through a platform-specific trait makes it explicit that it's platform-specific. Ideally the `WindowBuilderExt` trait should directly expose members if https://github.com/rust-lang/rfcs/pull/1546 gets merged.

frewsxcv:
Result of removing the `pub` field:

```
coreyf@frewbook-pro ~/D/r/glutin (activation-policy)> cargo build
   Compiling glutin v0.5.1 (file:///Users/coreyf/Development/rust/glutin)
src/os/macos.rs:33:9: 33:31 error: field `platform_specific` of struct `window::WindowBuilder` is private
src/os/macos.rs:33         self.platform_specific.activation_policy = activation_policy;
                           ^~~~~~~~~~~~~~~~~~~~~~
error: aborting due to previous error
error: Could not compile `glutin`.

To learn more, run the command again with --verbose.
```

frewsxcv:
> The NSApplicationActivationPolicyRegular type shouldn't be public, or we will need to bump glutin whenever cocoa gets bumped. It should either be a custom enum, or maybe just a bool.

Force pushed with these changes.

frewsxcv:
IRC:

```
10:28 AM <tomaka> frewsxcv: ah, my bad
10:28 AM <tomaka> frewsxcv: the solution is to mvoe WindowBuilder to lib.rs
```

Done.

PR #770, opened 2016-05-02
Update some dependencies

mbrubeck:


kyrylo:
This syntax seems to be not TOML-compliant, so `cargo run --example window` from the README can't be run successfully:

```
% cargo run --example window
failed to parse manifest at `/Users/kyrylo/Code/glutin/Cargo.toml`

Caused by:
  could not parse input as TOML
Cargo.toml:21:9 expected a key but found an empty string
Cargo.toml:21:9-21:10 expected `.`, but found `'`
```

Should I file an issue or am I missing something obvious?

kyrylo:
Sorry for false alarm, I didn't update the Rust compiler (was using 1.6.0).

PR #772, opened 2016-05-06
Update to android_glue 0.2

ozkriff:
https://github.com/tomaka/android-rs-glue/issues/93

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ shared_library = "0.1.0"
 gl_generator = "0.5"
 
 [target.'cfg(target_os = "android")'.dependencies.android_glue]
-version = "0.1"
+version = "0.2"
 
 [target.'cfg(target_os = "ios")'.dependencies]
 objc = "0.2"
--- a/src/api/android/ffi.rs
+++ b/src/api/android/ffi.rs
@@ -4,6 +4,7 @@
 #![allow(non_upper_case_globals)]
 
 use libc;
+use std::os::raw;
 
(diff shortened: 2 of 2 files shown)


tomaka:
The small blocker here is Servo. I don't think they want to change their complex Android build system immediately.

cc @glennw @larsbergstrom

larsbergstrom:
If it's just changes to the types in the glue code, that should hopefully be fine to land.

The only change we're thinking about to our current Android build system in Servo is to move from the  deprecated `ndk-build` / `ant` system to `gradle`, which is the more modern way of building them. The new 'cargo apk' command is awesome, but I'm pretty sure that if we tried to accommodate all of Servo's manifest, java, resources, icon, etc. requirements plus the stuff needed to support finicky ndk-gdb issues, the nvidia profiler, etc., we'd end up in a similar situation to where we are with glutin where Servo is pulling in a very different direction from the needs of every other user in the community.

ozkriff:
> manifest, java, resources, icon, etc. requirements plus the stuff needed to support finicky ndk-gdb issues, the nvidia profiler, etc.

Hmm, why cargo-apk doesn`t need that? As for me, sounds like useful features for many native android apps.

ozkriff:
Ping? :)

tomaka:
@larsbergstrom @glennw If you don't disagree, I'm going to merge this PR. When you update your fork from upstream glutin, you should add a commit that reverts it.

ozkriff:
crates.io version in still not updated :(

PR #777, opened 2016-05-18
Fix Windows event handling for window close

metajack:
PostQuitMessage should not be used since there may be multiple windows
open and PostQuitMessage causes the message loop to completely
terminate. Also, after a window is closed, its HWND is gone, and so we
need to use PostThreadMessage to send the wakeup message.

### DIFF
--- a/src/api/win32/callback.rs
+++ b/src/api/win32/callback.rs
@@ -36,7 +36,7 @@ struct MinMaxInfo {
 }
 
 /// Checks that the window is the good one, and if so send the event to it.
-fn send_event(input_window: winapi::HWND, event: Event) {
+pub fn send_event(input_window: Option<winapi::HWND>, event: Event) {
     CONTEXT_STASH.with(|context_stash| {
         let context_stash = context_stash.borrow();
         let stored = match *context_stash {
@@ -46,8 +46,13 @@ fn send_event(input_window: winapi::HWND, event: Event) {
 
         let &ThreadLocalData { ref win, ref sender, .. } = stored;
 
-        if win != &input_window {
-            return;
+        match input_window {
+            Some(ref in_win) => {
+                if win != in_win {
+                    return
+                }
+            }
+            None => (),
         }
 
         sender.send(event).ok();  // ignoring if closed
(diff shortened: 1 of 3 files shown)


tomaka:
This change means that the background thread would continue to run forever.

src/api/win32/callback.rs
```diff
@@ -342,7 +332,7 @@ pub unsafe extern "system" fn callback(window: winapi::HWND, msg: winapi::UINT,
 
         x if x == *super::WAKEUP_MSG_ID => {
             use events::Event::Awakened;
-            send_event(window, Awakened);
+            send_event(None, Awakened);
```
tomaka: I'd prefer if this directly loaded the sender from the stash. Having an `Option` in the signature of `send_event` doesn't really make sense.

src/api/win32/init.rs
```diff
@@ -81,8 +81,14 @@ pub fn new_window(window: &WindowAttributes, pf_reqs: &PixelFormatRequirements,
                     break;
                 }
 
-                user32::TranslateMessage(&msg);
-                user32::DispatchMessageW(&msg);   // calls `callback` (see the callback module)
+                // check for thread message to awaken
+                if msg.message == *super::WAKEUP_MSG_ID {
+                    use events::Event::Awakened;
+                    callback::send_event(None, Awakened);
```
tomaka: Why is this handled separately instead of being handled by the callback?

metajack: callback won't get invoked once there is no window. I tried that first :)

metajack:
Shouldn't there only be one message handling thread servicing multiple Windows? In that case, the thread should survive until termination.

I believe the current design tries to do one thread per window, but this does not seem correct. However, I may be missing something as I'm no Windows expert.

If the one thread per Window is correct, then I'm not sure how we are going to handle wakeups after window close. The wait iterator will will return None (contrary to the documentation saying this isn't possible) because the sender side has been destroyed. In Servo this causes the event loop to spin consuming all the CPU assuming you just loop on None instead of panicking.

src/api/win32/callback.rs
metajack: Currently the function verifies the passed in window is the correct one and ignores other messages. Are you suggesting to just remove that filter? That would work too, but I assumed you had it there for a reason.

tomaka:
I don't think there's a reason why we currently need one thread per window, but I know that several win32 functions apply to *the window in the current thread" (for example, changing the cursor when the window is hovered). If one thread is used for multiple windows, we might be blocked.

The situation about that `Awakened` event did not get a lot of thoughts, because I've always considered it as a huge hack and only Servo uses it.

Issue #779, opened 2016-05-23
impl Clone for WindowBuilder


jansegre:
Doing this would be useful for this my use case: I want a context `with_multisampling`, but it is not a hard requirement, so I simply end up trying `build_glium`, and if it `Err`s I have to set up a new `WindowBuilder` from scratch. It would be more ergonomic to set up my minimal required `WindowBuilder` and re use it (with `.clone()`).

What do you think about this?

tomaka:
Should be easy to implement.

LFalch:
Shouldn't this be closed now as of #780?

PR #780, opened 2016-05-23
Make `WindowBuilder` and `HeadlessRendererBuilder` derive `Clone`.

jansegre:
This should deal with #779. I'm somewhat uneasy it touches platform specific code. It is really trivial tough.


jansegre:
Forgot to mention, I included `HeadlessRendererBuilder` for symmetry.

tomaka:
Thanks

PR #792, opened 2016-06-22
Remove global imports from cocoa API in favour of explicitly importing traits and using namespacing for constants, types and functions.

mitchmindtree:
I was having trouble tracking down the source of many of the names (and their associated docs as a result) in the cocoa api module, so removed the global imports to try and make it a bit clearer.

I know opinions on global imports can differ between folks, so feel free to close this if you like, just thought I'd share.


tomaka:
Needs a rebase as it conflicts with one of your other PRs.

mitchmindtree:
Okydoke, rebasing done

PR #794, opened 2016-06-22
Rather than returning `Event::Awakened` for every unknown event type, only return one when a `NSApplicationActivatedEventType` is received (this is the event produced by `WindowProxy::wakeup_event_loop`)

mitchmindtree:
The original behaviour returns an `Event::Awakened` for every unknown event type (see [here](https://github.com/tomaka/glutin/blob/0c4cf714a103bf2fa67c77f8f55111ae78bc0e6d/src/api/cocoa/mod.rs#L266-L267)). This PR changes the behaviour so that rather than returning `Event::Awakened` if `NSEventToEvent` returns `None`, we instead `loop` until `NSEventToEvent` returns `Some(Event)`.

The `wait_events` docs mention that the `Iterator` should never return `None`, so it's possible that the returning of `Event::Awakened` might have originally been a hack to avoid returning `None` when `NSEventToEvent` returns `None`.

This may be a **breaking change** for downstream programs that depend on receiving `Awakened` on OS X for some reason (though they probably shouldn't, as the current behaviour does not seem to reflect the `Awakened` docs at all). _Edit: This PR still returns `Awakened`, but only when the event loop is actually awakened_.

I've tested the changes with the window.rs example (which uses the `WaitEventsIterator`) and it seems to handle all the events I can throw at it from my macbook pro without issues, still sitting between 0.0 - 0.1% cpu when idle. The only difference being it no longer returns seemingly random `Awakened` events. I'm on OS X 10.10.5.

@paulrouget sorry to bug you! I noticed you occasionally review cocoa-related PRs - would you mind reviewing this?

### DIFF
--- a/src/api/cocoa/mod.rs
+++ b/src/api/cocoa/mod.rs
@@ -243,28 +243,29 @@ impl<'a> Iterator for WaitEventsIterator<'a> {
     type Item = Event;
 
     fn next(&mut self) -> Option<Event> {
-        if let Some(ev) = self.window.delegate.state.pending_events.lock().unwrap().pop_front() {
-            return Some(ev);
-        }
+        loop {
+            if let Some(ev) = self.window.delegate.state.pending_events.lock().unwrap().pop_front() {
+                return Some(ev);
+            }
 
-        let event: Option<Event>;
-        unsafe {
-            let pool = NSAutoreleasePool::new(nil);
+            let event: Option<Event>;
 
-            let nsevent = appkit::NSApp().nextEventMatchingMask_untilDate_inMode_dequeue_(
-                appkit::NSAnyEventMask.bits() | appkit::NSEventMaskPressure.bits(),
-                NSDate::distantFuture(nil),
-                NSDefaultRunLoopMode,
-                YES);
-            event = NSEventToEvent(self.window, nsevent);
+            unsafe {
+                let pool = NSAutoreleasePool::new(nil);
 
-            let _: () = msg_send![pool, release];
-        }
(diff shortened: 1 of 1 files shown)


paulrouget:
@pcwalton do you mind looking at this?

pcwalton:
Can you test to make sure you still return from the event loop when you send `wakeup_event_loop`?

mitchmindtree:
@pcwalton I just added a commit that checks for the specific `NSEventSubtype::NSApplicationActivatedEventType` in order to return `Event::Awakened` when a `WindowProxy` calls `wakeup_event_loop`.

I created a test as you described that called `wakeup_event_loop` from a `WindowProxy` on a separate thread. Here's the code I used to do the test:

```rust
extern crate glutin;

mod support;

fn main() {
    let mut window = glutin::WindowBuilder::new().build().unwrap();
    window.set_title("A fantastic window!");
    let _ = unsafe { window.make_current() };

    let context = support::load(&window);

    let proxy = window.create_window_proxy();
    std::thread::spawn(move || {
        std::thread::sleep(std::time::Duration::from_secs(3));
        proxy.wakeup_event_loop();
    });

    for event in window.wait_events() {
        context.draw_frame((0.0, 1.0, 0.0, 1.0));
        let _ = window.swap_buffers();

        println!("{:?}", event);

        match event {
            glutin::Event::Closed => break,
            _ => ()
        }
    }
}
```

Using this code, the window successfully received a `Event::Awakened` after three seconds.

pcwalton:
Looks good to me! r=me
cc @tomaka @paulrouget

mitchmindtree:
Closing as winit is the proper place to address this now.

Issue #778, opened 2016-05-19
Segfault on exit on OS X when running window example


jbrd:
Am running on OS X 10.10.5 (Yosemite), with rustc 1.8.0. 

Running the window example crashes on exit. Steps to reproduce:

- Clone glutin and check out latest master branch
- Run `cargo run --example window`
- Close the window

Upon closing the window, the example will segfault. Details as follows:

```
Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000018

VM Regions Near 0x18:
--> 
    __TEXT                 00000001038b4000-0000000103938000 [  528K] r-x/rwx SM=COW  /Users/USER/*

Application Specific Information:
objc_msgSend() selector name: setDelegate:


Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libobjc.A.dylib               	0x00007fff8bf6e0dd objc_msgSend + 29
1   window                        	0x00000001038c8a98 glutin::message::send_unverified<objc::runtime::Object,(*mut objc::runtime::Object),()> + 88 (mod.rs:187)
2   window                        	0x00000001038ca88a api::cocoa::WindowDelegate.Drop::drop::had74467e0a12df1dQJa + 202 (<objc macros>:18)
3   window                        	0x00000001038b5bad glutin..api..cocoa..WindowDelegate::drop.9621::h4035d12b5366011b + 61
4   window                        	0x00000001038b5ad1 glutin..api..cocoa..Window::drop.9599::h56518e6bf43ec835 + 49
5   window                        	0x00000001038b5a96 glutin..Window::drop.9596::h2c081b64f8945d0a + 6
6   window                        	0x00000001038b5896 window::main + 1302 (window.rs:37)
7   window                        	0x00000001038f38e3 sys_common::unwind::try::try_fn::h662774255089384887 + 35
8   window                        	0x00000001038f214c __rust_try + 12
9   window                        	0x00000001038f37a1 rt::lang_start::h06c4303e25b79660ujz + 497
10  window                        	0x00000001038ba03a main + 42
11  libdyld.dylib                 	0x00007fff902b15c9 start + 1
```

Can anybody else repro this issue?

jbrd:
Dependency info:

```
   Compiling bitflags v0.7.0
   Compiling lazy_static v0.2.1
[...]

PR #795, opened 2016-06-22
cocoa - Change `IdRef` `Drop` implementation to use `autorelease` instead of `release`. Fixes Yosemite segfault that occurs upon `Closed`.

mitchmindtree:
This still relinquishes ownership over the object when `IdRef` is dropped, however leaves the actual releasing of the object until the enclosing autorelease pool is released (which is the end of main if none are explicitly instantiated any higher in the stack). All of the objects wrapped by `IdRef` within our cocoa module exist for the lifetime of the `Window` anyway, so the difference between `release` and `autorelease` should be negligible.

This fixes the segfault that occurs when closing the window on OS X Yosemite (see #778). I've tested this in the `window.rs` example as well as a downstream conrod example and both no longer crash following the `Closed` event.

@fkaa @pcwalton I'm still very new to obj-c, would appreciate your thoughts

### DIFF
--- a/src/api/cocoa/mod.rs
+++ b/src/api/cocoa/mod.rs
@@ -799,7 +799,7 @@ impl IdRef {
 impl Drop for IdRef {
     fn drop(&mut self) {
         if self.0 != nil {
-            let _: () = unsafe { msg_send![self.0, release] };
+            let _: () = unsafe { msg_send![self.0, autorelease] };
         }
     }
 }


pcwalton:
This seems like it's papering over a deeper memory management issue. What is the exact life cycle of these objects?

mitchmindtree:
I'm AFK atm but will dig into this further tomorrow. I left some findings at  #778 a little earlier in case you're interested

mitchmindtree:
@pcwalton I did some more research into the exact lifetimes of each of the `IdRef` objects and left my results at the relevant segfault issue [here](https://github.com/tomaka/glutin/issues/778#issuecomment-227974237). It would be great to get your thoughts on what's going or hear if you have any ideas on how to proceed.

SSheldon:
@mitchmindtree, I read up on #778 and commented there too. The overrelease appears to be caused by the window releasing itself on close, which we can fix by setting `releasedWhenClosed` to `NO` 😊

I'd prefer that fix over changing these releases to autoreleases; I think making these autoreleases actually leaks the memory, because when drop is run on the `IdRef`s we aren't inside of an `NSAutoreleasePool`.

mitchmindtree:
@SSheldon I just read your follow up on #778 - amazing work tracking it down! Totally agreed, that sounds like a much better fix, I'll close this in favour of opening another PR. Thanks for the awesome detective work 😸

PR #796, opened 2016-06-22
Fix stability issues of HeadlessContext on Mac OS X

davll:
The PR fixed the issues below:
- Cocoa objects leak when instances of `HeadlessContext` are dropped, since `id` instances are actually raw pointers and therefore won't release in the end of lifetime.
- Sometimes headless test fails on initialization phase because `[NSApplication sharedApplication]` is not called when creating `NSOpenGLContext`.
- `GlContext.is_current()` is broken (unimplemented...)

### DIFF
--- a/src/api/cocoa/headless.rs
+++ b/src/api/cocoa/headless.rs
@@ -5,6 +5,8 @@ use GlAttributes;
 use GlContext;
 use PixelFormatRequirements;
 
+use super::IdRef;
+
 use core_foundation::base::TCFType;
 use core_foundation::string::CFString;
 use core_foundation::bundle::{CFBundleGetBundleWithIdentifier, CFBundleGetFunctionPointerForName};
@@ -17,7 +19,7 @@ use api::cocoa::helpers;
 pub struct PlatformSpecificHeadlessBuilderAttributes;
 
 pub struct HeadlessContext {
-    context: id,
+    context: IdRef,
 }
 
 impl HeadlessContext {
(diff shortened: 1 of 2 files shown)


src/api/cocoa/headless.rs
```diff
@@ -90,5 +92,14 @@ impl GlContext for HeadlessContext {
     }
 }
 
+impl Drop for HeadlessContext {
+    fn drop(&mut self) {
+        if self.is_current() {
+            unsafe { NSOpenGLContext::clearCurrentContext(nil) };
+        }
+        unsafe { msg_send![self.context, release] };
```
SSheldon: There's an `IdRef` struct defined in the parent module that'd be perfect for this; just wrap the context in `IdRef::new` when it's being allocated and it'll be released automatically whenever it goes out of scope or the struct containing it is dropped.

src/api/cocoa/headless.rs
```diff
@@ -35,6 +36,7 @@ impl HeadlessContext {
                 return Err(OsError(format!("Could not create the pixel format")));
             }
             let context = NSOpenGLContext::alloc(nil).initWithFormat_shareContext_(pixelformat, nil);
+            msg_send![pixelformat, release];
```
SSheldon: Similarly, I think using `IdRef` here would also be a good call. You could put that right where the pixel format is allocated, and then you're guaranteed that it'll get released by the end of the scope, even if someone were to add an early return or there's a panic or etc.

davll:
@SSheldon 
OK, I wrapped internal cocoa objects with `IdRef` as you mentioned.

davll:
Is it planned to improve type safety of `IdRef` or `id`?
I don't think they are type-safe, someone might misuse a object with class A as class B and therefore lead to severe bugs.

PR #797, opened 2016-06-26
cocoa - Ensure the window does not set itself to `autorelease` upon closing as we release it manually via IdRef's drop method.

mitchmindtree:
This is a better alternative to #795 discovered by @SSheldon that fixes the segfault discussed at #778.

Tested on OS X 10.10.5.


SSheldon:
Looks good to me 👍

PR #798, opened 2016-06-30
owner_events is now true when grabbing the pointer

### LINKED ISSUE #696: Mouse Grabbing and MouseMoved Event
I'm not sure that it is right, that there are no MouseMoved Events, while the Mouse is grabbed?

Only tested on x11

See grabbing.rs example

mraof:
Fixes #696


tomaka:
Thanks

PR #800, opened 2016-07-04
x11: handle X11's multi-key, a.k.a compose key.

emilio:
This is needed for servo to land non-qwerty keyboard support.

r? @tomaka


tomaka:
Why is that a separate event and not a virtual key?

emilio:
What do you mean? It's an x keycode, and I added it as a `VirtualKeyCode` enum variant.

I'm not sure if it's X-specific, unfortunately. I'd expect wayland to also have something like this at least, and probably OSX too, but I haven't got machines to test.

tomaka:
Humm I think I'm becoming crazy.

PR #804, opened 2016-08-17
Added two more PartialEq's

VictorKoenders:


tomaka:
It doesn't really make sense to have an order for mouse cursors and mouse states.

VictorKoenders:
I'm my project I have a desired_cursor_state and i check if that's different from my dirt current cursor state.

Is there a reason to not make them PartialEq?

tomaka:
Oh I misread this as `PartialOrd`, sorry.

PR #806, opened 2016-08-21
Add support to Linux headless implementation for specifying OpenGL version.

glennw:
This allows creation of GL3+ headless OSMesa contexts on Linux.


tomaka:
What happens if I request OpenGL 6.2? Will it panic?

glennw:
It will return the same error as it previously did when context creation fails (https://github.com/tomaka/glutin/pull/806/files#diff-0c4e87795d81babd64a9dd953c4a9590R103)

PR #816, opened 2016-10-08
Emscripten fix

badboy:
As I'm currently trying out more and more emscripten things, I rely on glutin & glium for graphic things.
Here are some minimal changes necessary to make it compile and usable.


tomaka:
Thanks!

PR #818, opened 2016-10-14
Fix X11 WaitEventsIterator busy loop

jwilm:
Certain events were not being checked for in the PollEventsIterator
and would be left in the X11 queue forever. XPeekEvent in the
WaitEventsIterator would never block because an item was still in the
queue causing a busy loop. The problem is solved by using XNextEvent
in PollEventsIterator with a check to make sure it will never block.

I added the log crate for debugging, but realize now that you might not want it in glutin. Lmk if you want it removed or feel free to patch on top of this.

### DIFF
--- a/src/api/x11/window.rs
+++ b/src/api/x11/window.rs
@@ -181,18 +181,26 @@ impl<'a> Iterator for PollEventsIterator<'a> {
             }
 
             let mut xev = unsafe { mem::uninitialized() };
-            let res = unsafe { (xlib.XCheckMaskEvent)(self.window.x.display.display, -1, &mut xev) };
 
-            if res == 0 {
-                let res = unsafe { (xlib.XCheckTypedEvent)(self.window.x.display.display, ffi::ClientMessage, &mut xev) };
-
-                if res == 0 {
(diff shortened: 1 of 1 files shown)


src/api/x11/window.rs
```diff
@@ -181,18 +181,29 @@ impl<'a> Iterator for PollEventsIterator<'a> {
...
+            // types of out the queue to prevent busy looping when there's an
+            // event of another type glutin doesn't care about.
+            unsafe {
+                let count = (xlib.XPending)(self.window.x.display.display);
+                if count == 0 {
+                    return None;
                 }
-            }
+
+                let res = (xlib.XNextEvent)(
```
tomaka: Nit: everything should be on the same line.

tomaka:
Thanks for the PR.

Please remove the `log` and `bitflags` crate however.

jwilm:
Oh, didn't realize bitflags wasn't for this patch :laughing:. I originally built this on top of another branch with some other features..

Anyway, fixed both the nit and the extra deps. Should be good to go.

jwilm:
Should I redirect this PR over to winit?

tomaka:
@jwilm Yes please

PR #822, opened 2016-11-02
Implement display info for Android.

wvdschel:
To be used as example code for reviewing https://github.com/tomaka/android-rs-glue/pull/116/commits/9986312ba1634cb771aa62ee71767fdf5f445452


francesca64:
If you're still interested, this would have to be moved to [winit](https://github.com/tomaka/winit). Though, is this blocked by tomaka/android-rs-glue#116?

PR #831, opened 2016-11-24
Remove assertions now that macOS window constraints are implemented.

frewsxcv:
As per https://github.com/tomaka/winit/pull/89


tomaka:
These asserts should have been removed earlier.

PR #832, opened 2016-11-24
Always set the swap interval on GLX

jrmuizel:
This matches the behaviour of the WGL backend and allows
toggling vsync in situations where the swap interval defaults
to 1

### DIFF
--- a/src/api/glx/mod.rs
+++ b/src/api/glx/mod.rs
@@ -228,50 +228,49 @@ impl<'a> ContextPrototype<'a> {
             },
         };
 
+        let interval = if self.opengl.vsync { 1 } else { 0 };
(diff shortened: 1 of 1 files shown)


tomaka:
The problem is that "system default" should be a valid option as well.

This has been something that I've wanted to change for some time. In my opinion the `vsync` option should be an enum with three variants (enabled, disabled, system default).

jrmuizel:
So when I first tested this I must have made a mistake because it doesn't actually work. 0 is not a valid value to pass to SwapIntervalSGI and the mesa drivers don't support SwapIntervalExt. The only way to disable vsync with mesa is through SwapIntervalMESA.

Issue #839, opened 2016-12-15
Integrate with `winit`


kvark:
I want the following chain of conversions to be accessible:
`winit::WindowBuilder` -> `glutin::WindowBuilder` -> `glutin::Window` -> `winit::Window`

The first link is needed for the window constructors to have a unified interface: getting `winit::WindowBuilder` on the input and working from it.

The last link is needed for the window-agnostic frameworks/apps, which just get `winit::Window` and from it use the events as well as `swap_buffers`.

This is a blocker for https://github.com/gfx-rs/gfx/issues/907
cc @ozkriff

Plasticcaz:
I'm looking for something to do, maybe I'll have a go at this.

kvark:
@Plasticcaz thanks for taking a bite at it!
If the work is non trivial, and you'd like to get the feedback sooner rather than later, then please by all means submit a PR with [WIP] before the name.

Plasticcaz:
Thanks... 

Looking at src/platform and src/api/android, it's obvious to me how I could easily expose winit_window from a window in Windows, Android, and MacOS, but I'm not sure with the other platforms... 

Linux appears to not keep the winit window... It looks like it takes what it needs, and then throws it out.

francesca64:
Is this still an issue?

kvark:
I believe we are happy now :)

PR #840, opened 2016-12-18
Integrate with `winit'

Plasticcaz:
This is an attempt at solving issue #839. 

So far, I have added a 'constructor' method on glutin::WindowBuilder that takes in a winit::WindowBuilder.

I also have put methods on the Window structs for Windows, MacOS, and Android to convert a Window into a winit::Window. I am unsure how to achieve this on Linux, as the Linux platform code does not appear to keep a winit::Window inside its Window struct.

Another thing that needs to be taken into consideration is that currently the methods to get the winit::Window from the platform Window consume the platform Window. I am unsure if this is the desired behavior or not, as I am not familiar with the exact reasons we did this.

These changes were by no means difficult, but I appreciate being allowed to make them.

Please let me know any points I missed, or anything I should do differently.

### DIFF
--- a/src/api/android/mod.rs
+++ b/src/api/android/mod.rs
@@ -33,25 +33,23 @@ pub struct PlatformSpecificWindowBuilderAttributes;
 pub struct PlatformSpecificHeadlessBuilderAttributes;
 
 impl Window {
-    pub fn new(
-        _: &WindowAttributes,
-        pf_reqs: &PixelFormatRequirements,
-        opengl: &GlAttributes<&Window>,
-        _: &PlatformSpecificWindowBuilderAttributes,
-        winit_builder: winit::WindowBuilder,
-    ) -> Result<Window, CreationError> {
+    pub fn new(_: &WindowAttributes,
+               pf_reqs: &PixelFormatRequirements,
+               opengl: &GlAttributes<&Window>,
+               _: &PlatformSpecificWindowBuilderAttributes,
+               winit_builder: winit::WindowBuilder)
+               -> Result<Window, CreationError> {
         let winit_window = winit_builder.build().unwrap();
         let opengl = opengl.clone().map_sharing(|w| &w.context);
         let native_window = unsafe { android_glue::get_native_window() };
         if native_window.is_null() {
             return Err(OsError(format!("Android's native window is null")));
         }
-        let context = try!(EglContext::new(
-            egl::ffi::egl::Egl,
-            pf_reqs,
-            &opengl,
-            egl::NativeDisplay::Android
-        ).and_then(|p| p.finish(native_window as *const _)));
+        let context = try!(EglContext::new(egl::ffi::egl::Egl,
+                                           pf_reqs,
+                                           &opengl,
+                                           egl::NativeDisplay::Android)
+            .and_then(|p| p.finish(native_window as *const _)));
         Ok(Window {
             context: context,
             winit_window: winit_window,
@@ -111,6 +109,11 @@ impl Window {
 
     pub unsafe fn platform_display(&self) -> *mut libc::c_void {
         self.winit_window.platform_display()
+    }
+
+    #[inline]
+    pub fn winit_window(self) -> winit::Window {
+        self.winit_window
     }
 
     pub unsafe fn platform_window(&self) -> *mut libc::c_void {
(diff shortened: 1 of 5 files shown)


tomaka:
Thanks for the PR.

- I'm not a fan of sneaking code style changes in other PRs. I've been accepting a bit everything since I've always thought that eventually rustfmt would come and solve everything. But time passes and rustfmt still isn't an acceptable solution. Anyway, two thirds of your PR are code style changes, I'm still going to accept it but it crosses a bit the line.

- The Linux window is here: https://github.com/tomaka/glutin/blob/master/src/platform/linux/api_dispatch.rs#L23

- `to_winit_window` should be named `into_winit_window` because you're taking ownership of the window.

- The PR also needs `as_winit_window` that takes a `&self` and returns a reference to the window.

- You also need to add the two functions to the `Window` function at the root of the crate.

Plasticcaz:
Yeah, I apologize. I have rustfmt on autorun-on-save in Visual Studio code. I did notice the style changes after I committed.

I shall make the other changes you suggested.

Plasticcaz:
I have made the changes you have suggested, but there is one thing that is giving me a "What on Earth moment.

`
pub fn into_winit_window(self) -> winit::Window {
    self.window.into_winit_window()
}
`
The above method on the Window struct in src/window.rs is giving me a strange error. It is telling me that I "Cannot move out of borrowed content", which is strange, because I don't see any borrowing going on in that function.

I am developing on Windows, so I double checked that I didn't make window.into_winit_window() borrow self, and I don't, so I think it's safe to assume that the function call does not result in a borrow. I admit I'm a bit stumped.

Do we need the ability to take ownership of the winit::Window, @kvark?

mitchmindtree:
@Plasticcaz I think the error you're getting is because `window` is still owned by `self` when attempting to call the `self.window.into_winit_window()` method. If you do something along these lines it should work:

```rust
pub fn into_winit_window(self) -> winit::Window {
    let Window { window, .. } = self;
    window.into_winit_window()
}
```

Plasticcaz:
@mitchmindtree, I appreciate the advice. I wasn't aware that we had struct deconstruction like that. I guess I've never had to use it until now. I tried that, but it's still giving me the borrow error.

`
pub fn into_winit_window(self) -> winit::Window {
        let Window { window } = self;
        window.into_winit_window()
}
`

If you look into the platform window's implementation of into_winit_window(), you'll see that I simply return self.winit_window, and the compiler throws the rest of the fields away automatically, I guess:

`
    // in src/platform/windows/window.rs:89 in my latest commit
    pub fn into_winit_window(self) -> winit::Window {
        self.winit_window
`
This is really confusing me. I thought I had a good handle on borrow rules, and that tackling this issue, was probably too easy for me, lol.

kvark:
Fyi, I don't need ownership. A mutable borrow would be sufficient.

mitchmindtree:
Ah I just realised my advice is totally unnecessary. I tested `self.window.into_winit_window()` and it seems to work fine for me. I thought rustc wasn't smart enough to infer that this move is OK but I was wrong.

@Plasticcaz I just tested your fork and it seems to work fine for me when I remove the `unimplemented!()` and re-add your commented code at src/window.rs line 483.

Plasticcaz:
That's weird. I just realized I'm running Rustc 1.12 (64 bit msvc) on this machine. The latest is 1.13 isn't it? Maybe that is the cause of the error I was getting?

Anyway, as @kvark has stated he doesn't need ownership of the winit window, I don't think it's worth investigating further in this instance.

kvark:
Please use the standard Rust mechanisms of type casting: `AsRef/AsRefMut`.

Plasticcaz:
So, you want me to return self.winit_window.as_ref() instead of &self.winit_window, @kvark?

kvark:
@Plasticcaz no, I want you to implement `AsRef<winit::Window>` and `AsMut<winit::Window>` instead of having the separate methods like `as_winit_window` or `as_winit_window_mut`

tomaka:
Err. I liked that these methods are a bit "hidden".
I wanted to add a big warning saying how the user must specifically use winit 0.x in order for it to work and that the situation will improve when private/shared dependencies are added to Cargo.

Plasticcaz:
Ah, so implement the traits on the window type... I see.

@tomaka are you saying we shouldn't do that?

Also if I'm keeping the methods, I should put a warning that the user should use the same version of winit that glutin is using?

kvark:
@tomaka if you want to make them less visible, you could have your own `AsWindow` trait or something

kvark:
@tomaka ping. I think this is stalled, and is awaiting your reaction

tomaka:
Well I didn't consider merging since it's still marked as `[WIP]`, but otherwise it looks good.

kvark:
@tomaka I believe it's not clear what to do with `as_winit_window`. I suggested to implement `AsRef<winit::Window>`, and you expressed a concern that this is a bit too exposed. So I suggested having `glutin::AsWindow` trait. So, what would you prefer?

Plasticcaz:
Yeah, I was waiting on @tomaka's input on what you asked for, @kvark.

tomaka:
I prefer the way it currently is.

Plasticcaz:
If that's the way it is, then we can consider this no longer WIP.

I'll still check back tomorrow.

Thanks for allowing me to make these simple changes. I feel like I've learned a little about the process of making pull requests and working on a library, even if the changes were basically trivial.

kvark:
@tomaka you must be kidding me... What difference does it make for you to have `as_winit_window` in any generic way (either through `AsRef/AsMut`, or your own `AsWindow`)? You are not going to be using this PR's logic, I get it, but since you did bother to at least respond here, why not make this at least satisfy our needs, especially since it's not actually you doing the code (thanks @Plasticcaz).

tomaka:
If someone writes `fn foo<T: AsRef<winit::Window>>(t: &T) {}` in order to match `Arc<Window>` or `&Window`, I don't want it to automatically match a glutin window.

kvark:
@tomaka 
  1. in what use case would matching `glutin::Window` to `winit::Window` be harmful?
  2. what's wrong with proposed `AsWindow` trait?

tomaka:
1. It can be harmful if the user tries to use for example Vulkan on a glutin window.
But that's not the primary reason. The primary reason is this problem where the user uses a different version of winit than glutin does. I want the conversion to a winit window to be explicit and not implicit.

2. Nothing wrong with it, but I don't think it's better than a regular method either. It's specifically the semantics of using `AsRef` that I want to avoid.

PR #843, opened 2017-01-06
Create winit window with correct XVisual and screen when using X11

rerion:
My attempt at solving tomaka/glutin#834
See also tomaka/winit#106

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ build = "build.rs"
 lazy_static = "0.2.0"
 libc = "0.2"
 shared_library = "0.1.0"
-winit = { git = "https://github.com/rerion/winit" }
+winit = "0.5.7"
 
 [build-dependencies]
 gl_generator = "0.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ build = "build.rs"
 lazy_static = "0.2.0"
 libc = "0.2"
 shared_library = "0.1.0"
-winit = "0.5.6"
+winit = { git = "https://github.com/rerion/winit" }
 
 [build-dependencies]
 gl_generator = "0.5"
--- a/src/platform/linux/api_dispatch.rs
+++ b/src/platform/linux/api_dispatch.rs
@@ -1,5 +1,6 @@
 use libc;
 use winit;
+use wayland_client;
 
 use ContextError;
 use CreationError;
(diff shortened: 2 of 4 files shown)


src/platform/linux/x11.rs
```diff
@@ -109,10 +110,20 @@ impl Window {
...
-    ) -> Result<Window, CreationError> {
-        let display = winit_window.get_xlib_xconnection().unwrap();
-        let screen_id = winit_window.get_xlib_screen_id().unwrap() as _;
+        winit_builder: winit::WindowBuilder,
+    ) -> Result<(Window, winit::Window), CreationError> {
+        use api::glx::ffi::Xlib;
+        use winit::NativeMonitorId;
+
+        let xlib = Xlib::open().unwrap();
+        let display = unsafe { (xlib.XOpenDisplay)(ptr::null()) };
```
tomaka: We should gather the X display from winit instead of opening one here. This would require a new function in winit.

src/platform/linux/x11.rs
```diff
@@ -109,10 +110,20 @@ impl Window {
     pub fn new(
         pf_reqs: &PixelFormatRequirements,
         opengl: &GlAttributes<&Window>,
-        winit_window: &winit::Window,
-    ) -> Result<Window, CreationError> {
-        let display = winit_window.get_xlib_xconnection().unwrap();
-        let screen_id = winit_window.get_xlib_screen_id().unwrap() as _;
+        winit_builder: winit::WindowBuilder,
+    ) -> Result<(Window, winit::Window), CreationError> {
+        use api::glx::ffi::Xlib;
```
tomaka: Please put imports at the top of the file, not inside functions.

src/platform/linux/api_dispatch.rs
```diff
@@ -75,33 +76,39 @@ impl Window {
...
-            let opengl = opengl.clone().map_sharing(|w| match w.display_server {
-                DisplayServer::Wayland(ref w) => w,
-                _ => panic!()       // TODO: return an error
-            });
-            DisplayServer::Wayland(try!(wayland::Window::new(
-                pf_reqs,
-                &opengl,
-                &winit_window,
-            )))
+        let window = match wayland_client::default_connect() {
```
tomaka: This is not a great solution, but I guess I'll accept it for now.

Cargo.toml
```diff
@@ -14,7 +14,7 @@ build = "build.rs"
 lazy_static = "0.2.0"
 libc = "0.2"
 shared_library = "0.1.0"
-winit = "0.5.6"
+winit = { git = "https://github.com/rerion/winit" }
```
tomaka: Don't forget to change this.

src/platform/linux/x11.rs
```diff
@@ -172,19 +183,25 @@ impl Window {
                     template.visualid = p.get_native_visual_id() as ffi::VisualID;
 
                     let mut num_visuals = 0;
-                    let vi = (display.xlib.XGetVisualInfo)(display.display, ffi::VisualIDMask,
+                    let vi = (xlib.XGetVisualInfo)(display, ffi::VisualIDMask,
                                                            &mut template, &mut num_visuals);
-                    display.check_errors().expect("Failed to call XGetVisualInfo");
+                    // display.check_errors().expect("Failed to call XGetVisualInfo");
```
tomaka: Don't forget to uncomment this.

src/platform/linux/x11.rs
mitchmindtree: @tomaka Interesting, I quite like this style as it tends to clarify where the import is actually used. That said, it can be annoying to refactor to the top level in the case that it's used in more than one function, so I can see how some would prefer to always import at the top.

rerion:
I've made some changes. How does it look now? 
As to `match wayland_client::default_connect()` this is pretty much how winit chooses which backend to use. The only other thing that comes to my mind is instead of replicating that behaviour query ask winit which backend is in use, but that would require another function in winit, I think.
Also I can't change the dependancy before changes from tomaka/winit#106 are released.

tomaka:
> As to match wayland_client::default_connect() this is pretty much how winit chooses which backend to use. The only other thing that comes to my mind is instead of replicating that behaviour query ask winit which backend is in use, but that would require another function in winit, I think.

Yeah that's what I was thinking.
We may want to change how winit chooses its backend in the future, and it's bad if glutin decides to use x11 while winit used wayland for example.

But as I said this should be fixed later.

PR #845, opened 2017-01-11
Readd Clone to WindowBuilder

Jascha-N:
Adds the `Clone` auto-derive back to `WindowBuilder` after it was somehow removed in #819. Only tested on Windows.


tomaka:
Thanks

PR #846, opened 2017-01-19
Check for libEGL.dll

kvark:
This is effectively all we need to support Angle.

### DIFF
--- a/src/platform/windows/mod.rs
+++ b/src/platform/windows/mod.rs
@@ -32,25 +32,27 @@ lazy_static! {
     // An EGL implementation available on the system.
     static ref EGL: Option<EglWrapper> = {
         // the ATI drivers provide an EGL implementation in their DLLs
-        let dll_name = if cfg!(target_pointer_width = "64") {
+        let ati_dll_name = if cfg!(target_pointer_width = "64") {
             b"atio6axx.dll\0"
         } else {
             b"atioglxx.dll\0"
         };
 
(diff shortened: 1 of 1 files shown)


src/platform/windows/mod.rs
```diff
@@ -32,25 +32,28 @@ lazy_static! {
...
             b"atioglxx.dll\0"
         };
 
-        let dll = unsafe { kernel32::LoadLibraryA(dll_name.as_ptr() as *const _) };
+        for dll_name in [b"libEGL.dll\0", ati_dll_name] {
+            let dll = unsafe { kernel32::LoadLibraryA(dll_name.as_ptr() as *const _) };
+            if dll.is_null() {
+                continue;
+            }
+            info!("Loading EGL symbols from {}", dll_name);
```
tomaka: Glutin hasn't been using logging stuff so far, so it doesn't make sense to add it here in particular.

kvark:
@tomaka I strongly suggest to use logging. There is nearly not enough variety of error messages, and even if you extend that - would be great to have some extra information on what glutin is doing. This PR's logic is a good example of why logging would be helpful.
Anyhow, I just removed the `info!` usage.

tomaka:
Usually I think that using log reveals that the design of the library is wrong.
But glutin and other similar libraries try to "just work" by being a big magic instead of having a very precise behavior and exposing every single possible situation, so I guess it makes sense to use logging for them.

tomaka:
Looks like winit broke even though it passed CI recently.

PR #850, opened 2017-01-24
Call [NSOpenGLContext update] on window size changes.

ojacobson:
OS X OpenGL applications must call <https://developer.apple.com/reference/appkit/nsopenglcontext/1436135-update> on window size changes. This ensures that the window and the OpenGL framebuffer share the same dimensions - apparently calling `glViewport` isn't sufficient for this.

This builds on tomaka/winit#118.

### DIFF
--- a/src/platform/macos/mod.rs
+++ b/src/platform/macos/mod.rs
@@ -59,7 +59,7 @@ impl<'a> Iterator for WaitEventsIterator<'a> {
     fn next(&mut self) -> Option<winit::Event> {
         let event = self.winit_iterator.next();
         match event {
-            Some(winit::Event::Resized(_, _)) => self.window.update_context(),
+            Some(winit::Event::Resized(_, _)) => unsafe { self.window.context.update() },
             _ => {},
         }
         event
@@ -77,7 +77,7 @@ impl<'a> Iterator for PollEventsIterator<'a> {
     fn next(&mut self) -> Option<winit::Event> {
         let event = self.winit_iterator.next();
         match event {
-            Some(winit::Event::Resized(_, _)) => self.window.update_context(),
+            Some(winit::Event::Resized(_, _)) => unsafe { self.window.context.update() },
             _ => {},
         }
         event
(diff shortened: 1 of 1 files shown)


src/platform/macos/mod.rs
```diff
@@ -49,6 +48,42 @@ pub struct Window {
...
+    winit_iterator: winit::WaitEventsIterator<'a>,
+}
+
+impl<'a> Iterator for WaitEventsIterator<'a> {
+    type Item = winit::Event;
+
+    fn next(&mut self) -> Option<winit::Event> {
+        let event = self.winit_iterator.next();
+        match event {
+            Some(winit::Event::Resized(_, _)) => self.window.update_context(),
```
tomaka: Please call directly `self.context.update()` here, no need for a separate function.

src/platform/macos/mod.rs
```diff
@@ -49,6 +48,42 @@ pub struct Window {
...
+    winit_iterator: winit::PollEventsIterator<'a>,
+}
+
+impl<'a> Iterator for PollEventsIterator<'a> {
+    type Item = winit::Event;
+
+    fn next(&mut self) -> Option<winit::Event> {
+        let event = self.winit_iterator.next();
+        match event {
+            Some(winit::Event::Resized(_, _)) => self.window.update_context(),
```
tomaka: Same

src/platform/macos/mod.rs
```diff
@@ -151,6 +186,12 @@ impl Window {
         }
     }
 
+    fn update_context(&self) {
```
tomaka: And remove this.

src/platform/macos/mod.rs
ojacobson: Done!

The call is slightly longer - `unsafe { self.window.context.update() }` - but I can copy a reference to `context` into the iterator instead of a ref to the window, if that reads better.

tomaka:
r+
Please ping me when CI passes (can take some time because OS X builders are usually busy).

ojacobson:
Will do.

ojacobson:
Tests have passed. It's probably worth holding off on merging this until you're happy with the corresponding winit pull request, though.

Issue #695, opened 2016-01-02
Fix the interaction between the nVidia prietary driver and GLX


tomaka:
See #694

dariost:
The best way to do this seems to be like [glxinfo](https://cgit.freedesktop.org/mesa/demos/tree/src/xdemos/glxinfo.c#n268) does it, which is to try every possibile OpenGL version starting from the latest.

francesca64:
This was fixed by #858

PR #858, opened 2017-02-17
Fix the interaction between the NVIDIA proprietary driver and GLX

dariost:
This fixes issue #695
I've tested it with the NVIDIA proprietary driver and with LLVMpipe
It tries all OpenGL versions, like glxinfo does


tomaka:
Looks good.
Could you add some comments over the `for` loop to explain that we try every single OpenGL version because of non-compliant drivers?

dariost:
Done

tomaka:
Thanks!
As usual travis-OSX takes forever.

PR #865, opened 2017-03-04
Fix transparency for macOS and glx

### LINKED ISSUE #639: transparent example does not work on Arch Linux with X11 and Intel driver
I am using compton as my compositor, yet the window background is solid black:
2015-10-15-094630_1440x900_scrot

The reported pixel format is:

```
Pixel format of the window: PixelFormat { hardware_accelerated: true, color_bits: 24, alpha_bits: 8, depth_bits: 24, stencil_bits: 8, stereoscopy: false, double_buffer: true, multisampling: None, srgb: false }
```

And OpenGL version:

```
OpenGL version 3.3 (Core Profile) Mesa 11.0.3
```

Any ideas?

mystal:
The transparent example currently gives opaque windows on macOS and glx. This PR fixes both cases.

Fixes #639

### DIFF
--- a/src/api/glx/mod.rs
+++ b/src/api/glx/mod.rs
@@ -490,7 +490,6 @@ unsafe fn choose_fbconfig(glx: &ffi::glx::Glx, extensions: &str, xlib: &ffi::Xli
         let mut num_configs = 1;
         let configs = glx.ChooseFBConfig(display as *mut _, screen_id, descriptor.as_ptr(),
                                         &mut num_configs);
-        println!("Num configs found: {}", num_configs);
         if configs.is_null() { return Err(()); }
         if num_configs == 0 { return Err(()); }
 
@@ -498,6 +497,7 @@ unsafe fn choose_fbconfig(glx: &ffi::glx::Glx, extensions: &str, xlib: &ffi::Xli
             let configs = slice::from_raw_parts(configs, num_configs as usize);
             configs.iter().find(|&config| {
                 let vi = glx.GetVisualFromFBConfig(display as *mut _, *config);
+                // Transparency was requested, so only choose configs with 32 bits for RGBA.
                 let found = !vi.is_null() && (*vi).depth == 32;
                 (xlib.XFree)(vi as *mut _);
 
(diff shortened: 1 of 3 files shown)


src/api/glx/mod.rs
```diff
@@ -487,13 +488,32 @@ unsafe fn choose_fbconfig(glx: &ffi::glx::Glx, extensions: &str, xlib: &ffi::Xli
     // calling glXChooseFBConfig
     let fb_config = {
         let mut num_configs = 1;
-        let result = glx.ChooseFBConfig(display as *mut _, screen_id, descriptor.as_ptr(),
+        let configs = glx.ChooseFBConfig(display as *mut _, screen_id, descriptor.as_ptr(),
                                         &mut num_configs);
-        if result.is_null() { return Err(()); }
+        println!("Num configs found: {}", num_configs);
```
mitchmindtree: Looks like you may have accidentally left a `println` here ^

src/api/glx/mod.rs
```diff
@@ -487,13 +488,32 @@ unsafe fn choose_fbconfig(glx: &ffi::glx::Glx, extensions: &str, xlib: &ffi::Xli
...
         if num_configs == 0 { return Err(()); }
-        let val = *result;
-        (xlib.XFree)(result as *mut _);
-        val
+
+        let config = if transparent {
+            let configs = slice::from_raw_parts(configs, num_configs as usize);
+            configs.iter().find(|&config| {
+                let vi = glx.GetVisualFromFBConfig(display as *mut _, *config);
+                let found = !vi.is_null() && (*vi).depth == 32;
```
mitchmindtree: Where does the `32` in this line come from? Would be  nice to have a comment for this

mitchmindtree:
Looks good! I left a couple inline comments for small things.

Here's a picture of what it looks like on my system (OS X 10.10.5):

screen shot 2017-03-05 at 11 44 35 pm

For some reason the title bar shows despite the call to `.with_decorations(false)` [here](https://github.com/mystal/glutin.git), however it seems this was also the case prior to your PR, so it's probably still fine to go ahead with your changes (unless you have any ideas? would be great to get it fixed!).

mystal:
Thanks for the review! I just pushed some cleanups to address them.

I'd like to fix the decorations as well, though I'm not sure what's needed yet. Will keep looking into it. I can send a separate PR for that since it might require changes in winit.

mitchmindtree:
Ok great, thanks for the changes! Looks good to me :+1:

PR #867, opened 2017-03-05
Better error handling

canndrew:
* `Display` impl for `CreationError` gives the details of the cause of the error.
 * `Window::new` returns `Err` instead of panicking if it can't get an X11 connection.


tomaka:
Thanks

PR #868, opened 2017-03-06
Handle android activity life cycle and event loop

MortimerGoro:
Handle android activity life cycle and event loop

Related PR: https://github.com/tomaka/winit/pull/153

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fix android activity life cycle
+
 # Version 0.13.0 (2018-02-28)
 
 - Update winit dependency to 0.11.1
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ winit = "0.11.1"
 gl_generator = "0.9"
 
 [target.'cfg(target_os = "android")'.dependencies.android_glue]
-version = "0.2"
+version = "0.2.3"
 
 [target.'cfg(target_os = "ios")'.dependencies]
 objc = "0.2"
--- a/src/api/android/mod.rs
+++ b/src/api/android/mod.rs
@@ -16,11 +16,34 @@ use PixelFormatRequirements;
 
 use api::egl;
 use api::egl::Context as EglContext;
+use std::cell::Cell;
+use std::sync::Arc;
+use winit::os::android::EventsLoopExt;
 
 mod ffi;
 
-pub struct Context {
+struct AndroidContext {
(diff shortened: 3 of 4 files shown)


MortimerGoro:
I did a rebase and adapted the code to the new EventLoop API

MortimerGoro:
@tomaka I see that the API has changed and EventLoop has been moved to winit. Now I can't create the EventLoop override to handle EGL life cycle events on Android:
 
- If I create a EventsLoop override in Glutin (like this PR did), it can't be used for the `glutin::GlWindow::new(window, gl_context, &events_loop).unwrap()`creation because the argument uses winit::EventLoop type
- If I create a EventsLoop override in Winit, I don't have access to the gl_context defined in winit.

We need the android life cycle merged to unfork glutin in Servo. What's the best way to do this now?

tomaka:
Err I knew there would be a quirk with exposing winit.

tomaka:
I thought a bit about this, and considered the following solutions:

- Letting the user handle the situation explicitly when they receive a `Suspend` event.
- Wrapping around the `EventsLoop` again.
- Try automatically detecting this in `make_current` and `is_current`.

But in the end I think the best thing to do is make a tweak to the winit's `EventsLoop` to make it possible for glutin to register a callback when a suspend event happens.
This is a bit of a hack, but I think that it doesn't have any downside.

The solution would be specific to Android (with a trait hidden with `#[cfg(target_os = "android")]`.
Do you think that's ok?

MortimerGoro:
Yes, I though doing a similar thing too but with all events. Doing with suspend event only it's ok. I don't think that we'll need to handle other events directly in Glutin.

paulrouget:
@tomaka Is what you describe here https://github.com/tomaka/glutin/pull/868#issuecomment-319442402 still the approach you recommend?

tomaka:
@paulrouget The discussion has more or less moved there: https://github.com/tomaka/winit/issues/231
There is still no good solution for this right now.

paulrouget:
@MortimerGoro winit 0.11.1 has been published.

MortimerGoro:
@paulrouget thanks for the notfication. PR updated!

@tomaka We need to update winit dependency from 10.x to 11.x here. Do you want a separate PR for that change?

tomaka:
@MortimerGoro Yes please.

MortimerGoro:
@tomaka r? PR ready

tomaka:
Thanks, looks good except for a few style issues.

PR #870, opened 2017-03-13
Remove ambiguous, unused `WindowAttributes` parameter

mitchmindtree:
I think this might have been left over since the pre-winit days. Rather
than being removed, it looks like `Default::default` was used instead.
This made working in the backends quite confusing as the
`WindowAttributes` are already available via the `winit::WindowBuilder`
arg and both contained different values.


mitchmindtree:
Closing as this was fixed in #901

PR #881, opened 2017-05-07
Bump version to expose WM_CLASS fix in winit

### LINKED ISSUE #879: WM_CLASS and WM_NAME not set before window is mapped
Many Linux window managers rely on being able to match against a new window's `WM_CLASS` or `WM_NAME` at launch to place them on the right workspace, set floating properties, etc. Since the window manager only looks at the value of these properties when the window is first mapped, ICCCM [requires](https://tronche.com/gui/x/icccm/sec-4.html#WM_CLASS) (for `WM_CLASS`) that
> This property must be present when the window leaves the Withdrawn state and may be changed only while the window is in the Withdrawn state. Window managers may examine the property only when they start up and when the window leaves the Withdrawn state, but there should be no need for a client to change its state dynamically.

While no such requirement exists for `WM_NAME`, setting it before the window is mapped is [advantageous](https://bbs.archlinux.org/viewtopic.php?pid=1581353#p1581353) for the same reason. Unfortunately, glutin currently does not appear to do this (neither does [Spotify](https://community.spotify.com/t5/Desktop-Linux-Windows-Web-Player/WM-CLASS-window-property-on-Linux/td-p/1247042) fwiw). It sets these properties *after* mapping the new window. While using `xprop` to inspect the window launched by `examples/window.rs` does show the right `WM_NAME` and `WM_CLASS` ("A fantastic window!"), this information is not visible to window managers. For example, neither of these two xmonad window hooks work:

```haskell
className =? "A fantastic window!" --> doShift "1"
```
```haskell
title =? "A fantastic window!" --> doShift "1"
```

On the other hand, matching against "" does move the window, suggesting that the window is initially mapped without title or class.

jonhoo:
winit 0.6.4 fixes https://github.com/tomaka/winit/pull/168, which in turn fixes #879.

Depends on https://github.com/tomaka/winit/pull/171.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "glutin"
-version = "0.8.0"
+version = "0.8.1"
 authors = ["The glutin contributors, Pierre Krieger <pierre.krieger1708@gmail.com>"]
 description = "Cross-platform OpenGL context provider."
 keywords = ["windowing", "opengl"]
@@ -14,7 +14,7 @@ build = "build.rs"
 lazy_static = "0.2.0"
 libc = "0.2"
 shared_library = "0.1.0"
-winit = "0.6.3"
+winit = "0.6.4"
 
 [build-dependencies]
 gl_generator = "0.5"


tomaka:
Closing this as there's no reason to do a PR in glutin. Running `cargo update` will fix the bug.

PR #883, opened 2017-05-14
Fix and update README example

LPGhatguy:
Changes include:
* Update to glutin 0.8.0 API
* Explicitly unwrap() to avoid compiler warnings
* Use builder API to specify dimensions, title, and v-sync
* Use traditional update loop, which avoids redrawing multiple times per sync when lots of events are happening


tomaka:
Thanks

PR #885, opened 2017-05-14
Add CHANGELOG.md

LPGhatguy:
The change log file I'm proposing is sparse and intended mostly to spark discussion.

I had some trouble porting some code utilizing Glutin to 0.8.0. The `README` example no longer compiled and there was no change log describing what exactly had changed. The commit history had a lot of noise and minor changes, when what I really wanted to see was something like this file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,6 @@
+# Change Log
+
+## 0.8.0 - April 28, 2017
+
+### Breaking
+* Updated to winit 0.6.0 `EventsLoop` API (#864)


trishume:
An alternative is just to add notes to the Github release tags. That's what a lot of other projects do.

jwilm:
In lieu of a CHANGELOG, you can try looking at the diff of some _examples_ to get a quick view of what's changed. Not a full solution, but at least you know the examples compile :)

LPGhatguy:
It would be nice if releases were tagged at all. Not just for tagging a version history, but to know what commit of the repository I'm actually using if I pull the crate from crates.io.

PR #888, opened 2017-05-18
Fixed android: added 'id' method

ozkriff:


ozkriff:
Could you publish 0.8.1?

PR #894, opened 2017-06-06
Enable option to not switch to dedicated GPU on mac also with an integrated GPU

tomfogg:
Adding this flag allows an app built with glutin to run without triggering a switch to the dedicated GPU on a mac (eg on a macbook with low power intel GPU and high power dedicated GPU).

This is mainly for [this issue on the alacritty project](https://github.com/jwilm/alacritty/issues/210)

From testing, the app also needs to set `NSSupportsAutomaticGraphicsSwitching` to `true` in the Info.plist in its app bundle to stop the GPU switch. That means this flag is probably OK to add by default since app behaviour won't change without also having that Info.plist setting.

There are some details about multiple GPU support in [this technical note](https://developer.apple.com/library/content/technotes/tn2229/_index.html)


tomaka:
As usual I'm not familiar enough with MacOS to have an opinion on this.
I'll merge in a few days if nobody is opposed to it.

tomaka:
Err I totally forgot about this PR.

tomaka:
Thanks for the insight @trishume

lilianmoraru:
Seems to fail the CI on macOS(Rust nightly): https://travis-ci.org/tomaka/glutin/builds/251043521

trishume:
@lilianmoraru the CI breakage isn't related to this PR, it seems to be the `cocoa` crate that broke.

Issue #880, opened 2017-05-05
Can't initialize with an existing winit::EventsLoop


ebkalderon:
It would be nice to be able to initialize glutin with a `winit::EventsLoop` and not just `glutin::EventsLoop`. If I am working with multiple winit-based graphics backends alongside glutin, it's annoying not being able to pass in the vanilla `winit::EventsLoop` into `glutin::WindowBuilder::build()`.

Perhaps we could add a `glutin::EventsLoop::from_winit_events_loop()` method just like the existing `glutin::WindowBuilder::from_winit_builder()`?

torkleyy:
Just took a look at it and it is really confusing how many `EventsLoop`s there are. Why are there platform wrappers in `glutin`? Isn't it possible to let `winit` do that?

mitchmindtree:
@torkleyy winit handles platform-specific window stuff, while glutin handles platform-specific OpenGL stuff.

ebkalderon:
@mitchmindtree If that's the case, would you mind explaining what platform-specific OpenGL stuff resides in `glutin::EventsLoop`? Looking at the code, I can't seem to find any.

mitchmindtree:
`glutin::EventsLoop` is just a wrapper around `glutin::platform::EventsLoop` - the actual platform that is used [depends on which `cfg` is satisfied](https://github.com/tomaka/glutin/blob/master/src/platform/mod.rs#L3-L20) at compile time.

As for the platform `EventsLoop`s themselves, [here's the macOS specific GL context](https://github.com/tomaka/glutin/blob/master/src/platform/macos/mod.rs#L60), [here's the wayland one](https://github.com/tomaka/glutin/blob/master/src/platform/linux/wayland.rs#L15), [the x11 one](https://github.com/tomaka/glutin/blob/master/src/platform/linux/x11.rs#L47), etc.

The reason the distinction exists is to allow for the window-specific stuff (winit) to be used with non-GL graphics contexts (like Vulkan).

mitchmindtree:
All that said, I agree with OPs suggestion that it would be nice if we had a method for building a `glutin::EventsLoop` from a `winit::EventsLoop`.

ebkalderon:
@mitchmindtree Thanks for the detailed explanation! Indeed, I think it would be nice to be able to initialize from a `winit::EventsLoop`. If no one is willing to work on it, would you mind if I made a PR?

kvark:
[...]

PR #897, opened 2017-06-14
Redesign - Remove winit `Window` and `EventsLoop` wrappers. Add `Context` type.

mitchmindtree:
Ever since the windowing part of glutin was separated into its own crate ([winit](https://github.com/tomaka/winit)), much of glutin's API has felt clunky and unnecessary. In particular, the wrapping around winit's `Window`, `WindowBuilder` and `EventsLoop` can be frustrating to maintain, requiring wrapper methods around *all* methods on each of those types, even if `glutin` itself doesn't do anything interesting with those methods. This means that any time winit's API is tweaked in the slightest, glutin normally also requires updating.

After giving it some thought, I've been thinking that `glutin` might benefit from a redesign that removed all `Window` and `EventsLoop` related code in favour of a simpler API purely aimed at setting up and using an OpenGL context. This PR is currently a draft of that change, mainly to get some feedback and thoughts from other glutin/glium users.

So far the API and linux backends have been updated, along with the window example to demonstrate what usage might look like.

I haven't looked closesly at how these changes will work with glium yet, so there might still be some more API problems to solve. I'll take a closer look at this soon.

@tomaka a while back you mentioned that glutin could use a redesign. If you get a chance, it would be great to get your thoughts on this and what you originally had in mind.


kvark:
I was always in favour of less wrapping and more re-exports for `glutin`. Thanks @mitchmindtree for your awesome work!

mitchmindtree:
OK, I believe this redesign is complete and ready for feedback, review and testing. I've been able to test the changes successfully on linux+gnome+wayland and macOS but have not been able to test the running examples on windows as my VM for win10 doesn't have GL support. If a windows user could test this, that would be really helpful.

ebkalderon:
@mitchmindtree Just in time! I was just about to submit a pull request to fix #880. This is even better, thank you.

mitchmindtree:
Thanks for the support btw @kvark! Is there any chance you'd be interested in reviewing this? Would love to move forward on these changes but I'd like to get some reviews to give @tomaka some peace of mind about merging. cc @ebkalderon, would you be interested in reviewing also?

jwilm:
@mitchmindtree I started looking at this, but it looks like it needs to be updated for the latest winit changes (mainly the examples).

mitchmindtree:
@jwilm the PR is against the latest published version of winit (0.6.4). I've submitted tomaka/winit#205 to publish winit 0.7 - if that lands first, I'll update this PR to require 0.7.

jwilm:
@mitchmindtree sounds good! Seems like it would make sense to get that published before landing this rather than having additional API churn.

I'm certainly looking forward for all the dust to settle here and get Alacritty back on the published Glutin and winit. :smile:

mitchmindtree:
Okydoke, the PR now also updates to the recently published winit **0.7.0**!

mitchmindtree:
I just realised the `ContextBuilder` was missing a `.with_shared_lists` method and had no way of setting the `sharing` field, so I just added the method.

jwilm:
Thanks for getting it updated @mitchmindtree!

I attempted to integrate this latest version [into a branch](https://github.com/jwilm/alacritty/compare/glutin-winit-take-two) of Alacritty, and I'm running into different bugs on both macOS and Linux (X11) . On macOS, it seems that there's an issue with window dimensions/transform/offset that was not present in the previous version. From what I can tell, I didn't make any changes that should have caused that. On X11 with i3, the window only stays open for a brief moment before closing with an error about not being able to find the root window. Again, it doesn't seem like something that would be caused by changes in application code.

<details><summary>Backtrace of X11 issue; fails during context creation</summary>
<pre>
jwilm@jwilm-desk ➜  ~/code/alacritty  RUST_BACKTRACE=1 cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/alacritty`
thread 'main' panicked at 'Failed to get root window: XError { description: "BadMatch (invalid parameter attributes)", error_code: 8, request_code: 154, minor_code: 5 }', /checkout/src/libcore/result.rs:859
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:60
             at /checkout/src/libstd/panicking.rs:355
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:371
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:549
   5: std::panicking::begin_panic
             at /checkout/src/libstd/panicking.rs:511
   6: std::panicking::begin_panic_fmt
             at /checkout/src/libstd/panicking.rs:495
   7: rust_begin_unwind
             at /checkout/src/libstd/panicking.rs:471
   8: core::panicking::panic_fmt
             at /checkout/src/libcore/panicking.rs:69
   9: core::result::unwrap_failed
             at /checkout/src/libcore/macros.rs:29
  10: <core::result::Result<T, E>>::expect
             at /checkout/src/libcore/result.rs:761
  11: glutin::platform::platform::x11::Context::new
             at /home/jwilm/.cargo/git/checkouts/glutin-02ca2afa52543597/2181730/src/platform/linux/x11.rs:224
  12: glutin::platform::platform::Context::new
             at /home/jwilm/.cargo/git/checkouts/glutin-02ca2afa52543597/2181730/src/platform/linux/mod.rs:47
  13: glutin::Context::new
             at /home/jwilm/.cargo/git/checkouts/glutin-02ca2afa52543597/2181730/src/lib.rs:254
  14: glutin::ContextBuilder::build
             at /home/jwilm/.cargo/git/checkouts/glutin-02ca2afa52543597/2181730/src/lib.rs:240
  15: alacritty::window::Window::new
             at src/window.rs:205
  16: alacritty::display::Display::new
             at src/display.rs:140
  17: alacritty::run
             at src/main.rs:92
  18: alacritty::main
             at src/main.rs:45
  19: __rust_maybe_catch_panic
             at /checkout/src/libpanic_unwind/lib.rs:98
  20: std::rt::lang_start
             at /checkout/src/libstd/panicking.rs:433
             at /checkout/src/libstd/panic.rs:361
             at /checkout/src/libstd/rt.rs:57
  21: main
  22: __libc_start_main
  23: _start
</pre>
</details>

jwilm:
Running the window example on this branch fails as well.

<details><summary><code>cargo run --example window</code> output</summary>
<pre>
jwilm@jwilm-desk ➜  ~/code/mitchmindtree/glutin  RUST_BACKTRACE=1 cargo run --example window
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/examples/window`
thread 'main' panicked at 'glx::MakeCurrent failed', src/api/glx/mod.rs:112
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:60
             at /checkout/src/libstd/panicking.rs:355
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:371
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:549
   5: std::panicking::begin_panic
             at /checkout/src/libstd/panicking.rs:511
   6: glutin::api::glx::Context::make_current
             at ./src/api/glx/mod.rs:112
   7: glutin::platform::platform::x11::Context::make_current
             at ./src/platform/linux/x11.rs:245
   8: glutin::platform::platform::Context::make_current
             at ./src/platform/linux/mod.rs:61
   9: glutin::Context::make_current
             at ./src/lib.rs:271
  10: window::main
             at ./examples/window.rs:15
  11: std::panicking::try::do_call
             at /checkout/src/libstd/panicking.rs:454
  12: __rust_maybe_catch_panic
             at /checkout/src/libpanic_unwind/lib.rs:98
  13: std::rt::lang_start
             at /checkout/src/libstd/panicking.rs:433
             at /checkout/src/libstd/panic.rs:361
             at /checkout/src/libstd/rt.rs:57
  14: main
  15: __libc_start_main
  16: _start
</pre>
</details>

jwilm:
The window example runs on macOS, and the triangle is right where it should be. I'll double check my code again regarding that issue.

mitchmindtree:
@jwilm Thanks for taking the time to have a look!

> On macOS, it seems that there's an issue with window dimensions/transform/offset that was not present in the previous version. From what I can tell, I didn't make any changes that should have caused that.

I had a look through your changes and it looks like you might need to be calling `self.window.context.resize(w, h)` somewhere around [here](https://github.com/jwilm/alacritty/blob/b292ef798f25e5e9059afbaddcab03f61db87f6a/src/display.rs#L253)? I did a quick search and I can't seem to find it being called anywhere else.

Previously as glutin wrapped the winit EventsLoop, glutin would automatically do the equivalent of this when the EventsLoop received `Resize` events behind the scene. If you take a look at the examples you'll notice that after this redesign, each of them now call `resize` manuallly on the `Context` when a resize event is received. This is basically all that the `glutin::EventsLoop` used to do that the `winit::EventsLoop` did not already. This might also address some of the x11 weirdness that you were experiencing, as I believe the context does not get updated at all without this call (Resized always seems to be the first event emitted on X11 too).

> On X11 with i3, the window only stays open for a brief moment before closing with an error about not being able to find the root window.

The window example seems to work fine for me on this branch using Arch+Gnome3+X11

screenshot from 2017-06-25 11-43-24

I wonder if this is also related to the previous fix I mentioned? Can you also confirm that the i3+X11 issues did not already occur before switching to this branch?

> Running the window example on this branch fails as well. (x11)

This seems really strange as it works fine for me. Are you sure there are no local modifications to the branch? Did the example work on your system prior to switching to this branch?

jwilm:
@mitchmindtree thanks for the in-depth follow-up!

> I had a look through your changes and it looks like you might need to be calling self.window.context.resize(w, h) somewhere around here?

This was *exactly* the problem. Everything is now working as expected on that platform. This is going to be an important migration note when the patch is published.

Regarding X11, I've done some more testing.

* winit master branch: **works**
* Glutin master branch: **works**
* Glutin redesign branch: **glx::MakeCurrent failed**

I'm about to head to bed, but I'll try and make time to dig into that tomorrow. Seems easier for me to investigate since you are unable to repro.

mitchmindtree:
Just thought I'd mention that I went ahead and made [a new branch](https://github.com/glium/glium/compare/master...mitchmindtree:glutin_redesign?expand=1) from [the glium "winit-api-update" PR](https://github.com/glium/glium/pull/1580) that updates to this redesign to make sure that it does not introduce any incompatible changes downstream. All glium examples and tests that previously ran successfully continued to do so on X11. You can try the branch for yourself [here](https://github.com/mitchmindtree/glium/tree/glutin_redesign).

jwilm:
I haven't had much time to dig into this today, but I did have an opportunity to test the glium examples. They fail with the `glx::MakeCurrent` panic as well.

<details>
<summary>Backtrace</summary>
<pre>
     Running `target/debug/examples/teapot`
thread 'main' panicked at 'glx::MakeCurrent failed', /home/jwilm/.cargo/git/checkouts/glutin-02ca2afa52543597/2181730/src/api/glx/mod.rs:112
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:60
             at /checkout/src/libstd/panicking.rs:355
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:371
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:549
   5: std::panicking::begin_panic
             at /checkout/src/libstd/panicking.rs:511
   6: glutin::api::glx::Context::make_current
             at /home/jwilm/.cargo/git/checkouts/glutin-02ca2afa52543597/2181730/src/api/glx/mod.rs:112
   7: glutin::platform::platform::x11::Context::make_current
             at /home/jwilm/.cargo/git/checkouts/glutin-02ca2afa52543597/2181730/src/platform/linux/x11.rs:245
   8: glutin::platform::platform::Context::make_current
             at /home/jwilm/.cargo/git/checkouts/glutin-02ca2afa52543597/2181730/src/platform/linux/mod.rs:61
   9: glutin::Context::make_current
             at /home/jwilm/.cargo/git/checkouts/glutin-02ca2afa52543597/2181730/src/lib.rs:271
  10: <glium::backend::glutin::GlutinBackend as glium::backend::Backend>::make_current
             at ./src/backend/glutin/mod.rs:244
  11: glium::context::Context::new
             at ./src/context/mod.rs:162
  12: glium::backend::glutin::Display::new_inner
             at ./src/backend/glutin/mod.rs:111
  13: glium::backend::glutin::Display::with_debug
             at ./src/backend/glutin/mod.rs:86
  14: glium::backend::glutin::Display::new
             at ./src/backend/glutin/mod.rs:64
  15: teapot::main
             at ./examples/teapot.rs:19
  16: std::panicking::try::do_call
             at /checkout/src/libstd/panicking.rs:454
  17: __rust_maybe_catch_panic
             at /checkout/src/libpanic_unwind/lib.rs:98
  18: std::rt::lang_start
             at /checkout/src/libstd/panicking.rs:433
             at /checkout/src/libstd/panic.rs:361
             at /checkout/src/libstd/rt.rs:57
  19: main
  20: __libc_start_main
  21: _start
</pre>
</details>

mitchmindtree:
@jwilm I just added a commit - it *might* have fixed the issue you were running into but I'm not entirely sure it is related. Would you mind giving it a go?

I came across [this issue](https://github.com/tomaka/glutin/issues/834) which seemed related to the panic you're running into. It mentioned some issues wrt to the screen_id, and I remembered that I had an issue with updating the screen_id code to the redesign. Ralith had [mentioned a solution](https://github.com/tomaka/winit/issues/197#issuecomment-310581878) but I had forgotten to add it until now. I'll be curious to hear if this fixes the issue.

Thanks again for reviewing this!

jwilm:
Thanks for the patch! That didn't fix the primary issue here, but it did resolve an issue where the window was always opening in a particular workspace.

Regarding the bug at hand, I've been looking into it this morning. Not sure how much more time I'll have right now so I'll share what I learned.

It looks like `glXMakeCurrent` is generating this error:

```
BadMatch (invalid parameter attributes) (code: 8, request code: 154, minor code: 5)
```

Referencing the [`glXMakeCurrent` docs](ftp://www.x.org/pub/X11R6.8.0/doc/glXMakeCurrent.3.html) (btw, is there any way to get these as man pages?), we see this reason:

> **BadMatch** is generated if `drawable` was not created with the same X screen and visual as `ctx`. It is also generated if `drawable` is **None** and `ctx` is not **NULL**.

Since we have a `drawable` and ` ctx`, it seems like there must be a mismatch of X screen and visuals between them.

jwilm:
Another development:

In Alacritty, I create the window and context like this:

```rust
        let window = WindowBuilder::new()
            .with_title(title)
            .build(&event_loop)?;
        let context = ContextBuilder::new()
            .with_vsync()
            .build(&window)?;
```

And this causes glutin to panic while fetching the root window with the same **BadMatch** error (trace is in an earlier comment). If I comment out `with_vsync()` on the context builder, the panic moves to the same place as all of the examples which is at `glx::make_current()`.

jwilm:
This was removed in the latest branch:

```rust
        let winit_window = winit_builder
            .with_x11_visual(&visual_infos as *const _)
            .with_x11_screen(screen_id)
            .build(events_loop).unwrap();
```

It looks like, previously, the context was created before the window, and then the XVisualInfo and screen id from the context were used to actually create the window. It doesn't appear there was any code added to replace this behavior. This seems to support the error code above about a visuals mismatch or screen mismatch.

I've run out of time for the morning. Additionally, I'm kind of reaching the edge of my knowledge about X APIs and glX. If you have any suggestions on how to proceed at this point, they would be most welcome.

tomaka:
I'm really dubious about the fact that the context builder can be made safe.
For example if the user can in some way make the library crash by creating an OpenGL when a DirectX already exists and was created in a safe way as well with another library, then it's going to need to be marked as `unsafe`.

mitchmindtree:
> For example if the user can in some way make the library crash by creating an OpenGL when a DirectX already exists and was created in a safe way with another library, then it's going to need to be marked as unsafe.

@tomaka Ahh I see. Is it possible to check if a DirectX context has already been created within the windows backend to avoid this? I wonder if the OS provides some way of returning whether or not it is safe to create a GL context, or if it returns some error so that we can `panic!` or return an `Err` instead of leading to unsafety. 

> This was removed in the latest branch

@jwilm thanks for finding this, I must have removed it and forgotten to come back to it. I'll have a think about how to address this and come back to it.

jwilm:
@mitchmindtree had one other thought -

[glXIntro](https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glXIntro.xml) shows the X Window being created using the exact same FBConfig as the GLXContext. Maybe stashing a copy of the FBConfig used during window creation so that it may be used again during context creation would work?

mitchmindtree:
@tomaka surely the [`wglCreateContext`](https://github.com/tomaka/glutin/blob/master/src/api/wgl/mod.rs#L333) and [`wglCreateContextAttribsARB`](https://github.com/tomaka/glutin/blob/master/src/api/wgl/mod.rs#L317) function calls would fail (return `NULL`) if a DirectX context already exists? [The documentation](
https://msdn.microsoft.com/en-us/library/windows/desktop/dd374379(v=vs.85).aspx) mentions that it should return `NULL` upon failure, however I cannot find anything specific on the problem that you mention.

I would test this however I don't have access to windows with openGL (my VM instance of windows 10 does not support OpenGL).

src/lib.rs
```diff
@@ -56,130 +57,251 @@ extern crate cocoa;
...
+    }
+
+    /// Request the backend to be stereoscopic.
+    #[inline]
+    pub fn with_stereoscopy(mut self) -> Self {
+        self.pf_reqs.stereoscopy = true;
+        self
+    }
+
+    /// Sets whether sRGB should be enabled on the window. `None` means "I don't care".
```
kvark: I don't understand the use case of "I don't care", to be honest
Although it's a matter of a separate discussion... just thought it might be a good time to revisit this considering the breaking changes coming anyway.

src/lib.rs
```diff
@@ -56,130 +57,251 @@ extern crate cocoa;
...
+    /// Sets the robustness of the OpenGL context. See the docs of `Robustness`.
+    #[inline]
+    pub fn with_gl_robustness(mut self, robustness: Robustness) -> Self {
+        self.gl_attr.robustness = robustness;
+        self
+    }
+
+    /// Requests that the window has vsync enabled.
+    #[inline]
+    pub fn with_vsync(mut self) -> Self {
```
kvark: contrary to all other options, this one doesn't allow you to disable vsync, which seems inconsistent

src/platform/linux/mod.rs
```diff
@@ -1,22 +1,109 @@
...
 mod wayland;
 mod x11;
 
+pub enum Context {
+    X(x11::Context),
+    Wayland(wayland::Context)
+}
+
+impl Context {
+
```
kvark: nit: quite a few new lines that I don't think need to be here?

src/platform/linux/mod.rs
```diff
@@ -1,22 +1,109 @@
...
+    #[inline]
+    pub fn new(
+        window: &winit::Window,
+        pf_reqs: &PixelFormatRequirements,
+        gl_attr: &GlAttributes<&Context>,
+    ) -> Result<Self, CreationError>
+    {
+
+        if wayland_client::default_connect().is_ok() {
```
kvark: nit: it would look better with a match

src/platform/linux/wayland.rs
```diff
@@ -1,158 +1,80 @@
...
-                    .and_then(|p| p.finish(egl_surface.ptr() as *const _))
-                )
-            };
-            Window {
-                egl_surface: Arc::new(egl_surface),
-                context: context,
+        gl_attr: &GlAttributes<&Context>,
+    ) -> Result<Self, CreationError>
+    {
+        let (w, h) = window.get_inner_size().unwrap();
```
kvark: `get_inner_size` is deprecated

src/platform/linux/wayland.rs
```diff
@@ -1,158 +1,80 @@
...
+        let context = {
+            let libegl = unsafe { dlopen::dlopen(b"libEGL.so\0".as_ptr() as *const _, dlopen::RTLD_NOW) };
+            if libegl.is_null() {
+                return Err(CreationError::NotSupported);
             }
+            let egl = ::api::egl::ffi::egl::Egl::load_with(|sym| {
+                let sym = CString::new(sym).unwrap();
+                unsafe { dlopen::dlsym(libegl, sym.as_ptr()) }
+            });
+            let gl_attr = gl_attr.clone().map_sharing(|_| unimplemented!()); // TODO
```
kvark: TODO needs to be addressed before merging?

src/lib.rs
mitchmindtree: Hmmm it looks like this immediately defaults to `false` in the builder method if `None` was given. I'll change this to take a `bool` and add a note about defaulting to `false`.

src/platform/linux/wayland.rs
mitchmindtree: This `unimplemented!()` existed before this PR (and also exists in most of the other platforms), so I don't think it should necessary to address it here. I've added a tracking issue [here](https://github.com/tomaka/glutin/issues/899).

src/api/android/mod.rs
```diff
@@ -11,219 +11,121 @@ use winit;
...
 
     #[inline]
-    fn get_pixel_format(&self) -> PixelFormat {
-        self.context.get_pixel_format()
+    pub fn get_pixel_format(&self) -> PixelFormat {
+        self.egl_context.get_pixel_format()
     }
 }
 
 pub struct HeadlessContext(EglContext);
```
mitchmindtree: Ah thanks for pointing this out! I should have added a comment about this in the top-level comment.

I decided to remove the `GlContext` trait for a few reasons:

1. It was not serving any purpose for genericity within the crate itself.
2. Previously the role of these methods would have been less clear if implemented directly on the Window itself, however that is no longer the case as the type itself indicates the GL `Context` nature.
3. By implementing the methods directly on the types, users no longer need to import the `GlContext` trait to call them.
4. If users of the crate need to remain generic over `Context` and `HeadlessContext`, it is trivial for them to write a trait that does this. Personally, I don't think this use case is common enough to be worth the indirection of requiring another similarly named trait, but I could be wrong on this.

@kvark what are your thoughts on this? I guess I'd be OK with re-adding the `GlContext` trait if there was enough desire for it, as long as we could also leave the inherent method implementations on the types themselves as well so that users who don't need to remain generic over the two types don't need to know about the trait.

mitchmindtree:
@kvark thanks for reviewing! I've addressed each of your notes and left a couple of comments.

jwilm:
@mitchmindtree any thoughts on how to resolve the code removal on x11?

mitchmindtree:
@jwilm I'm just having a look at this now :)

To summarise the issue - before this PR glutin's x11 backend would build the window after calling the following two methods:

```rust
    .with_x11_visual(&visual_infos as *const _)
    .with_x11_screen(screen_id)
```

I believe the `.with_x11_screen` method has already been accounted for, as we now retrieve the `screen_id` from the pre-built `Window` rather than the `WindowBuilder` (see [here](https://github.com/tomaka/glutin/pull/897/files#diff-e24e3ccc2abef223529e2f0fbdba35d3R127) in the diff).

This leaves the `.with_x11_visual` call, which passes a pointer to a `XVisualInfos` struct used within the x11 winit window implementation. Specifically, the struct seems to be used for:

1. Creating the `colormap` field on the XSetWindowAttributes struct [here](https://github.com/tomaka/winit/blob/master/src/platform/linux/x11/window.rs#L174) (if no `XVisualInfos` was provided, `0` is used).
2. Calling `XCreateWindow`, providing the [depth](https://github.com/tomaka/winit/blob/master/src/platform/linux/x11/window.rs#L202) and [visual](https://github.com/tomaka/winit/blob/master/src/platform/linux/x11/window.rs#L207) (if no `XVisualInfos` was provided, these values are copied from the parent).

In glutin itself, the `visual_infos` only seems to be directly used [to create a color map](https://github.com/mitchmindtree/glutin/blob/a6e0d99758a587775b1005c6d99095c0cddf6026/src/platform/linux/x11.rs#L223). The created colormap seems to go entirely unused until `XFreeColormap` is called when the `Context` is dropped. (Side note: I wonder if we can get rid of this colormap? Looks like it might have been left over after the winit transition).

[This SO answer](https://stackoverflow.com/a/2254063/1711917) mentions that we should be able to create the GL context from the the window's existing `Visual` using `XGetWindowAttributes`. @jwilm do you think this would be acceptable? Would this behave differently to your suggestion of storing the window's fbconfig?

mitchmindtree:
@jwilm I've added a commit with the changes I mentioned in my previous comment - would you mind having a go and let me know if you're still running into that same error? (The travis error is just a macos-specific error that appeared in nightly after some breaking change).

One problem I noticed while testing them was that on `master` branch the transparent.rs example works perfectly on X11, however on this PR branch the background is no longer transparent and is black instead. I'm guessing this has something to do with the fact that the window is now responsible for providing the `Visual` to the glx context rather than the other way around. Originally the glx was responsible for producing the `XVisualInfo` via `glxGetVisualFromFBConfig`, however after this PR the visual is the default `Visual` for an `XWindow`, which I believe is `CopyFromParent`.

@jwilm one possible fix might be to use the FBConfig to get the visual during window creation and then stash the fbconfig somehow as you mentioned in your previous comment? That said, I haven't looked into what the FBConfig actually is yet. If it is an opengl-specific thing then this is probably a no-go as winit is designed for use with other graphics contexts in mind.

mitchmindtree:
One possible alternative to the redesign as it currently stands would be to have the `ContextBuilder::build` argument take a `winit::WindowBuilder` as an argument rather than a `winit::Window`, and return the built `winit::Window` alongside the `Context`. E.g.

```rust
    let mut events_loop = glutin::winit::EventsLoop::new();
    let window_builder = glutin::winit::WindowBuilder::new()
        .with_title("A fantastic window!");
    let (window, context) = glutin::ContextBuilder::new()
        .build(window_builder, &events_loop)
        .unwrap();
```

Although this would be slightly less flexible, it would still largely reduce the maintenance burden and would solve both the X11 issue I mention in my previous comment and @tomaka's issue wrt possible safety issues when colliding with an existing graphics context.

mitchmindtree:
Going to close this in favour of #901 as it solves some issues that block this PR and generally seems to be the preferred direction.

PR #901, opened 2017-07-05
Redesign V2 - Build the `Context` with `WindowBuilder` rather than `&Window`

mitchmindtree:
This is a follow up to the proposal at #897. More specifically, it is an implementation of the idea proposed in [this comment](https://github.com/tomaka/glutin/pull/897#issuecomment-312989699).

It is almost entirely the same as #897, but changes the `ContextBuilder` API slightly to address two primary issues:

1. X11 seems to require creating the GL context prior to creating the window in order to behave correctly (see [this comment](https://github.com/tomaka/glutin/pull/897#issuecomment-312656186)).
2. A user may attempt to create a GL context for a window with some other existing graphics context (e.g. DirectX) resulting in possibly unsafe behaviour.

Rather than building the `Context` with a borrowed, pre-built `winit::Window`, this PR proposes building the `Context` with a `winit::WindowBuilder`. This gives glutin control over the order of window and context creation solving **1**, while also ensuring that the `Window` cannot already be associated with some other graphics context which solves **2**.

Although slightly less flexible than #897, we still get the primary benefit of no longer having to maintain wrappers around the `Window` and `EventsLoop` types. This simplifies glutin's role to providing a cross-platform GL `Context` that is compatible with winit and can work nicely *alongside* it.

Here's an example of using glutin after this redesign:

```rust
let mut events_loop = glutin::winit::EventsLoop::new();
let window_builder = glutin::winit::WindowBuilder::new()
    .with_title("A fantastic window!");
let (window, context) = glutin::ContextBuilder::new()
    .build(window_builder, &events_loop)
    .unwrap();
```

See the updated examples for more detailed usage.

I'm going to leave #897 open for now in case it turns out to be possible to solve the X11 creation order problem some other way and in case we decide the existing graphics context issue should not be our problem.

cc @tomaka, @jwilm, @LPGhatguy, @kvark, @ebkalderon, @brendanzab


LPGhatguy:
I like that this pull request removes more code than it adds.

It address the issue I run into with trying to test out winit patches using a project I'm working on that uses glutin.

tomaka:
I much prefer this one.

brendanzab:
> Although slightly less flexible

What flexibility is reduced compared to the other PR? Could you also provide the previously proposed API alongside the V2 one?

mitchmindtree:
> What flexibility is reduced compared to the other PR?

By reduced flexibility, I mean that a `Context` can not be built from an existing `winit::Window` and instead must control the window building stage.

> Could you also provide the previously proposed API alongside the V2 one?

Sure! This is the example from my original comment demonstrating **V2**:

```rust
let mut events_loop = glutin::winit::EventsLoop::new();
let window_builder = glutin::winit::WindowBuilder::new()
    .with_title("A fantastic window!");
let (window, context) = glutin::ContextBuilder::new()
    .build(window_builder, &events_loop)
    .unwrap();
```

And this is the equivalent using the **V1** #897 API (and in turn is also subject to the two problems I mentioned in the original comment):

```rust
let mut events_loop = glutin::winit::EventsLoop::new();
let window = glutin::winit::WindowBuilder::new()
    .with_title("A fantastic window!")
    .build(&events_loop)
    .unwrap();
let context = glutin::ContextBuilder::new()
    .build(&window)
    .unwrap();
```

jwilm:
This looks great! I'm away from my Linux machine until Sunday; I will give it a try once I'm back.

brendanzab:
Yeah, awesome work @mitchmindtree in figuring out how to satisfy such a difficult design problem! I'm cool with this approach! 👍

Cargo.toml
```diff
@@ -14,7 +14,7 @@ build = "build.rs"
 lazy_static = "0.2.0"
 libc = "0.2"
 shared_library = "0.1.0"
-winit = "0.6.3"
+winit = "0.7.0"
```
willglynn: Does this specifically need winit 0.7.0? [winit 0.7.2](https://crates.io/crates/winit) is now out.

mitchmindtree:
Thanks @burtonageo, I addressed each of your comments!

I also just updated the README and root-level documentation. I think this should be ready to go now as long as everyone is happy.

tomaka:
My review now:

- The window and the context should really be in only one struct. If someone destroys the window before the context, everything explodes. That's hugely unsafe.
- The window builder, events loop, etc. from winit should be pub-reexported in lib.rs. I repeat this often but for me it is totally wrong to require the user to depend on winit.
- I don't understand the motivation for removing the `GlContext` trait.

mitchmindtree:
> The window and the context should really be in only one struct. 

How would you feel about something like a `Facade` or `Display` type for this (I'm open to a better name if someone has one)? Something simple like:

```rust
pub struct Facade {
    context: Context,
    window: winit::Window,
}

impl Facade {
    pub fn context(&self) -> &Context {
        &self.context
    }
    pub fn window(&self) -> &winit::Window {
        &self.window
    }
}
```

I guess `Deref` could be `impl`d for the type, but it's not really clear to me which type it should deref to, so I'd probably prefer borrowing the context or window directly, e.g. `facade.context().swap_buffers()` or `facade.window().get_inner_size_points()`. I'd prefer not to add wrapper methods for all the methods on both types as this would just recreate the same maintenance burden that this redesign is trying to avoid.

> The window builder, events loop, etc. from winit should be pub-reexported in lib.rs. I repeat this often but for me it is totally wrong to require the user to depend on winit.

The user is depending on winit either way when using glutin, but I can understand if you'd prefer the user did not need to know about winit. I'm happy to make this change.

> I don't understand the motivation for removing the GlContext trait.

I explained my motivations for removing the GlContex trait [here](https://github.com/tomaka/glutin/pull/897#discussion_r124712154) - what don't you understand about this?

tomaka:
> I explained my motivations for removing the GlContex trait here - what don't you understand about this?

For me the only valid reason for removing it is the third, and it's pretty minor.

> How would you feel about something like a Facade or Display type for this (I'm open to a better name if someone has one)? Something simple like:

Alternatively, make it `Deref` to `winit::Window` but also implement `GlContext` on it as well?

mitchmindtree:
@tomaka I just added some commits that:

- Re-exports the public winit API from lib.rs so that the user doesn't need to know about winit.
- Adds a `GlWindow` type which combines the `Window` and `Context` (I guess we should look into controlling the order of dropping and how this should be handled for each OS?).
- Re-adds the `GlContext` and provides implementations for `Context`, `HeadlessContext` and `GlWindow`.

tomaka:
> I guess we should look into controlling the order of dropping

The drop order has been stabilized recently in Rust, so it's not a problem.

tomaka:
Looks good to me

mitchmindtree:
@tomaka I just bumped the glutin version to 0.9 ready for publishing - just let me know if you want me to change it back.

jwilm:
Have had a chance to test this branch now; everything works as expected. Great work @mitchmindtree! :clap:

PR #905, opened 2017-07-11
Add support for winit's choice of XWayland

alexheretic:
With winit#199 XWayland is usable with env var `WINIT_UNIX_BACKEND`. Glutin can no longer assume wayland is winit's preferred backend whenever available. 
By checking `winit::os::unix::get_x11_xconnection()` we can tell if winit has a preference for X.

This allows `WINIT_UNIX_BACKEND=x11 cargo run --example window` to run the example under XWayland, instead of panic.


alexheretic:
@vberger if you have time maybe you could glance over this?

vberger:
Well, this looks good, and should likely have been the proper way of doing it from the start. 👍

PR #907, opened 2017-07-14
Fix context creation on windows

msiglreith:
As proposed in https://github.com/tomaka/glutin/issues/906#issuecomment-315436660:

The windows backend creates a new dummy window for loading context specific functions.
This dummy window reuses the existing class of the real window, but this interferes with thread-local storage, which are not initialized on the main thread.
Registering a new dummy specific class with the default callback for event handling resolved this issue.

Fixes #906
cc @tomaka

PR #908, opened 2017-07-15
Fix multiple registrations for dummy window class

msiglreith:
Follow-up to #907 

Creating multiple windows with associated GL contexts will cause the dummy window class to be registered multiple times (https://github.com/tomaka/glutin/pull/907#issuecomment-315532372). Multi registrations will emit an error which we want to silently ignore.

Following the approach from https://github.com/tomaka/winit/blob/master/src/platform/windows/window.rs#L456-L460.


ConorLPBoyle:
Multiwindow is now fixed. Now to see if gfx-rs triangle will run...

PR #936, opened 2017-09-29
Publish 0.9.3

bogdad:
Hi!

Not very sure how the publish process works,
I need this fix published https://github.com/tomaka/glutin/pull/925

Is it a good time to publish? :)

Thank you,
Vladimir

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "glutin"
-version = "0.9.2"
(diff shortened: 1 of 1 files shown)


tomaka:
There was a breaking change unfortunately, so we shouldn't publish master as 0.9.3 but as 0.10.0 instead.

bogdad:
fixed!

PR #939, opened 2017-10-03
Implement access of raw context handles

### LINKED ISSUE #935: Getting raw pointer to platform GL context
I am writing some type-safe Rust bindings to [RenderDoc](https://renderdoc.org), a fairly popular graphics API debugger, and I am trying to get it to play nice with glutin-based applications.

When RenderDoc gets injected into the running application, the target application calls the [SetActiveWindow](https://renderdoc.org/docs/in_application_api.html#_CPPv215SetActiveWindow23RENDERDOC_DevicePointer22RENDERDOC_WindowHandle) function to tell RenderDoc which window to display the debug overlay and which graphics device to capture command lists from. To paraphrase the documentation:

```c
void SetActiveWindow(RENDERDOC_DevicePointer device, RENDERDOC_WindowHandle wndHandle)
```
> #### Note: 
> `RENDERDOC_DevicePointer` is a typedef to `void *`. The contents of it are API specific:
>
> * For D3D11 it must be the `ID3D11Device` device object.
> * **For OpenGL it must be the `HGLRC` or `GLXContext` context object.**
> * For Vulkan it must be the dispatch table pointer within the `VkInstance`. This is a pointer-sized value at the location pointed to by the `VkInstance`. NOTE - this is not the actual `VkInstance` pointer itself.
>
> `RENDERDOC_WindowHandle` is a typedef to `void *`. It is the platform specific `HWND`, `xcb_window_t`, or Xlib `Window`.

`glutin::GlWindow` can give me the raw window handle via the (albeit deprecated) `platform_window()` method, but I can't seem to acquire a raw pointer to the underlying GL context. Is there any way to do this with glutin at present? If not, how realistically could a feature like this be added? I am willing to make a pull request if needed.

ebkalderon:
### Added

* Add `GlContextExt` trait, which grants type-safe access to the underlying OpenGL context (fixes #935).

### Fixed

* Minor formatting changes, i.e. fixing some spacing, reordering imports, etc.
* Fix small doc comment typo.

CC @tomaka @kvark

### DIFF
--- a/src/api/osmesa/mod.rs
+++ b/src/api/osmesa/mod.rs
@@ -192,7 +192,7 @@ impl OsMesaContext {
     }
 
     #[inline]
-    pub unsafe fn as_mut_ptr(&self) -> osmesa_sys::OSMesaContext {
+    pub unsafe fn raw_handle(&self) -> osmesa_sys::OSMesaContext {
         self.context
     }
 }
--- a/src/api/osmesa/mod.rs
+++ b/src/api/osmesa/mod.rs
@@ -2,11 +2,6 @@
 
 extern crate osmesa_sys;
 
-use std::error::Error;
-use std::ffi::CString;
-use std::fmt::{Debug, Display, Error as FormatError, Formatter};
-use std::{mem, ptr};
-
 use Api;
 use ContextError;
 use CreationError;
@@ -17,6 +12,12 @@ use PixelFormat;
 use PixelFormatRequirements;
 use Robustness;
 use libc;
+
+use std::error::Error;
+use std::ffi::CString;
+use std::fmt::{Debug, Display, Error as FormatError, Formatter};
+use std::{mem, ptr};
+use std::os::raw::c_void;
 
 pub mod ffi {
     pub use super::osmesa_sys::OSMesaContext;
@@ -192,8 +193,8 @@ impl OsMesaContext {
     }
 
     #[inline]
-    pub unsafe fn raw_handle(&self) -> osmesa_sys::OSMesaContext {
-        self.context
+    pub unsafe fn raw_handle(&self) -> *mut c_void {
+        self.context as *mut _
     }
 }
 
--- a/src/api/osmesa/mod.rs
+++ b/src/api/osmesa/mod.rs
@@ -18,6 +18,10 @@ use PixelFormatRequirements;
 use Robustness;
 use libc;
 
+pub mod ffi {
+    pub use super::osmesa_sys::OSMesaContext;
+}
+
 pub struct OsMesaContext {
     context: osmesa_sys::OSMesaContext,
     buffer: Vec<u32>,
@@ -185,6 +189,11 @@ impl OsMesaContext {
     #[inline]
     pub fn get_pixel_format(&self) -> PixelFormat {
         unimplemented!();
+    }
+
+    #[inline]
+    pub unsafe fn as_mut_ptr(&self) -> osmesa_sys::OSMesaContext {
+        self.context
     }
 }
 
--- a/src/api/wgl/mod.rs
+++ b/src/api/wgl/mod.rs
@@ -199,7 +199,7 @@ impl Context {
     }
 
     #[inline]
-    pub unsafe fn as_mut_ptr(&self) -> winapi::HDC {
+    pub unsafe fn raw_handle(&self) -> winapi::HDC {
         self.hdc
     }
 }
--- a/src/api/wgl/mod.rs
+++ b/src/api/wgl/mod.rs
@@ -197,11 +197,6 @@ impl Context {
     pub fn get_pixel_format(&self) -> PixelFormat {
         self.pixel_format.clone()
     }
-
-    #[inline]
-    pub unsafe fn raw_handle(&self) -> winapi::HDC {
-        self.hdc
-    }
 }
 
 unsafe impl Send for Context {}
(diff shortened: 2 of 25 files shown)


tomaka:
Thanks for the PR.

I haven't read the PR in details yet, but I have three remarks:
- I don't think the method should be named `as_mut_ptr`, because that supposes that it returns a pointer.
- I'd prefer if the code in the `platform` module didn't import anything from the `os` module. The retrieval methods should be directly implemented on the context, and not through a trait.
- The `GlContext` trait implementations shouldn't use types that aren't integers or void pointers. I don't know if it's the case right now in your PR, but for example returning an opaque object or a pointer to an opaque object wouldn't be good.

ebkalderon:
I appreciate the helpful feedback!

1. Sounds good. Perhaps `raw_handle()` or `raw()` or some similar permutation be a better name? Or do you have another preference in mind?
2. Will do. Shall I implement `GlContextExt` directly on the top-level `glutin::Context` instead?
3. I've already made sure of that, but please double-check in case of mistakes.

tomaka:
> Sounds good. Perhaps raw_handle() or raw() or some similar permutation be a better name?

I'd say `raw_handle` looks good.

> Will do. Shall I implement GlContextExt directly on the top-level glutin::Context instead?

The `GlContextExt` trait shouldn't be used or mentionned anywhere, except in the files of the `os` module. In other words it should be totally isolated in this module.

It's not necessarily a hard-written rule, but that's how all other traits of `os` work right now, and it's better when everything uses the same design.

ebkalderon:
> The `GlContextExt` trait shouldn't be used or mentionned anywhere, except in the files of the `os` module. In other words it should be totally isolated in this module.

I believe you might have misunderstood my original question, but thanks for the helpful guideline. I was asking whether I should implement `GlContextExt` for `glutin::Context` in the same manner that `WindowExt` is implemented for `winit::Window` ([relevant source](https://docs.rs/winit/0.8.2/src/winit/os/unix.rs.html#128-197)).

tomaka:
> I believe you might have misunderstood my original question, but thanks for the helpful guideline. I was asking whether I should implement GlContextExt for glutin::Context in the same manner that WindowExt is implemented for winit::Window (relevant source).

Well, the answer is yes. Just move the `impl ... for ...` to the various submodules of `os`.
I'm confused by your question because implementing `GlContextExt` on `Context` is the whole point of the PR!

ebkalderon:
Sorry if I wasn't being clear then, haha. I meant implementing for `glutin::Context` in contrast to implementing it against the `platform::Context`s. Give me a break, I have been running very low on coffee these past few days!

src/api/wgl/mod.rs
```diff
@@ -198,6 +197,11 @@ impl Context {
     pub fn get_pixel_format(&self) -> PixelFormat {
         self.pixel_format.clone()
     }
+
+    #[inline]
+    pub unsafe fn raw_handle(&self) -> winapi::HDC {
```
tomaka: Not a void pointer.

src/os/macos.rs
```diff
@@ -1,6 +1,28 @@
...
 pub use winit::os::macos::ActivationPolicy;
 pub use winit::os::macos::MonitorIdExt;
 pub use winit::os::macos::WindowBuilderExt;
 pub use winit::os::macos::WindowExt;
+
+use {Context, HeadlessContext};
+use os::GlContextExt;
+
+impl GlContextExt for Context {
+    type Handle = id;
```
tomaka: Not a void pointer.

src/api/wgl/mod.rs
ebkalderon: Good catch! Thank you, I should have made it into an `HGLRC` instead.

src/api/osmesa/mod.rs
```diff
@@ -186,6 +190,11 @@ impl OsMesaContext {
     pub fn get_pixel_format(&self) -> PixelFormat {
         unimplemented!();
     }
+
+    #[inline]
+    pub unsafe fn raw_handle(&self) -> osmesa_sys::OSMesaContext {
```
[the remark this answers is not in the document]
ebkalderon: Isn't `OSMesaContext` a mutable pointer to an opaque struct meant to be used identically to `*mut c_void`?

src/os/macos.rs
ebkalderon: What type should I use instead? I am not too familiar with OpenGL on Mac OS X.

src/api/osmesa/mod.rs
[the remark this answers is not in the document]
tomaka: They are equivalent, but using a `c_void` makes it more semver-friendly. Sorry for not going into details.
You should return a `*mut c_void` and cast the `OSMesaContext` to it with `as`.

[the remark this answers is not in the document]
ebkalderon: Okay, will do.

src/os/macos.rs
tomaka: A `*mut c_void`. https://github.com/servo/cocoa-rs/blob/master/src/base.rs#L17

ebkalderon: Gotcha. Same rules as with `OSMesaContext`, then.

src/api/osmesa/mod.rs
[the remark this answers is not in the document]
tomaka: No, please use the `std::os::raw` one.
The `libc` dependency should be removed eventually, but since doing so requires a major version bump, I never think about doing it.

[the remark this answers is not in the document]
ebkalderon: Got it, `std::os::raw` it is!

tomaka:
Thanks!

PR #945, opened 2017-10-15
Allow any winit >=0.8.1 to track the 0.8 series automatically

pedrocr:


mitchmindtree:
Wouldn't this mean that if a user depends on glutin and winit publishes some breaking changes (in a 0.9 release for example) that the user's code would possibly break as a result? I think I'd prefer if we kept updating one version at a time to avoid downstream breakages.

pedrocr:
Ah, you're right, I should have done a different condition. Just a sec

pedrocr:
This should now allow `>=0.8.1 && <0.9.0` if I've read the docs correctly here:

http://doc.crates.io/specifying-dependencies.html#caret-requirements

vberger:
Note that the cargo documentation states that the version requirement `"a.b.c"` is interpreted as `"^a.b.c"`, so this PR does not change cargo behavior (though it makes the dependency more explicit).

pedrocr:
No need for this then, closing

PR #955, opened 2017-11-21
Fixed warnings on Mac when building example

paulevans:
Repro using example from "Try it!" section on front page:
cargo run --example window

Warnings were:
warning: unused `#[macro_use]` import
  --> src/lib.rs:38:1
   |
38 | #[macro_use]
   | ^^^^^^^^^^^^
   |
   = note: #[warn(unused_imports)] on by default

warning: value assigned to `res` is never read
   --> src/platform/macos/mod.rs:141:17
    |
141 |             let mut res = false;
    |                 ^^^^^^^
    |
    = note: #[warn(unused_assignments)] on by default

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,6 +35,8 @@
 #[macro_use]
 extern crate lazy_static;
 
+// Caca requires shared_library macro_use on Linux
+#[cfg(target_os = "linux")]
 #[macro_use]
 extern crate shared_library;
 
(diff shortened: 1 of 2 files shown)


paulevans:
I think that was the build env rather than my change, so re-opening to see if the same thing happens again to trigger a travis build.

kvark:
I believe this was fully covered in #963 and can now be closed. Sorry!

paulevans:
Closed as directed

PR #960, opened 2017-12-01
Bump gl_generator to new version

McGittyHub:
This is needed for getting glutin to work when using rustc nightly on release mode on windows.


McGittyHub:
Rejoice!

PR #963, opened 2017-12-15
[macos] find latest GL version

### LINKED ISSUE #781: Software rendered context created on OS X 10.9
I'm using OS X 10.9.5, and tried running the beginnings of a game on my MacBook Pro 2011 with a Radeon 6490M. Whereas it ran at over 400 FPS on my Windows desktop (Radeon 5770 from 2010), it only got roughly 0.25 FPS on my laptop (and the CPU was maxed out). Taking into account the smaller resolution, it was about 6,400 times slower, so I figured it was using software rendering, and I verified it was. Here is how I am creating the window:

```
let display = glutin::WindowBuilder::new()
              .with_dimensions(850, 480)
              .with_title(String::from("Test"))
              //.with_multisampling(4)
              //.with_vsync()
              .with_depth_buffer(24)
              .build_glium()
              .unwrap();
```

I tried with and without multisampling and vsync, and it didn't make any difference.

I tried changing glutin to pass `NSOpenGLPFAAccelerated` and `NSOpenGLPFANoRecovery`, but then it fails to create any context. However, I changed [the version check](https://github.com/tomaka/glutin/blob/master/src/api/cocoa/helpers.rs#L21) to request a 3.2 context on this computer, and then it successfully creates a hardware accelerated context. Can anyone else with 10.9 reproduce this? Should glutin request a 3.2 context on 10.9? It seems bad to create a software rendered context.

kvark:
Fixes #781

When `Latest` non-compatibility version is requested, try 4.1 and 3.2 explicitly instead of relying on software versions.
Also pass in the HW accelerated and double-buffer flags according to the given options.
Note: according to the [docs](https://developer.apple.com/documentation/appkit/1436213-anonymous/nsopenglpfadoublebuffer?language=objc), not passing the double-buffer flag equals to requesting single buffering.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 Cargo.lock
 target/
+.DS_Store
 *~
 #*#
--- a/src/headless.rs
+++ b/src/headless.rs
@@ -33,7 +33,10 @@ impl<'a> HeadlessRendererBuilder<'a> {
     pub fn new(width: u32, height: u32) -> HeadlessRendererBuilder<'a> {
         HeadlessRendererBuilder {
             dimensions: (width, height),
-            pf_reqs: Default::default(),
+            pf_reqs: PixelFormatRequirements {
+                hardware_accelerated: None,
+                .. Default::default()
+            },
             opengl: Default::default(),
             platform_specific: Default::default(),
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,6 +35,7 @@
 #[macro_use]
 extern crate lazy_static;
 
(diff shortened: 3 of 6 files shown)


kvark:
The only non-mac specific change is with the headless context parameters, and it's also the one I'm least confident about. Previously, mac platform didn't ever pass the accelerated pixel format requirement, which resulted, so headless contexts worked. Now that I made it respect the acceleration flag, I had to avoid the default pixel requirements for the headless mode.

This is confusing, since I don't see headless context having clear semantics (hello, pure contexts of #942).

Anyhow, this PR is ready for review now, @tomaka . It is somewhat important since currently any Mac not supporting GL 4.1 switches to software rendering.

PR #964, opened 2017-12-19
Update dependencies

chetgurevitch:


tomaka:
I think this is a breaking change for MacOS.

PR #966, opened 2017-12-20
Update EmscriptenWebGLContextAttributes FFI to match emscripten 1.37.

codewiz:
The explicitSwapControl field was added to emscripten's API in 2016.

Without this field, the Rust FFI bindings allocate too little space for the struct, causing accesses to uninitialized memory.


tomaka:
Thanks!

PR #971, opened 2017-12-28
[osx] default PF attribs to use double buffering

### LINKED ISSUE #969: OS X: examples on current master don't work (blank screen)
On my Mac machine, examples don't run correctly (current master branch, 0.11 still works fine).  Windows are always empty.

According to @kvark, this could be connected to #963.

I noticed that, in contrast to my windows machine, double_buffer is always false. See output at start:

> Pixel format of the window's GL context: PixelFormat { hardware_accelerated: true, color_bits: 24, alpha_bits: 8, depth_bits: 24, stencil_bits: 8, stereoscopy: false, double_buffer: false, multisampling: None, srgb: true }
OpenGL version 4.1 NVIDIA-10.28.10 355.11.10.10.20.111

kvark:
Fixes  #969
cc @eldyer
r? @tomaka

### DIFF
--- a/src/platform/macos/helpers.rs
+++ b/src/platform/macos/helpers.rs
@@ -81,7 +81,10 @@ pub fn build_nsattributes(
         attributes.push(NSOpenGLPFAAccelerated as u32);
     }
 
-    if let Some(true) = pf_reqs.double_buffer {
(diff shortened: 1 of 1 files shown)


eldyer:
Confirmed, this fixes #969, examples are working again.

PR #975, opened 2018-01-02
Upgrade glutin to use winapi 0.3

bbrown683:
I went ahead and added support for winapi 0.3. There should be a very noticeable boost in the compilation speed as a result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,3 @@
 # Unreleased
-
+- Add support for winapi 0.3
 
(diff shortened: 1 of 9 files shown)


tomaka:
Thanks, looks good!
Could you add a line in `CHANGELOG.md` as well? (for example like [this model](https://raw.githubusercontent.com/tomaka/winit/011720848a43c52e7f0444302494c7b3ce39033c/CHANGELOG.md), since you're the first person ever to do that)

bbrown683:
Done!

PR #977, opened 2018-01-08
[macos] select compatibility profile if gl version is low

kvark:
This is a hotfix to #963

When the user (@ozkriff in this case... Ohai!) is not explicit about the profile they need, the code doesn't consider compatibility profile, but it should.
~~Also contains a version bump for supposed publishing. Not sure how you track release branches/versions? I don't see either branches or tags for latest releases.~~

r? @tomaka

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,3 @@
 # Unreleased
-- Add support for winapi 0.3
-
+- Add support for winapi 0.3 ([#975](https://github.com/tomaka/glutin/pull/975))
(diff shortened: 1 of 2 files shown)


tomaka:
Looks good. Needs an entry in the Changelog.

kvark:
@tomaka done

tomaka:
Also, you should either not bump glutin's version in the Cargo.toml, or adjust the changelog for the publication of 0.12.1 that will happen automatically if you do bump the version.

kvark:
@tomaka ok, I reverted the version change. Affected parties can just use explicit `Compatibility` profile in the mean time.

PR #995, opened 2018-02-23
Update winit to v0.11

MortimerGoro:


tomaka:
Please add an entry in the CHANGELOG as well.

PR #998, opened 2018-03-04
Add the `windows-static-egl` Cargo feature

### LINKED ISSUE #997: Support statically-linked libEGL? (On Windows)
https://github.com/servo/mozangle compiles ANGLE, on Windows with an EGL implementation. To avoid the complication of shipping `.dll` files separately, it is linked as a static library.

Glutin supports EGL on Windows, but loads it from a dynamic library like `libEGL.dll`. What do you think of adding support for statically-linked EGL?

Conditionally having something like `extern { fn eglGetProcAddress(…) }` directly in Glutin would require a compile-time flag, which isn’t great. So maybe that function could be (optionally) provided by users:

```rust
impl<'a> ContextBuilder<'a> {
    pub fn with_egl_get_proc_address<F>(mut self, get_proc_address: F) -> Self
        where F: Fn(&str) -> *const c_void
    {
        self.egl_get_proc_address = Some(Box::new(get_proc_address))
        self
    }
}
```

Since this is only relevant on Windows, maybe it should be in a `glutin::os::windows::ContextBuilderExt` trait.

Or maybe something more automagic could be made with weak symbols, but those don’t seem to be available in Stable Rust.

Thoughts?

SimonSapin:
Fixes #997: this allows using Glutin together with https://crates.io/crates/mozangle

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Fix android activity life cycle
+- Add the `windows-static-egl` Cargo feature. (See crate’s doc-comment.)
 
 # Version 0.13.0 (2018-02-28)
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "glutin"
-version = "0.13.0"
+version = "0.13.1"
 authors = ["The glutin contributors, Pierre Krieger <pierre.krieger1708@gmail.com>"]
 description = "Cross-platform OpenGL context provider."
 keywords = ["windowing", "opengl"]
@@ -9,6 +9,9 @@ readme = "README.md"
 repository = "https://github.com/tomaka/glutin"
 documentation = "https://docs.rs/glutin"
 build = "build.rs"
+
+[features]
+windows-static-egl = []
 
 [dependencies]
 lazy_static = "1"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,3 +15,4 @@ build: false
 
 test_script:
   - cargo test --verbose
+  - cargo build --features windows-static-egl --example windows_static_egl
--- a/examples/windows_static_egl.rs
+++ b/examples/windows_static_egl.rs
@@ -1,0 +1,20 @@
+//! This incomplete example will likely crash at runtime on Windows
+//! when the `windows-static-egl` feature is enabled,
+//! but it should still compile and link.
+
+extern crate glutin;
+
+use std::os::raw::{c_char, c_void};
+
+#[allow(non_snake_case)]
+#[no_mangle]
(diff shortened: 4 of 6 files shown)


SimonSapin:
Should mozangle be mentioned in the doc as a possible way to use this feature?

Should mozangle be added as in `dev-dependencies` and the program below added under `examples/`?

```rust
#[cfg(not(windows))]
fn main() {
    println!("This example only runs on Windows.");
}

#[cfg(windows)] extern crate mozangle;
#[cfg(windows)] extern crate glutin;

#[cfg(windows)]
fn main() {
    use glutin::GlContext;
    use glutin::os::GlContextExt;
    use glutin::os::windows::RawHandle;
    use mozangle::egl::get_proc_address;
    use mozangle::gles::ffi as gl;

    let mut events_loop = glutin::EventsLoop::new();
    let window = glutin::WindowBuilder::new();
    let context = glutin::ContextBuilder::new()
        // Requesting OpenGlEs is required to make glutin try to use EGL.
        .with_gl(glutin::GlRequest::Specific(glutin::Api::OpenGlEs, (2, 0)));
    let gl_window = glutin::GlWindow::new(window, context, &events_loop).unwrap();

    let handle = unsafe {
        gl_window.context().raw_handle()
    };
    match handle {
        RawHandle::Egl(_) => {},
        RawHandle::Wgl(_) => panic!("Expected an EGL context, found a WGL one.")
    }

    assert_ne!(gl_window.get_proc_address("glClearColor"), 0 as _);
    assert_eq!(gl_window.get_proc_address("glClearColor"), get_proc_address("glClearColor") as _);

    unsafe {
        gl_window.make_current().unwrap();
        gl::ClearColor(0., 0.5, 0., 1.);
    }

    events_loop.run_forever(|event| {
        match event {
            glutin::Event::WindowEvent { event, .. } => match event {
                glutin::WindowEvent::Closed => return glutin::ControlFlow::Break,
                glutin::WindowEvent::Resized(w, h) => gl_window.resize(w, h),
                _ => (),
            },
            _ => ()
        }

        unsafe {
            gl::Clear(gl::COLOR_BUFFER_BIT);
        }

        gl_window.swap_buffers().unwrap();
        glutin::ControlFlow::Continue
    });
}
```

SimonSapin:
The example program in the previous comment is subject to https://github.com/tomaka/winit/issues/415. (Maybe Clear + swap_buffers is slow in debug mode? And re-rendering on non-resize events is unnecessary.)

tomaka:
I'm not fan of this feature but I guess I will merge anyway.

Please solve the conflict, and also please remove the example because for such a niche feature I don't think an example really makes sense.

SimonSapin:
I’m not really a fan of it either. I think it’s preferable for WebRender/Wrench to use winit + EGL instead of this.

PR #1002, opened 2018-03-15
Make the `pf_reqs` field of `ContextBuilder` public.

SimonSapin:
Its type `PixelFormatRequirements` is already public with public fields.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "glutin"
-version = "0.13.1"
+version = "0.13.2"
 authors = ["The glutin contributors, Pierre Krieger <pierre.krieger1708@gmail.com>"]
 description = "Cross-platform OpenGL context provider."
 keywords = ["windowing", "opengl"]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -145,7 +145,7 @@ pub struct ContextBuilder<'a> {
     /// The attributes to use to create the context.
     pub gl_attr: GlAttributes<&'a Context>,
     // Should be made public once it's stabilized.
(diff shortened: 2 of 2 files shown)


tomaka:
This field was made private on purpose, although I don't remember why :-/

jdm:
According to https://github.com/tomaka/glutin/pull/608 it's because the members of PixelFormat were not considered stable that the time.

jdm:
According to https://github.com/tomaka/glutin/blame/209ab925f5325c35b8c35020e2d9a85f519f84c3/src/lib.rs#L817 the struct members have not changed in 3 years.

francesca64:
@jdm thanks for investigating this!

@SimonSapin are you still interested in this?

SimonSapin:
Yes, this would allow fixing a bug in some code in WebRender’s Wrench, even though nothing happens to "exercise" that bug at the moment.

This was for https://github.com/servo/webrender/pull/2529 where I forked some of Glutin’s `src/api/egl/mod.rs` in order to create an (E)GL context, but with statically-linked ANGLE. So this code receives `ContextBuilder` as a parameter, and needs to access its details.

The work-around was to use `PixelFormatRequirements::default()`, which was ok because the only user of this code happened not to use any `ContextBuilder` API that would change the pixel format requirements from the defaults.

Forking this EGL code would not be necessary in the first place if Glutin itself could support statically-linked EGL/ANGLE. I attempted that in https://github.com/tomaka/glutin/pull/998, but it resulted in an awkward public API.

francesca64:
Okay, cool. I'm not opposed to this, so provided you add a CHANGELOG entry we can merge this.

PR #1010, opened 2018-04-06
Update to version 0.12 of winit. Publish 0.14.0. Update CHANGELOG. Address deprecation warning in wayland backend.

mitchmindtree:
This also updates the wayland backend to not use deprecated
`get_inner_size_points` method and instead calculate the points from the
`get_inner_size` and `hidpi_factor` methods.


mitchmindtree:
cc @tomaka is there anything I can do to help move this through?

PR #1011, opened 2018-04-14
Use HTTPS Heroku badge

salty-horse:
The http access causes a mixed-content warning when visiting https://crates.io/crates/glutin


tomaka:
Thanks

PR #1014, opened 2018-04-24
Add error message to `NotSupported` variant

derekdreery:
This PR just adds some static strings to help debug a missing library. It's a problem I have had, and others on chat have had, and knowing the cause would have saved me time :smiley:


derekdreery:
The CI will fail because I need to push messages for non-unix platforms.

tomaka:
Thanks!

PR #1020, opened 2018-05-06
[Do Not Merge] Use EGL for Linux headless contexts

### LINKED ISSUE #505: Create a pixmap or a pbuffer on X11 for the headless context
Instead of using OSMesa by default, try to use hardware acceleration first.

fintelia:
OsMesa isn't the only way to make headless contexts on Linux. It would be nice to also support EGL headless contexts. This PR sketches out how to accomplish this. It is functional (on my Ubuntu machine...) but lacks some relevant error handling and any logic to select whether to use EGL or OsMesa.

This would resolve #505

### DIFF
--- a/src/platform/linux/mod.rs
+++ b/src/platform/linux/mod.rs
@@ -131,12 +131,9 @@ impl Context {
 #[derive(Clone, Default)]
 pub struct PlatformSpecificHeadlessBuilderAttributes;
 
-pub struct HeadlessContext(OsMesaContext);
-
-impl HeadlessContext {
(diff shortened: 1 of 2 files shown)


ZeGentzy:
#1031 and #1029 Will allow you to choose which backend to use & implements sharing. Would those be good?

ZeGentzy:
They all support "headless" contexts, as in, contexts with no visible window. Of course, some backends, like the emscripten and linux ones, achieve this by just spawning a hidden window.

PR #1021, opened 2018-05-09
Update to winit 0.15.0 + release glutin 0.16.0

francesca64:

PR #1022, opened 2018-05-16
Upgrade to winit 0.14

Xaeroxe:
A new cargo release of glutin following merging this would be appreciated.


vberger:
Seems like a duplicate of #1021

Xaeroxe:
Ah, so it is.  My bad, I should review existing PRs before I open mine.  Well I guess I should explain myself, we're trying to fix a bug in Amethyst and we think upgrading winit might help with that, but in order to do that we have to domino effect upgrade a few of our dependencies and this is one of them.

PR #1026, opened 2018-06-07
Fix gpu switching

### LINKED ISSUE #980: regression in 0.12, forces use of dedicated GPU on macbook
version 0.12 appears to have a regression of #894 which allowed apps using glutin to use the integrated GPU on macbooks. 0.12 appears to force the dedicated GPU on all the time

mkeeler:
Fixes #980

This PR contains 2 things. First users of the ContextBuilder can now specify hw accel and double buffering whereas before they were defaulted and couldn't be altered in this manner. Secondly on macOS when probing to determine the latest GL version we now take into account any requested attributes from the PixelFormatRequirements. This is to prevent forcing using a discrete GPU on macs when its not actually needed/desired.

Note: The addition of NSOpenGLPFAAllowOfflineRenderers is required for automatic graphics switching to work and the main build_nsattributes already always includes this attribute. In my testing of glutin via gfx-rs this attribute will not prevent an application from using the discrete gpu but rather just enable the switching if applicable.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -318,6 +318,24 @@ impl<'a> ContextBuilder<'a> {
         self.pf_reqs.srgb = srgb_enabled;
         self
     }
+    
+    /// Sets whether double buffering should be enabled
+    ///
+    /// The default value is `None`.
+    #[inline]
+    pub fn with_double_buffer(mut self, enabled: Option<bool>) -> Self {
+        self.pf_reqs.double_buffer = enabled;
+        self
+    }
+    
+    /// Sets whether hardware acceleration is required
+    ///
+    /// The default value is `Some(true)`
+    #[inline]
+    pub fn with_hw_accel(mut self, enabled: Option<bool>) -> Self {
+        self.pf_reqs.hardware_accelerated = enabled;
+        self
+    }
 }
 
 impl GlWindow {
--- a/src/platform/macos/helpers.rs
+++ b/src/platform/macos/helpers.rs
@@ -9,7 +9,8 @@ use cocoa::appkit::*;
 use cocoa::base::nil;
 
 pub fn get_gl_profile<T>(
-    opengl: &GlAttributes<&T>
+    opengl: &GlAttributes<&T>,
+    pf_reqs: &PixelFormatRequirements
 ) -> Result<NSOpenGLPFAOpenGLProfiles, CreationError> {
     let version = opengl.version.to_gl_version();
     // first, compatibility profile support is strict
@@ -37,14 +38,26 @@ pub fn get_gl_profile<T>(
     } else if let GlRequest::Latest = opengl.version {
         // now, find the latest supported version automatically
         let mut attributes = vec![
-            // Note: we assume here that non-accelerated contexts don't care
-            NSOpenGLPFAAccelerated as u32,
-            NSOpenGLPFAOpenGLProfile as u32,
-            0, // this is where we put the test profile
-            0,
+            NSOpenGLPFAAllowOfflineRenderers as u32,
         ];
+        
+        if let Some(true) = pf_reqs.hardware_accelerated {
+            attributes.push(NSOpenGLPFAAccelerated as u32);
+        }
+
+        if pf_reqs.double_buffer != Some(false) {
+            attributes.push(NSOpenGLPFADoubleBuffer as u32);
+        }
+        
+        attributes.push(NSOpenGLPFAOpenGLProfile as u32);
+        // where the gl profile should be stored
+        attributes.push(0);
+        attributes.push(0);
+        
+        let profile_idx = attributes.len() - 2;
+            
         for &profile in &[NSOpenGLProfileVersion4_1Core, NSOpenGLProfileVersion3_2Core] {
-            attributes[2] = profile as u32;
+            attributes[profile_idx] = profile as u32;
             let id = unsafe {
                 NSOpenGLPixelFormat::alloc(nil).initWithAttributes_(&attributes)
             };
--- a/src/platform/macos/headless.rs
+++ b/src/platform/macos/headless.rs
@@ -26,7 +26,7 @@ impl HeadlessContext {
                _: &PlatformSpecificHeadlessBuilderAttributes)
                -> Result<HeadlessContext, CreationError>
     {
-        let gl_profile = helpers::get_gl_profile(opengl)?;
+        let gl_profile = helpers::get_gl_profile(opengl, pf_reqs)?;
         let attributes = helpers::build_nsattributes(pf_reqs, gl_profile)?;
         let context = unsafe {
             let pixelformat = NSOpenGLPixelFormat::alloc(nil).initWithAttributes_(&attributes);
--- a/src/platform/macos/mod.rs
+++ b/src/platform/macos/mod.rs
@@ -61,7 +61,7 @@ impl Context {
 
         let view = window.get_nsview() as id;
 
-        let gl_profile = helpers::get_gl_profile(gl_attr)?;
+        let gl_profile = helpers::get_gl_profile(gl_attr, pf_reqs)?;
         let attributes = helpers::build_nsattributes(pf_reqs, gl_profile)?;
         unsafe {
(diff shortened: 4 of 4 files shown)


src/lib.rs
```diff
@@ -318,6 +318,24 @@ impl<'a> ContextBuilder<'a> {
         self.pf_reqs.srgb = srgb_enabled;
         self
     }
+    
+    /// Sets whether double buffering should be enabled
+    ///
+    /// The default value is `None`.
+    #[inline]
+    pub fn with_double_buffer(mut self, enabled: Option<bool>) -> Self {
```
francesca64: It's pretty minor, but I'd prefer a more specific parameter name than `enabled`, to better keep with the style of the other methods.

src/lib.rs
```diff
@@ -318,6 +318,24 @@ impl<'a> ContextBuilder<'a> {
...
+        self.pf_reqs.double_buffer = enabled;
+        self
+    }
+    
+    /// Sets whether hardware acceleration is required
+    ///
+    /// The default value is `Some(true)`
+    #[inline]
+    pub fn with_hw_accel(mut self, enabled: Option<bool>) -> Self {
```
francesca64: I'd prefer you don't abbreviate this method name. It's not like people have to call this more than once, so making it easier to understand and consistent with the member name would be nicer.

src/platform/macos/helpers.rs
```diff
@@ -9,7 +9,8 @@ use cocoa::appkit::*;
 use cocoa::base::nil;
 
 pub fn get_gl_profile<T>(
-    opengl: &GlAttributes<&T>
+    opengl: &GlAttributes<&T>,
+    pf_reqs: &PixelFormatRequirements
```
francesca64: Multi-line parameter lists should have trailing commas.

src/platform/macos/helpers.rs
```diff
@@ -37,14 +38,26 @@ pub fn get_gl_profile<T>(
     } else if let GlRequest::Latest = opengl.version {
         // now, find the latest supported version automatically
         let mut attributes = vec![
```
francesca64: We know the needed capacity here, so please use `Vec::with_capacity`. While you could specify the maximum possible capacity (which looks like it's 6), it would be much cooler to cast bools to `usize` and do addition so that you request the exact capacity wrt `PixelFormatRequirements`.

By that same token, you should be able to avoid the heap allocation entirely by creating an array of the maximum needed size, and then getting a slice using the capacity value I advised above.

src/lib.rs
```diff
@@ -318,6 +318,24 @@ impl<'a> ContextBuilder<'a> {
         self.pf_reqs.srgb = srgb_enabled;
         self
     }
+    
+    /// Sets whether double buffering should be enabled
```
francesca64: This sentence should end in a period.

src/platform/macos/helpers.rs
mkeeler: I like the static array approach. 6 u32s on the stack should be no problem at all.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+- Fix regression that prevented automatic graphics switching in MacOS (#[980](https://github.com/tomaka/glutin/issues/980))
```
francesca64: The octothorpe should be part of the link.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
...
+- Add ContextBuilder::with_double_buffer function
```
francesca64: and don't forget to use backticks.

CHANGELOG.md
mkeeler: Sure whichever way you like. 

There are these in the CHANGELOG already so I wasn't sure which way to go.

```
- Add support for winapi 0.3 ([#975](https://github.com/tomaka/glutin/pull/975))
- Fix MacOS to return compatibility profile if applicable (#[977](https://github.com/tomaka/glutin/pull/977))
```

src/lib.rs
```diff
@@ -319,21 +319,21 @@ impl<'a> ContextBuilder<'a> {
...
-        self.pf_reqs.double_buffer = enabled;
+    pub fn with_double_buffer(mut self, double_buffer: Option<bool>) -> Self {
+        self.pf_reqs.double_buffer = double_buffer;
         self
     }
     
-    /// Sets whether hardware acceleration is required
+    /// Sets whether hardware acceleration is required.
     ///
     /// The default value is `Some(true)`
```
francesca64: There should be a period here, too.

mkeeler:
For anyone who can test this out too:

`git clone https://github.com/mkeeler/glutin -b fix-gpu-switching`

Then in your projects Cargo.toml add

```
[patch.crates-io]
glutin = { path = "<path to the clone you just did>" }
```

When running the program you can use Activity Monitor to check if it requires the high perf gpu:

gpu

My test hardware:
* Macbook Pro 14,3
* MacOS 10.13.5
* Radeon Pro 560 & Intel HD 630

Test Programs:
* Alacritty - Does not require the high perf gpu. Does get a nice perf boost when it is in use though (on ac power or when another app that requires the high perf gpu is running).
* [gfx-rs flowmap example](https://github.com/gfx-rs/gfx/tree/pre-ll/examples/flowmap) - This requires the high perf gpu and forces macos to automatically switch over to it as soon as this program is run.

Notes:

The controlling terminal/shell, Terminal.app/Alacritty + bash in my case end up being marked as requiring the high perf gpu when the flowmap example runs and stays that way even after it exits. I don't know why the parent non-gl aware processes are affected but this sounds like a MacOS bug and unrelated to glutin.

francesca64:
> path to the clone you just did

This should be easier:
```toml
[patch.crates-io]
glutin = { path = "https://github.com/mkeeler/glutin", branch = "fix-gpu-switching" }
```

mkeeler:
nitpick away, I understand completely. I circled back and addressed double checked that all the comments were addressed and the only other one I needed to fix was the backticks in the changelog.

A couple additional notes.

* I ran the conrod examples documented here: http://docs.piston.rs/conrod/conrod/guide/chapter_2/index.html#running-the-conrod-examples and the end result is that the high perf gpu gets enabled.
* Secondly the note about the process requiring the high perf gpu seems like a bug in Activity Monitor or however macos tracks it. The high perf gpu actually gets transitioned back over to the integrated gpu when the gfx-rs and the conrod/piston examples are exited.

gabrielalmeida:
@mkeeler FYI

> For anyone who can test this out too:
>
> In your projects Cargo.toml add
>
>```
>[patch.crates-io]
>glutin = { git = "https://github.com/mkeeler/glutin", branch = "fix-gpu-switching" }
> ```
>

I've got the same result by building Alacritty master with your glutin's branch. It doesn't trigger discrete GPU anymore and reduced battery usage avg from ~40%+ to ~3%. Same hardware spec as yours and Mac OS 10.13.3. I couldn't notice any significant performance penalty.

Thanks!

francesca64:
Well, I'm ready to merge this. `ggez` is switching to glutin/winit, so if this causes problems for people who need high performance I'll find out soon enough.

One last thing though. Since the new builder methods are only implemented on one platform, that should be documented (and mentioned in the CHANGELOG), since people might get confused if they try to use them and they don't work. For consistency, this should be done in the same style as in winit: https://github.com/tomaka/winit/blob/e7a8efcfa0b0b3b6518fb9f11541cd5a8895965f/src/window.rs#L55

mkeeler:
@francesca64 It should be done. The options can actually apply to other non-macos platforms. They are just handled in src/api/egl, src/api/glx and src/api/wgl.

One thing I found is that the GLX api provider looks like it could be made to take the hardware acceleration value into account here:

https://github.com/tomaka/glutin/blob/master/src/api/glx/mod.rs#L479-L480

Similar to how its done for EGL here:

https://github.com/tomaka/glutin/blob/master/src/api/egl/mod.rs#L557-L564

That isn't a problem for me but something I noticed while determining which platforms my changes might affect.

PR #1031, opened 2018-06-17
Adds context sharing support for the X11 & Wayland backends

ZeGentzy:
Depends on: https://github.com/tomaka/winit/pull/572 https://github.com/tomaka/glutin/pull/1029

Currently both backends support shared contexts however the wayland backend requires they are made with the same events pool.


ZeGentzy:
Should be ready for review once https://github.com/tomaka/glutin/pull/1029 is done

src/lib.rs
```diff
@@ -18,16 +18,11 @@
 //! # }
 //! ```
 //!
-//! For contexts that are *not* associated with any particular window, see the HeadlessContext
-//! type.
-//!
 //! # Features
 //!
-//! This crate has two Cargo features: `window` and `headless`.
+//! This crate has two Cargo features: `window`.
```
fintelia: Presumably this should now say "This crate has *one* Cargo feature: window".

ZeGentzy:
Moved: https://github.com/tomaka/glutin/pull/1045

PR #1035, opened 2018-06-25
Update to winit 0.16.0 + release glutin 0.17.0

francesca64:
Besides all of the improvements included in winit 0.16.0, I've also fixed the iOS backend and implemented all of the `GlContext` methods for it. While it now compiles and runs, I only ever managed to get a black screen, though it's possible I was just doing a bad job with raw OpenGL calls... when I get a chance, I'll try running the glutin examples.

I removed two of the example programs, since they weren't really relevant to glutin anymore. Besides that, I switched all of the examples from `run_forever` to `poll_events`, since rendering in a `run_forever` handler causes the event queue to become clogged up rapidly.


CHANGELOG.md
```diff
@@ -1,8 +1,12 @@
 # Unreleased
 
+# Version 0.16.0 (2018-06-27)
```
petobens: This is probably `0.17.0` right?

francesca64: Whoops! Thanks for pointing this out.

PR #1036, opened 2018-06-26
Work around an Android emulator bug(?)

SimonSapin:
Do not pass a `CONTEXT_FLAGS_KHR` attribute to `egl.CreateContext` when it is at the default value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 - Fix regression that prevented automatic graphics switching in MacOS ([#980](https://github.com/tomaka/glutin/issues/980))
 - Add `ContextBuilder::with_double_buffer` function
 - Add `ContextBuilder::with_hardware_acceleration` function
+- Work around a presumed bug Android emulator bug
+  that would cause context creation to return `CreationError::OpenGlVersionNotSupported`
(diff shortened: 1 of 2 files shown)


SimonSapin:
For posterity: this is with emulator version 27.2.9 started with `-gpu guest` running on Linux with the image `system-images;android-28;google_apis;x86e`. (This is the only config where I was able to get GLES3 support on Linux, so far.)

francesca64:
This looks alright to me (though I still need to actually test it), but I'd like a CHANGELOG entry.

SimonSapin:
Sure. Should the entry describe the change factually, or its goal? In how much detail?

francesca64:
[winit's CHANGELOG](https://github.com/tomaka/winit/blob/master/CHANGELOG.md) is the closest thing I have to a style guide, though I admit I should really make a CONTRIBUTING that explains this.

Generally speaking, the entry shouldn't include implementation details, and should just concern the practical impact on application developers. In this case, enough information that someone who's encountered this issue before would recognize that it's been fixed.

SimonSapin:
Sounds good, added.

francesca64:
Thanks!

PR #1045, opened 2018-07-10
Adds context sharing support for the X11 & Wayland backends

ZeGentzy:
Depends on: tomaka/winit#572 #1029

Now both backends support shared contexts, however the wayland backend requires that shared contexts are made with the same events loop.

Signed-off-by: Hal Gentz zegentzy@protonmail.com


marek-g:
Oh, I've implemented context sharing for the X11 today. I looked here after (bad luck). My change is much simpler (and it isn't API breaking) but I believe it's because I do not support Wayland.

I think it makes no sense to create pull request now, but if someone wants to look, it's here:
https://github.com/marek-g/glutin/commit/b30ba37c0e18e9f451f4f16f81c3380782c6e3e2

marek-g:
@ZeGentzy You may be right. I've only tested it with X11/WGL. I can't wait for your changes to be merged :)

ZeGentzy:
Merged with https://github.com/tomaka/glutin/pull/1058

PR #1051, opened 2018-07-27
[WIP] Raw contexts

ZeGentzy:
Depends on: https://github.com/tomaka/glutin/pull/1045

A new API which allows us to make a context for a raw window. I plan to implement it for linux and windows. Currently untested.

Implemented:
 - [x] Linux
 - [ ] Windows

Tested:
 - [ ] Linux
 - [ ] Windows


ZeGentzy:
After working on this for some while, I realized it would be easier and more appropriate to add a `from_raw` function to winit's windows and eventsloops.

Issue #1055, opened 2018-08-02
Update to winit 0.17.0


felixrabe:
As per tomaka/winit@a32f7f2ec575 and tomaka/winit#608, winit 0.17.0 is [released](https://github.com/tomaka/winit/releases).

Pull request: #1056.

PR #1056, opened 2018-08-02
Update winit to 0.17.0 (#1055)

felixrabe:
That seems to work fine.

Tested on macOS 10.13.6 with the following commands:

    cargo test
    cargo run --example window
    cargo run --example fullscreen
    cargo run --example transparent  # ❤️ that one
    cargo run --example multiwindow


felixrabe:
Incoming!

francesca64:
Great!

felixrabe:
Never mind the "revert" commit above – I played around thinking "GitHub sure does not immediately create a revert (commit) when pushing that `Revert` button"...

PR #1072, opened 2018-10-15
Only toggle fullscreen on key press in fullscreen example

kuviman:
Otherwise you go back fullscreen when you release the key


francesca64:
Good catch! Thanks for the PR.

Issue #949, opened 2017-11-02
Ubuntu 17.10 / Wayland Error


gaites:
I have no idea what I am doing, so if there's an obvious solution, I'm sorry. 

> thread 'main' panicked at 'Library libwayland-egl.so could not be loaded.', /checkout/src/libcore/option.rs:839:4

> echo $WAYLAND_DISPLAY gives wayland-0

Ubuntu 17.10, Gnome 26. setting WAYLAND_DISPLAY=wayland-1 fixes the problem. Both unset and wayland-0 cause the error above.  

I fully expect that there's an obvious solution that I'm missing, any help appreciated. Thanks!

vberger:
Do you have the package `libwayland-egl1-mesa` installed? (assuming you are using the mesa graphic drivers)

gaites:
Here's the output of `sudo apt install libwayland-egl1-mesa`:

> libwayland-egl1-mesa is already the newest version (17.2.2-0ubuntu1).
libwayland-egl1-mesa set to manually installed.
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded. `

and the output of `lshw -c video`:

> description: VGA compatible controller
       product: 2nd Generation Core Processor Family Integrated Graphics Controller
       vendor: Intel Corporation
       physical id: 2
       bus info: pci@0000:00:02.0
       version: 09
       width: 64 bits
       clock: 33MHz
       capabilities: vga_controller bus_master cap_list rom
       configuration: driver=i915 latency=0
       resources: irq:35 memory:c0000000-c03fffff memory:b0000000-bfffffff ioport:6000(size=64) memory:c0000-dffff

I'm not sure if this would make any difference. I do have a hybrid graphics with an older AMD Radeon mobile card, but I'm not running with DRI_PRIME=1.

vberger:
Oh, I suspect I see the issue, which of the following files do you have on your system?

- `/usr/lib/libwayland-egl.so`
- `/usr/lib/libwayland-egl.so.1`
- `/usr/lib/libwayland-egl.so.1.0.0`

gaites:
We've got a winner in another directory:

> /usr/lib/x86_64-linux-gnu/libwayland-egl.so.1
/usr/lib/x86_64-linux-gnu/libwayland-egl.so.1.0.0

Is this a call for a symlink to /lib/usr?

vberger:
[...]

PR #1075, opened 2018-10-28
Wayland: make libEGL loading behaviour consistent

mkdewidar:
The x11 implementation will try to load libEGL.so if so.1 is not
available but the wayland one didn't and so either required the user to
create the sym link themselves or install the devel packages which did
it for them.

This resolves the latter issues faced in #949


mkdewidar:
Tested using `cargo run --example window` with only the `libEGL.so.1` in `/usr/lib64/` on Fedora Workstation 28 and all seemed to work fine.

francesca64:
This looks good to me, though this needs a CHANGELOG entry.

@vberger does this look good to you?

vberger:
LGTM yes :+1:

mkdewidar:
Sorry for the delayed reply, updated the CHANGELOG now.

francesca64:
Great, thanks!

Issue #1078, opened 2018-11-07
Expose the EGLDisplay


ceyusa:
I'm working on the [Rust bindings for GstGL](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/merge_requests/161) and I'm adding an example, in which I use glutin.

The example will render video frames as textures, but in order to share the window's GL context with the video processing pipeline, I need to create a wrapped context object, for which I need, in the case of EGL (which is my main target), the context's raw handler :ballot_box_with_check:, the GL API :ballot_box_with_check:, the platform :ballot_box_with_check: and the EGLDisplay :negative_squared_cross_mark:

My naive approach is to add a ``get_native_display()`` in ``GlContextExt`` trait which will boil down for each OS:

* Linux
  * EGLDisplay if EGL context
  * XDisplay if GLX context
* MacOS and iOS
  * I don't know :worried: 
* Windows
  * EGLDisplay if EGL context
  * Nothing if WGL 
* Android
  * EGLDisplay

What do you think?

ceyusa:
ping? :sweat_smile:

PR #1082, opened 2018-11-08
Add a way to retrieve the EGL display

ceyusa:
to say EGLDisplay ~or XDisplay~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Added NetBSD support.
 - **Breaking:** Removed `new_shared` function from `Context` and `GlWindow`, in favor of `new`.
 - Added `build` method to `ContextBuilder`.
+- Added `get_egl_display` method to `GlContextExt` trait and its implementation for platforms.
 
 # Version 0.19.0 (2018-11-09)
 
--- a/src/api/android/mod.rs
+++ b/src/api/android/mod.rs
@@ -165,4 +165,9 @@ impl Context {
     pub unsafe fn raw_handle(&self) -> egl::ffi::EGLContext {
         self.0.egl_context.raw_handle()
     }
+
+    #[inline]
+    pub unsafe fn get_egl_display(&self) -> egl::ffi::EGLDisplay {
+        self.0.egl_context.get_egl_display()
+    }
 }
--- a/src/api/egl/ffi.rs
+++ b/src/api/egl/ffi.rs
@@ -4,6 +4,7 @@
 extern crate winapi;
 
 pub use self::egl::types::EGLContext;
+pub use self::egl::types::EGLDisplay;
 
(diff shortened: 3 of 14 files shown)


jdm:
@francesca64 Are you the right person to review these changes?

jdm:
Note: this is addressing #1078.

francesca64:
@jdm I'm the right person, but there are currently a few other things I need to review, so it can be hard to find time.

@ceyusa thanks for the PR!

`Window::native_display` follows the style of the `platform_display` and `platform_window` methods that were deprecated in glutin/winit a while ago, and were recently removed. The convention we've moved to is using `WindowExt` methods per-platform, each with their own documentation and descriptive name (`native_display` isn't really an acceptable name, simply because it's so similar to the name of a method that was recently removed).

I'm also not a fan of having external types exposed, since it means that dependency upgrades could be breaking changes. We use void pointers (`*mut std::os::raw::c_void`) to get around this issue (see #939 for discussion).

Having macOS and iOS return `0` isn't idiomatic, and is very surprising/dangerous. My recommendation is to just create `WindowExt` methods (probably `get_egl_display`) for the platforms that can have an `EGLDisplay`. You can use an `Option`, and document on Windows that `None` indicates WGL, and on X11 that `None` indicates GLX (in which case, the `XDisplay` can be retrieved via [`WindowExt::get_xlib_display`](https://github.com/tomaka/winit/blob/aabf0e13b78e156aac3037c9051f49534195e489/src/os/unix.rs#L99); also including a note that Wayland always uses EGL).

ceyusa:
Thanks a lot for the review! :)

> My recommendation is to just create `WindowExt` methods (probably `get_egl_display`) for the platforms that can have an `EGLDisplay`.

If I understand correctly, this method should be in in `winit`, not int `glutin`. Do I understand correctly?

ceyusa:
> Do I understand correctly?

Nope. I did not :) there's no EGL code in winit...

ceyusa:
Sorry, I'm lost. I'm staring at the code and the only way that I see is creating a `GlContextExt` trait with the `get_egl_display()` method, implemented by the platform's GlContext.

Otherwise, I should add the method in `winit`'s trait and implement them in `glutin`, which looks odd to me.

francesca64:
Right, sorry. I think `GlContextExt` makes sense for this.

ceyusa:
\o/ all checks passed

@francesca64 whenever you can :)

francesca64:
Sorry for the delay! This looks good to me, but it would be great if you could add a CHANGELOG entry. Then we'll be good to merge.

ceyusa:
> Sorry for the delay! This looks good to me, but it would be great if you could add a CHANGELOG entry. Then we'll be good to merge.

Done :)

PR #1092, opened 2018-12-20
Unify Context new and new_shared functions, add ContextBuilder::buid method, and clarify docs

### LINKED ISSUE #1089: ContextBuilder missing a build() method
The `ContextBuilder` struct, unlike the `WindowBuilder` struct, does not have a `build()` method. It should, with a signature of `fn build(el: &EventsLoop, shareable_with_windowed_contexts: bool) -> Result<Self, CreationError>`.

### LINKED ISSUE #1091: Unnecessary API Restrictions on Context::new_shared
Right now, we have a separation between `Context::new` and `Context::new_shared` for safety reasons. `Context::new_shared` can apparently cause safety issues, with the docs describing two scenarios in which unsafe behavior can occur (`shareable_with_windowed_contexts` and platform-specific API misuse). However, looking at the documentation there doesn't seem to be anything inherent to creating a context with shared lists that's unsafe.

* Unsafe behavior is allowed if `shareable_with_windowed_contexts` doesn't match between contexts, but there's no reason we can't keep track of that internally and safely throw an appropriate error.
* The errors described in `README-SHARING.md` aren't related to safety, and either cannot occur or can be handled with an `Err`:
  * The following errors cannot occur since we're managing context handles:
    * > glXCreateNewContext can generate the following errors: GLXBadContext if share list is neither zero nor a valid GLX rendering context; **(GLX)**
    * > • If is neither zero nor a valid context handle, then ERROR_INVALID_OPERATION is generated. **(WGL)**
  * None of the address space issues are valid since Rust processes only exist in a single address space.
    * > In a multi-threaded environment, all threads may share the same virtual address space; however, this capability is not required, and implementations may choose to restrict their address space to be per-thread even in an environment supporting multiple application threads. **(EGL)**
      * The Rust threading and `Send`/`Sync` API only make sense with a shared address space - it shouldn't be necessary to consider this in the context sharing API
  * Any creation errors reported by the OS are handled with `Err`.
  * Context incompatibilities (e.g. trying to share an `OpenGL` and `OpenGL ES` context) should be reported by the OS.

What's more, a *different* API lets you perform context sharing without the `unsafe` tag: `GlWindow::new`, which accepts a `ContextBuilder` with the `shared` attribute. If that's safe, why isn't `Context::new_shared`? `Context::new_shared` is a strictly more useful version of `Context::new` and doesn't justify its `unsafe` tag, so it may make sense to merge those APIs.

@ZeGentzy You seem to have written much of the documentation for context sharing. Is there anything I'm missing here that requires the function be unsafe?

Osspial:
See discussion in #1091.

Fixes #1089, closes #1091.


Osspial:
Oh crap, it does! Can't believe I missed that - it should work now though.

ZeGentzy:
From the discussion:
> Unsafe behavior is allowed if shareable_with_windowed_contexts  doesn't match between contexts, but there's no reason we can't keep  track of that internally and safely throw an appropriate error.  

> Additionally, we might also want to make sure that we don't share contexts of different versions.

Is the code currently doing this? I think it already does the first, but how about the second? (I haven't checked however)

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Unreleased
 
 - Added NetBSD support.
+- Removed `new_shared` function from `Context` and `GlWindow`, in favor of `new`.
```
francesca64:
```suggestion
- **Breaking:** Removed `new_shared` function from `Context` and `GlWindow`, in favor of `new`.
```

PR #1093, opened 2018-12-29
Fix broken `link` in dlopen

varkor:
(Discovered in rust-lang/rust#57139.)


src/api/dlopen.rs
```diff
@@ -6,7 +6,7 @@ use std::os::raw::{c_void, c_char, c_int};
 pub const RTLD_LAZY: c_int = 0x001;
 pub const RTLD_NOW: c_int = 0x002;
 
-#[link="dl"]
+#[link(name="dl")]
```
francesca64:
```suggestion
#[link(name = "dl")]
```

PR #1099, opened 2019-02-05
macos: Fix `Context::is_current` incorrectly returning `false`

yvt:
The title says everything.

This fixes the issue <https://github.com/glium/glium/issues/1721> in the downstream library, which has been caused by an assertion failure due to the incorrectly implemented `Context::is_current`.


ZeGentzy:
Will be merged as a subset of #1101, thanks @yvt.

PR #1100, opened 2019-02-15
compare OpenGL contexts with == instead of isEqual

glowcoil:
I don't entirely understand why, but starting with a recent update to macOS, `is_current` has started returning false even when the OpenGL context is in fact current, tripping [this assert](https://github.com/glium/glium/blob/62d287fc9715ab0e10892da5237b478e5322fcaa/src/context/mod.rs#L647) in Glium (see [this issue](https://github.com/maps4print/azul/issues/73); I experienced it in a project I was working on as well). However, this seemed to fix it.

### DIFF
--- a/src/platform/macos/mod.rs
+++ b/src/platform/macos/mod.rs
@@ -191,12 +191,7 @@ impl Context {
 
             let pool = NSAutoreleasePool::new(nil);
             let current = NSOpenGLContext::currentContext(nil);
-            let res = if current != nil {
-                let is_equal: BOOL = msg_send![current, isEqual: context];
-                is_equal != NO
-            } else {
-                false
-            };
+            let res = current == **context;
             let _: () = msg_send![pool, release];
             res
         }


ZeGentzy:
Hey, does this solve the same problem as #1099? If so, which PR do you and @yvt think is better?

yvt:
Both PRs fix the issue in the same way except for the difference in the ways of checking equality — pointer comparison (this PR) or `- isEqual:` (#1099 and `master`).

I think they make no difference in 99.99% of use cases, but might break in the other rare cases. Pointer comparison is in general more fragile against changes in the internal implementation. For example, since `NSOpenGLContext` is a wrapper of a low-level object, it's possible that `currentContext` could return a different object for the same underlying object.

Imagine the case where you have a `NSOpenGLContext` and for some reason you activate it by directly calling `CGLSetCurrentContext`. Now, what would `currentContext` be supposed to do? (a) It could maintain a global hash table that stores mappings to the original `NSOpenGLContext`. But then it would require a process-global critical section. Furthermore, it would still have to return `nil` for contexts not created via `NSOpenGLContext`, thus deviating from the documented behavior. (b) Or, it could simply construct a new `NSOpenGLContext` pointing the same `CGLContextObject`, which behaves exactly the same as the original object, except for its pointer identity.

glowcoil:
Yeah, I'll concur that @yvt's is the more correct of the two.

ZeGentzy:
#1099 has been merged into #1058 in favor of this PR, as you both seam to agree it's the more correct of the two. Thanks folks!

PR #1107, opened 2019-03-02
Update code example in README.md

wschella:
The example was not representative of the current API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,51 +38,44 @@ gl = "*"
 extern crate gl;
 extern crate glutin;
 
-use glutin::dpi::*;
-use glutin::ContextTrait;
+use glutin::dpi;
+use glutin::GlContext;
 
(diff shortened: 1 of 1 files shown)


ZeGentzy:
This example is representative of the upcoming 0.20 api, however I do understand the confusion, so I'll merge this now and revert it before release. Thanks.

wschella:
Ah I understand my troubles now. No need to keep this in then, it might be confusing for those building from git now :wink:

ZeGentzy:
@wschella Is this non-confusing? https://github.com/tomaka/glutin/blob/068fb05f3f195358c0e0a1de0a494448bd37738f/README.md

wschella:
@ZeGentzy Ah yes! A sweet little reminder that GitHub doesn't reflect the version people are probably using.

PR #1108, opened 2019-03-02
Use the shared_library dependency only on required platforms

iceiix:
`shared_library` was added in https://github.com/tomaka/glutin/pull/403, but it is only used in src/api/caca/ffi.rs. 

The caca backend module is [conditionally-compiled](https://github.com/tomaka/glutin/blob/master/src/api/caca/mod.rs#L1-L7) to only certain platforms, so this pull request conditionally compiles the dependency as well, to the same set of target_os's.

This change saves some compilation time (but my specific use case is compiling glutin for the wasm32-unknown-unknown target, where shared_library isn't available)


ZeGentzy:
Please make this compile. `lazy_static` is also used in the egl backend w/ android. 

https://github.com/tomaka/glutin/blob/23b3b101e554e521f38c1179f90d3d2f278b857c/src/api/egl/mod.rs#L1

iceiix:
Fixing Android, next to fix is merge conflicts...

PR #1126, opened 2019-03-20
 Added headless guards to EGL and GLX. Gets rid of libc dependency. Update changelog, fixed multiwindow example. 

ZeGentzy:


glutin/src/api/egl/mod.rs
```diff
@@ -565,40 +571,89 @@ impl Drop for Context {
...
+            // ourselves, but what if some other library is using the display.
+            //
+            // On Linux, we could preload a little lib that does ref counting on
+            // that level, but:
+            //      A) What about other platforms?
+            //      B) Do you *really* want all glutin programs to preload a
+            //      library?
+            //      C) Who the hell is going to maintain that?
+            //
+            // egl.Terminate(self.display);
```
ZeGentzy: Hey @vberger, can you take a look at this and make sure I'm not glossing over something obvious?

One of the many ways you can reproduce is, with master and in weston, trying to close two of the three windows opened by the multiwindow example.

ZeGentzy: This works with weston, but is still broken with X11 (different events loops give different native displays with Wayland but not X11). https://termbin.com/dyv4

ZeGentzy: This little patch is needed to make sure X11 uses EGL:
```
diff --git a/glutin/src/api/glx/mod.rs b/glutin/src/api/glx/mod.rs
index 731891f..64ca4ca 100644
--- a/glutin/src/api/glx/mod.rs
+++ b/glutin/src/api/glx/mod.rs
@@ -64,7 +64,7 @@ use std::sync::Arc;
 use std::{mem, ptr, slice};
 
 lazy_static! {
-    pub static ref GLX: Option<Glx> = Glx::new().ok();
+    pub static ref GLX: Option<Glx> = { Glx::new().ok(); None };
 }
 
 #[derive(Debug)]
```

Issue #784, opened 2016-06-09
Moving contexts between threads.


jwilm:
I've been trying to find a way with Glutin to separate input polling/waiting from the render thread. With GLFW, there is a way to [use the render context on a separate thread](https://github.com/PistonDevelopers/glfw-rs/blob/master/examples/render_task.rs) which would achieve  my goal. Is this possible with current Glutin APIs?

tomaka:
Glutin also provides `make_current`.
What's lacking is the possibility to call `make_current(null)` at the end, but you don't need it unless you want to change your rendering thread from time to time.

jwilm:
Is Window actually `Send` + `Sync`? The methods are all `&self`, and I don't get any errors sending to another thread or using it in parallel from two threads. It is a bit surprising since that's been uncommon in these sorts of libraries historically, and the fact that `WindowProxy` exists also suggests that shouldn't be true of `Window`.

tomaka:
Window can be Send+Sync on Windows (the OS) and Linux, but not OSX.

I don't know OSX well enough and I have yet to get an answer as to whether it's possible to fix this in a hacky way.

fkaa:
On OSX the context can be Send+Sync, not the Window.

jwilm:
I haven't looked at the Mac implementation at all, but missing `make_current(null)` doesn't seem to be a problem there. Must not be glx on that platform? There have been no issues there polling for input from one thread (as long as it's the main thread) and using the context from a secondary thread.

On Linux with X and glx, `make_current(null)` or `clear_context()` would be nice to have.

ZeGentzy:
Merged with #1129

Issue #501, opened 2015-06-25
Add a way to recreate a context with the same window


tomaka:
In case of context loss.

tiby312:
This is the problem I am experiencing on android. If you minimize your app, and then come back to it, android may have destroyed the context. It is not guaranteed to have remained.

ZeGentzy:
Afaik, this is achievable via the raw context extensions (even though it is not the most ergonomic thing). Sadly, this extension is only implemented for Windows and Linux. Feel free to contribute implementations of this extension for other platforms.

kchibisov:
Fixed in #1439.

PR #1129, opened 2019-03-27
[WIP] 0.21 Api changes.

ZeGentzy:
Compiles with the following backends:
- [x] Linux | X11 | EGL
- [x] Linux | X11 | GLX
- [x] Linux | Wayland | EGL
- [x] Linux | OSMesa
- [x] Windows | WGL
- [x] Windows | EGL
- [x] Web | Emscripten
- [x] macOS
- [x] iOS
- [x] Android

Has been tested with the following backends:
- [x] Linux | X11 | EGL
- [x] Linux | X11 | GLX
- [x] Linux | Wayland | EGL
- [x] Linux | OSMesa
- [ ] Windows | WGL
- [ ] Windows | EGL
- [ ] Web | Emscripten
- [ ] macOS
- [ ] iOS
- [ ] Android

For #501 and #784


ZeGentzy:
This PR comes with a slew of changes.
 - The removal of libcaca: It is currently impossible for users to use libcaca (afaik), and it seams out of place for glutin.
 - Contexts are now take either `PossiblyCurrentContext` or `NotCurrentContext` as a generic. The goal for this is to eventually implement `Send + Sync` on contexts which are `NotCurrentContext`s, but not on `PossiblyCurrentContext`. Contexts can be converted between the two via `make_not_current` and `make_current`.

Emphasis should be put on the "Possibly", however, because a user can just do this:
```
let c1: Context<PossiblyCurrentContext> = c1.make_current_context().unwrap();
let c2: Context<PossiblyCurrentContext> = c2.make_current_context().unwrap();

c1.get_proc_address("Uh oh, this is undefined.");
```

There is no way to stop this from happening, which is why we also expose `treat_as_not_current`, which in the example above would be called on `c1`  after making `c2` current.

ZeGentzy:
https://gitlab.freedesktop.org/mesa/mesa/merge_requests/533

PR #1132, opened 2019-04-01
Attempt to implement context sharing on Mac

tene:
I have no idea if this is the right thing to do or not, but with this patch I can run the sharing example.

I've never done any cocoa programming at all; I just followed the "Sharing Rendering Context Resources" section of https://developer.apple.com/library/archive/documentation/GraphicsImaging/Conceptual/OpenGL-MacProgGuide/opengl_contexts/opengl_contexts.html and did what seemed obvious, and I was surprised that it actually seemed to work.

I don't have another platform to run the examples on right now to actually check if the sharing example renders identically compared to other platforms.

This will contribute to tomaka/glutin#899

### DIFF
--- a/glutin/src/platform/macos/mod.rs
+++ b/glutin/src/platform/macos/mod.rs
@@ -60,9 +60,16 @@ impl Context {
         let transparent = wb.window.transparent;
         let win = wb.build(el)?;
 
-        if gl_attr.sharing.is_some() {
-            unimplemented!()
-        }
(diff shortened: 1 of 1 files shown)


ZeGentzy:
@tene Does the sharing example render the same as the window example? 

You and @fschutt probably know more about this, so if it looks good to you folks, then it lgtm.

tene:
@ZeGentzy Yes, the sharing example and the window example look identical to me, except for different default window sizes.

PR #1144, opened 2019-04-14
Update x11-dl

atouchet:


ZeGentzy:
You also have to bump the crate version for this to take effect.

Question: Are you sure it is not suing 2.18.3 already?

atouchet:
This is the [version that winit already requires](https://github.com/rust-windowing/winit/blob/master/Cargo.toml#L68) so it is unlikely that anyone would be using a previous version. However there were [issues with using 2.18.2 with new versions of rustc](https://github.com/erlepereira/x11-rs/pull/92).

ZeGentzy:
Ok, update ` glutin_glx_sys` to `0.1.5` and glutin to the next RC.

Issue #1154, opened 2019-05-11
Crash on iOS when building context when not debugging


hamaluik:
I'm trying to get glutin up and running on iOS, but I keep getting crashes / panics when I launch the app manually (i.e., without debugging through XCode). This happens both in the simulator & on a device (iPhone 7+). Strangely, it works just fine when launched through XCode.

Here is the code that is crashing:

https://github.com/hamaluik/rust-ios-graphics/blob/master/rust/src/launch.rs

It seems to only crash when I try to build a context. If I build a window without a context it works fine.

As far as I can tell it's not caused by winit, as this code works just fine: https://github.com/hamaluik/rust-ios-graphics/blob/f656470383b3b7c87c689a8b07d42e56da71dfa9/rust/src/launch.rs

Any ideas?

ZeGentzy:
I am not familiar with iOS, so you'll have to look into it yourself.

(I'm afraid we lack a maintainer for iOS).

hamaluik:
I started going through the source and found links to UIKit (which I already had), but also CoreFoundation and GLKit: https://github.com/rust-windowing/glutin/blob/master/glutin_gles2_sys/src/lib.rs#L82-L84. I was just including the OpenGlES framework in my XCode project, but when I also include CoreFoundation and GLKit, it works! I feel like a bit of a dummy now.

PR #1156, opened 2019-05-11
Add notes about compiling on iOS to README

hamaluik:
re: rust-windowing/glutin#1154

Add notes to the README about getting it compile and run properly on iOS—for now just including the explicitly and implicitly required frameworks.


hamaluik:
Thanks!

PR #1162, opened 2019-05-25
Update parking_lot to 0.8 and wayland-client to 0.23

Eijebong:


atouchet:
This should probably be updated in winit as well.

Eijebong:
Fun, I didn't see any dupe when I updated them here... I'll open a PR in winit, thanks

Eijebong:
> error: aborting due to 59 previous errors

Might be a while though :laughing:

vberger:
The update of wayland-client will come with eventloop-2.0.

Eijebong:
:+1: Oh ok, will not lose my time with it then, thanks !

Eijebong:
Should I remove the bump from that PR then ?

ZeGentzy:
Glutin master is now based against eventloop-2.0, so this is fine :)

Eijebong:
Oohh so that's why I didn't see any dupes, there weren't any :p

PR #1164, opened 2019-06-01
Fix build failure

chrisduerr:
This updates glutin to the latest winit commit to resolve all build
failures which occured because of name changes.

It also locks winit to a fixed revision, so things can be updated in a
more controlled manner.

This should at least fix the build failures on Linux, I'll try to update it
if CI complains about other platforms. Let me know if you'd prefer to
keep the winit version pointing to the branch instead of the commit.

### DIFF
--- a/glutin/Cargo.toml
+++ b/glutin/Cargo.toml
@@ -19,7 +19,7 @@ serde = ["winit/serde"]
 [dependencies]
 lazy_static = "1.3"
 #winit = "^0.19.1"
-winit = { git = "https://github.com/rust-windowing/winit.git", branch = "eventloop-2.0"}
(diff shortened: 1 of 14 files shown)


ZeGentzy:
Ah, silly me, silly me. Anyways, I think you'll also need to update the examples.

chrisduerr:
I think macos/windows/linux should all build now. Is it correct that the other platforms do not work yet @ZeGentzy? It seems they still need some work in the event loop 2 branch of winit.

ZeGentzy:
Yeah. windows/linux/macos all should build with glutin and winit. iOS has been ported in winit, but not in glutin. Android and emscripten still need to be ported in winit. [stdweb has yet to land in winit](https://github.com/rust-windowing/winit/pull/797).

PR #1168, opened 2019-06-11
fix osx panic

blakehawkins:
Deleted code in `mod` causes a panic on osx.

This circumvents the problem.

### DIFF
--- a/glutin_examples/examples/support/mod.rs
+++ b/glutin_examples/examples/support/mod.rs
@@ -11,34 +11,9 @@ pub struct Gl {
     pub gl: gl::Gl,
 }
 
-extern "system" fn dbg_callback(
-    source: gl::types::GLenum,
-    etype: gl::types::GLenum,
-    _id: gl::types::GLuint,
-    severity: gl::types::GLenum,
-    _msg_length: gl::types::GLsizei,
-    msg: *const gl::types::GLchar,
-    _user_data: *mut std::ffi::c_void,
-) {
-    unsafe {
-        println!(
-            "dbg_callback {:#X} {:#X} {:#X} {:?}",
-            source,
-            etype,
-            severity,
-            std::ffi::CStr::from_ptr(msg),
-        );
-    }
-}
-
 pub fn load(gl_context: &glutin::Context<PossiblyCurrent>) -> Gl {
     let gl =
         gl::Gl::load_with(|ptr| gl_context.get_proc_address(ptr) as *const _);
-
-    unsafe {
-        gl.Enable(gl::DEBUG_OUTPUT);
-        gl.DebugMessageCallback(dbg_callback, std::ptr::null());
-    }
 
     let version = unsafe {
         let data = CStr::from_ptr(gl.GetString(gl::VERSION) as *const _)
--- a/glutin_examples/examples/window.rs
+++ b/glutin_examples/examples/window.rs
@@ -5,20 +5,21 @@ use glutin::event_loop::{ControlFlow, EventLoop};
 use glutin::window::WindowBuilder;
 use glutin::ContextBuilder;
 
-fn main() {
+fn main() -> Result<(), glutin::CreationError> {
     let el = EventLoop::new();
     let wb = WindowBuilder::new().with_title("A fantastic window!");
 
     let windowed_context =
-        ContextBuilder::new().build_windowed(wb, &el).unwrap();
+        ContextBuilder::new().build_windowed(wb, &el)?;
 
-    let windowed_context = unsafe { windowed_context.make_current().unwrap() };
+    let windowed_context = unsafe {
+       windowed_context.make_current().expect("Make current fail")
+    };
 
     println!(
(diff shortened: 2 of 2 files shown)


glutin_examples/examples/support/mod.rs
```diff
@@ -11,35 +11,10 @@ pub struct Gl {
...
-    }
-}
-
 pub fn load(gl_context: &glutin::Context<PossiblyCurrent>) -> Gl {
     let gl =
         gl::Gl::load_with(|ptr| gl_context.get_proc_address(ptr) as *const _);
 
-    unsafe {
-        gl.Enable(gl::DEBUG_OUTPUT);
-        gl.DebugMessageCallback(dbg_callback, std::ptr::null());
```
ZeGentzy: Instead of removing this code, can we instead wrap it with a `#[cfg(not(target_os = "macos"))]`? Why is this code causing it to panic? How would it even?

glutin_examples/examples/window.rs
```diff
@@ -5,20 +5,21 @@ use glutin::event_loop::{ControlFlow, EventLoop};
...
     let el = EventLoop::new();
     let wb = WindowBuilder::new().with_title("A fantastic window!");
 
     let windowed_context =
-        ContextBuilder::new().build_windowed(wb, &el).unwrap();
+        ContextBuilder::new().build_windowed(wb, &el)?;
 
-    let windowed_context = unsafe { windowed_context.make_current().unwrap() };
+    let windowed_context = unsafe {
+       windowed_context.make_current().expect("Make current fail")
```
ZeGentzy: This `expect` should stay as an `unwrap`, imho.

glutin_examples/examples/window.rs
```diff
@@ -36,9 +37,10 @@ fn main() {
                 }
                 WindowEvent::RedrawRequested => {
                     gl.draw_frame([1.0, 0.5, 0.7, 1.0]);
-                    windowed_context.swap_buffers().unwrap();
+                    windowed_context.swap_buffers().expect("Swapbuffer fail");
```
ZeGentzy: This `expect` should stay as an `unwrap`, imho.

glutin_examples/examples/window.rs
```diff
@@ -5,20 +5,21 @@ use glutin::event_loop::{ControlFlow, EventLoop};
...
 use glutin::ContextBuilder;
 
-fn main() {
+fn main() -> Result<(), glutin::CreationError> {
     let el = EventLoop::new();
     let wb = WindowBuilder::new().with_title("A fantastic window!");
 
     let windowed_context =
-        ContextBuilder::new().build_windowed(wb, &el).unwrap();
+        ContextBuilder::new().build_windowed(wb, &el)?;
```
ZeGentzy: If creating the context fails, then this example *should* fail.

ZeGentzy:
Could you explain why you are getting panics, and how your changes fix them?

blakehawkins:
Hey @ZeGentzy

Apologies for ignoring you. I was doing a hacking evening with glutin and needed to make this change for it to work with my (up-to-date) macbook pro 15.

Realistically, I won't be able to follow-through with this PR, but hopefully it will be a good reference if the issue comes up again.

I don't recall what the error specifically was, but it basically amounted to DebugMessageCallback not working on my platform (OSX, launched from native terminal).

Sorry again for ignoring your questions until now.

Cheers,

glutin_examples/examples/support/mod.rs
blakehawkins: I guess the answer to the question depends on whether you want to maintain both code branches or not.

My view (with extremely limited experience) is that those debug lines provide pretty much no value, so removing them seems sensible.

Also, given that this is a reference example, keeping some code that circumvents what's ultimately a bug that you're not responsible for, seems a bit painful.

glutin_examples/examples/window.rs
blakehawkins: I disagree, `expect` is pretty much exclusively better than `unwrap`

glutin_examples/examples/window.rs
blakehawkins: I don't understand this comment. The `?` will definitely fail the example by returning a `Result::Err` variant

glutin_examples/examples/window.rs
ZeGentzy: And I disagree with your disagreement :)

`expect` is only better than a `unwrap` if the error message provides something that the `unwrap` doesn't already.

glutin_examples/examples/window.rs
ZeGentzy: Sorry, when I said fail, I meant panic. It would be inconsistent if all the other parts of the code panicked, and this one was returning an error.

glutin_examples/examples/support/mod.rs
ZeGentzy: Ok sure. If you want these ~20 lines removed, I guess it makes sure.

PR #1171, opened 2019-06-20
Update Glutin build to latest Winit master and make it compile

Osspial:
The build on Windows was failing because of the late `with_visibility`->`with_visible` rename. This fixes that.

### DIFF
--- a/glutin/Cargo.toml
+++ b/glutin/Cargo.toml
@@ -19,7 +19,7 @@ serde = ["winit/serde"]
 [dependencies]
 lazy_static = "1.3"
 #winit = "^0.19.1"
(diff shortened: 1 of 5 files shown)


ZeGentzy:
Needs `with_transparent` too.

ZeGentzy:
You guys changed nsview?
```
error[E0599]: no method named `nsview` found for type `winit::window::Window` in the current scope
  --> glutin/src/platform_impl/macos/mod.rs:72:24
   |
72 |         let view = win.nsview() as id;
   |                        ^^^^^^
```

Osspial:
Yeah, we changed the `ns{whatever}` functions on macOS to `ns_{whatever}`. We got a PR making that change short after I did all the function name changes, and I merged it since using snake_case there is less surprising than having it be all one word.

PR #1179, opened 2019-07-01
Add better attribute handling for sRGB in WGL

sumit0190:
Based on a discussion in #1175 , this PR changes `choose_pixel_format_id` to ensure that the right extension for SRGB is always added to the descriptor, and its value is set according to the `with_srgb` parameter supplied by the user.

Tested on Windows 10 x64; no warnings/errors reported.

### DIFF
--- a/glutin/src/api/wgl/mod.rs
+++ b/glutin/src/api/wgl/mod.rs
@@ -701,26 +701,28 @@ unsafe fn choose_arb_pixel_format_id(
         out.push(gl::wgl_extra::STEREO_ARB as raw::c_int);
         out.push(if pf_reqs.stereoscopy { 1 } else { 0 });
 
-        if pf_reqs.srgb {
-            if extensions
-                .split(' ')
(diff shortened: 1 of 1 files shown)


glutin/src/api/wgl/mod.rs
```diff
@@ -701,26 +701,28 @@ unsafe fn choose_arb_pixel_format_id(
...
+        } else if extensions
+            .split(' ')
+            .find(|&i| i == "WGL_EXT_framebuffer_sRGB")
+            .is_some()
+        {
+            out.push(
+                gl::wgl_extra::FRAMEBUFFER_SRGB_CAPABLE_EXT as raw::c_int,
+            );
+            out.push(pf_reqs.srgb as raw::c_int);
+        } else {
```
ZeGentzy: Maybe `} else if pf_reqs.srgb {` instead of `} else {` followed by `if`...

Should've caught that when I wrote it the first time, sorry.

Issue #1170, opened 2019-06-18
Port iOS backend to Eventsloop v2


ZeGentzy:

rukai:
Hey, just a heads up, this is the main issue blocking https://github.com/gfx-rs/gfx/pull/2888

PR #1192, opened 2019-07-16
Get iOS compiling again

rukai:
Seeing as there is noone else around with iOS to work on https://github.com/rust-windowing/glutin/issues/1170 and its blocking gfx from using winit 0.20 and I dont have access to iOS

Its time for everyones favorite development technique.
...
Throwing code at CI until it compiles.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,13 +36,13 @@ matrix:
       os: osx
       rust: stable
 
-    # # iOS
-    # - env: TARGET=x86_64-apple-ios
-    #   os: osx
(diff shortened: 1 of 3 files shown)


ZeGentzy:
I'd try to get `cargo check --target x86_64-apple-ios` to work. Better than waiting for CI.

rukai:
Ooh, can I do that on Linux? Regardless iOS is compiling now.

ZeGentzy:
That's how I got macos functional :)

Issue #1200, opened 2019-08-09
“gles2 function was not loaded on” on macOS since v0.21.0


tonsky:
I’m on macOS 10.14.6 with MacBook Pro (15-inch, 2018)
Same happens on master. v0.20.0 and before runs just fine
Thanks!

```
[~/work/glutin] git checkout v0.21.0
[~/work/glutin] RUST_BACKTRACE=1 cargo run --example window
Pixel format of the window's GL context: PixelFormat { hardware_accelerated: true, color_bits: 24, alpha_bits: 8, depth_bits: 24, stencil_bits: 8, stereoscopy: false, double_buffer: true, multisampling: None, srgb: true }
thread 'main' panicked at 'gles2 function was not loaded', /Users/nikita.prokopov/work/glutin/target/debug/build/glutin_examples-ea04917525f54098/out/gl_bindings.rs:1158:13
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:59
             at src/libstd/panicking.rs:197
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:211
   4: <std::panicking::begin_panic::PanicPayload<A> as core::panic::BoxMeUp>::get
             at src/libstd/panicking.rs:474
   5: <alloc::collections::CollectionAllocErr as core::convert::From<core::alloc::AllocErr>>::from
             at /rustc/3c235d5600393dfe6c36eeed34042efad8d4f26e/src/libstd/panicking.rs:408
   6: window::support::gl::missing_fn_panic
             at ./target/debug/build/glutin_examples-ea04917525f54098/out/gl_bindings.rs:1158
   7: window::support::gl::Gles2::DebugMessageCallback
             at ./target/debug/build/glutin_examples-ea04917525f54098/out/gl_bindings.rs:2269
   8: window::support::load
             at glutin_examples/examples/support/mod.rs:40
   9: window::main
             at glutin_examples/examples/window.rs:18
  10: std::rt::lang_start::{{closure}}
             at /rustc/3c235d5600393dfe6c36eeed34042efad8d4f26e/src/libstd/rt.rs:64
  11: std::panicking::try::do_call
[...]

PR #1201, opened 2019-08-09
Fix examples on macos.

ZeGentzy:
Does this fix your issue in #1200, @tonsky?


tonsky:
Yes it does. Thank you!

PR #1203, opened 2019-08-10
Set VSync in EGL

### LINKED ISSUE #1202: EGL backend doesn't actually disable VSync
I'm making a Wayland app with a raw context (using `smithay-client-toolkit` directly rather than `winit`). I noticed that VSync ends up enabled regardless of the `ContextBuilder` setting (so `swap_buffers` waits for the frame callback, which I don't want). I looked through the EGL code and nothing actually calls `egl.SwapInterval(display, 0)` there to disable VSync (it defaults to enabled, see Notes [here](https://www.khronos.org/registry/EGL/sdk/docs/man/html/eglSwapInterval.xhtml)). I inserted this call into `make_current()` (since it operates on the current context) and that did indeed disable VSync (`swap_buffers` no longer waits for the frame callback).

YaLTeR:
Fixes #1202. Tested on Wayland.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Fixed attribute handling for sRGB in WGL.
+- Fixed VSync being always enabled on EGL.
 
 # Version 0.22.0-alpha1 (2019-06-21)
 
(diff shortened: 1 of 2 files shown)


ZeGentzy:
lgtm, ty

PR #1207, opened 2019-08-16
Damaged buffer swap

kennylevinsen:
Implement damaged buffer swap for EGL.

This allows compositors to do significantly less work, especially for very large window buffers with small changes (think blinking cursor in a large window on a 4k display). It can theoretically also be used to aid screencopy, such as remote desktop-like scenarios (waypipe, rdp, vnc, ...).

On Wayland, clients are expected to always report correct damage, so this allows for glutin consumers to follow suit.

A version of this on the backport21 branch is used by this WIP PR for alacritty: https://github.com/jwilm/alacritty/pull/2724.

Outstanding:
- Should we communicate damage support separately, or just let clients try and see?
- Only EGL is implemented, and only Wayland has been tested. Do we have other platforms that also support damage that we wish to integrate?

### DIFF
--- a/glutin/src/api/egl/mod.rs
+++ b/glutin/src/api/egl/mod.rs
@@ -13,6 +13,7 @@
 mod egl {
     use super::ffi;
     use crate::api::dlloader::{SymTrait, SymWrapper};
+    use libloading;
 
     #[derive(Clone)]
     pub struct Egl(pub SymWrapper<ffi::egl::Egl>);
@@ -21,11 +22,55 @@ mod egl {
     unsafe impl Sync for Egl {}
 
     impl SymTrait for ffi::egl::Egl {
-        fn load_with<F>(loadfn: F) -> Self
+        fn load_with<F>(lib: &libloading::Library, _: F) -> Self
         where
             F: FnMut(&'static str) -> *const std::os::raw::c_void,
         {
-            Self::load_with(loadfn)
+            let f = move |s: &'static str| -> *const std::os::raw::c_void {
+                // Check if the symbol is available in the library directly. If
+                // it is, just return it.
+                match unsafe {
+                    lib.get(
+                        std::ffi::CString::new(s.as_bytes())
+                            .unwrap()
+                            .as_bytes_with_nul(),
+                    )
+                } {
+                    Ok(sym) => return *sym,
+                    Err(_) => (),
+                };
+
+                // The symbol was not available in the library, so ask
+                // eglGetProcAddress for it. Note that eglGetProcAddress was
+                // only able to look up extension functions prior to EGL 1.5,
+                // hence this two-part dance.
+                match unsafe { lib.get(b"eglGetProcAddress") } {
+                    Ok(sym) => {
+                        let sym: libloading::Symbol<
+                            unsafe extern "C" fn(
+                                *const std::os::raw::c_void,
+                            )
+                                -> *const std::os::raw::c_void,
+                        > = sym;
+                        let ptr = unsafe {
+                            sym(std::ffi::CString::new(s.as_bytes())
+                                .unwrap()
+                                .as_bytes_with_nul()
+                                .as_ptr()
+                                as *const std::os::raw::c_void)
+                        };
+                        if ptr.is_null() {
+                            panic!("unable to load symbol {:}", s)
+                        }
+                        return ptr;
+                    }
+                    Err(_) => (),
+                };
+
+                std::ptr::null()
+            };
+
+            Self::load_with(f)
         }
     }
 
--- a/glutin/src/api/egl/mod.rs
+++ b/glutin/src/api/egl/mod.rs
@@ -14,12 +14,29 @@ mod egl {
     use super::ffi;
     use crate::api::dlloader::{SymTrait, SymWrapper};
     use libloading;
+    use std::sync::{Arc, Mutex};
+
+    #[cfg(unix)]
+    use libloading::os::unix as libloading_os;
+    #[cfg(windows)]
+    use libloading::os::windows as libloading_os;
 
     #[derive(Clone)]
     pub struct Egl(pub SymWrapper<ffi::egl::Egl>);
 
     /// Because `*const raw::c_void` doesn't implement `Sync`.
     unsafe impl Sync for Egl {}
+
+    type EglGetProcAddressType = libloading_os::Symbol<
+        unsafe extern "C" fn(
+            *const std::os::raw::c_void,
+        ) -> *const std::os::raw::c_void,
+    >;
+
+    lazy_static! {
+        static ref EGL_GET_PROC_ADDRESS: Arc<Mutex<Option<EglGetProcAddressType>>> =
+            Arc::new(Mutex::new(None));
+    }
 
     impl SymTrait for ffi::egl::Egl {
         fn load_with(lib: &libloading::Library) -> Self {
@@ -37,27 +54,32 @@ mod egl {
                     Err(_) => (),
                 };
 
-                // The symbol was not available in the library, so ask
-                // eglGetProcAddress for it. Note that eglGetProcAddress was
-                // only able to look up extension functions prior to EGL 1.5,
-                // hence this two-part dance.
-                match unsafe { lib.get(b"eglGetProcAddress") } {
-                    Ok(sym) => {
+                let mut egl_get_proc_address =
+                    (*EGL_GET_PROC_ADDRESS).lock().unwrap();
+                if egl_get_proc_address.is_none() {
+                    unsafe {
                         let sym: libloading::Symbol<
                             unsafe extern "C" fn(
                                 *const std::os::raw::c_void,
                             )
                                 -> *const std::os::raw::c_void,
-                        > = sym;
-                        unsafe {
-                            sym(std::ffi::CString::new(s.as_bytes())
-                                .unwrap()
-                                .as_bytes_with_nul()
-                                .as_ptr()
-                                as *const std::os::raw::c_void)
-                        }
+                        > = lib.get(b"eglGetProcAddress").unwrap();
+                        *egl_get_proc_address = Some(sym.into_raw());
                     }
-                    Err(_) => std::ptr::null(),
+                }
+
+                // The symbol was not available in the library, so ask
+                // eglGetProcAddress for it. Note that eglGetProcAddress was
+                // only able to look up extension functions prior to EGL 1.5,
+                // hence this two-part dance.
+                unsafe {
+                    (egl_get_proc_address.as_ref().unwrap())(
+                        std::ffi::CString::new(s.as_bytes())
+                            .unwrap()
+                            .as_bytes_with_nul()
+                            .as_ptr()
+                            as *const std::os::raw::c_void,
+                    )
                 }
             };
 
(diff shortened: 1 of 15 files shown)


glutin/src/api/egl/mod.rs
```diff
@@ -626,6 +671,51 @@ impl Context {
...
+
+        for rect in rects {
+            ffirects.push(rect.x as ffi::egl::types::EGLint);
+            ffirects.push(rect.y as ffi::egl::types::EGLint);
+            ffirects.push(rect.width as ffi::egl::types::EGLint);
+            ffirects.push(rect.height as ffi::egl::types::EGLint);
+        }
+
+        let ret = unsafe {
+            egl.SwapBuffersWithDamageKHR(
```
ZeGentzy: We should check that the extension is present, and error if it is not.

glutin/src/api/egl/mod.rs
```diff
@@ -626,6 +671,51 @@ impl Context {
...
+        rects: &[Rect],
+    ) -> Result<(), ContextError> {
+        let egl = EGL.as_ref().unwrap();
+        let surface = self.surface.as_ref().unwrap().lock();
+        if *surface == ffi::egl::NO_SURFACE {
+            return Err(ContextError::ContextLost);
+        }
+
+        let mut ffirects: Vec<ffi::egl::types::EGLint> =
+            Vec::with_capacity(rects.len());
```
ZeGentzy: Needs to be with capacity of `rects.len() * r`

glutin/src/api/egl/mod.rs
```diff
@@ -626,6 +671,51 @@ impl Context {
...
+        let ret = unsafe {
+            egl.SwapBuffersWithDamageKHR(
+                self.display,
+                *surface,
+                ffirects.as_mut_ptr(),
+                rects.len() as ffi::egl::types::EGLint,
+            )
+        };
+
+        if ret == 0 {
```
ZeGentzy: `ret == ffi::egl::FALSE`

glutin/src/lib.rs
```diff
@@ -759,3 +761,13 @@ impl<S> Default for GlAttributes<S> {
         }
     }
 }
+
+// Rectangles to submit as buffer damage.
+#[allow(dead_code)]
+#[derive(Debug)]
```
ZeGentzy: I think we should also implement things like `Clone`, `PartialEq`, `Eq`, and so on.

ZeGentzy:
I'm not sure exactly how it would work, but for X11 + glx/egl we should also support X DAMAGE.

https://www.x.org/releases/X11R7.7/doc/damageproto/damageproto.txt

cc @murarth

kennylevinsen:
For X11 EGL, it should already work (I guess?). I don't know anything about X DAMAGE, but the more platforms the merrier, as long as we can keep it simple to use (and won't be too much work to implement). Easier to sell the idea of implementing damage tracking when more platforms benefit. :)

On that note, does anyone know if there is something similar for Windows and macOS? I couldn't immediately find anything, but am also not at all familiar with their APIs.

Regarding CI failures: I'll add the appropriate dummy error implementations for the various unsupported platforms so that builds will pass soon.

ZeGentzy:
Can we get a example that demonstrates damage? Maybe something that draws a random color everytime the user presses enter, but only swaps a random sub-portion of the window. Just something that would only work correctly if we have damage implemented correctly, just so we have something to test against.

kennylevinsen:
I'm going to make your suggested example, but I'd just point out that relying on *only* damage to be displayed is undefined behavior, as the compositor might need to read more than the damage. Unhiding a window, for example, likely requires a full draw of the buffer. EGL_KHR_partial_update, which provides buffer age information, is needed for partial buffer draws.

I don't think it's a problem at all, though, it should just be noted. For correctness testing, I run a nested Sway instance with damage highlighting enabled, which masks out undamaged areas to make damage clearly visible, but that's a bit out of scope for embedded glutin tests.

kennylevinsen:
I added a quick example. Color changes when the mouse moves over the window, but only the lower left 100x100px are damaged. However, again, do note that not damaging all changed content results in undefined behavior, as the compositor is expects the rest of the buffer to be unchanged.

I'm going to try to get the other platforms to pass, but I might have to brute-force the CI a little bit to get there.

Do you have any opinion on whether or not there should be a "damage_supported" method or similar to probe for availability? I'm leaning towards "yes".

kennylevinsen:
Off-topic: CI takes *30 minutes*! One of the builds alone have a 28 minute runtime. *On linux*. Dang.

kennylevinsen:
Ready for review. Travis failure is a single case of nightly hanging, and circleci also fails on master.

glutin/src/api/egl/mod.rs
```diff
@@ -21,11 +22,55 @@ mod egl {
     unsafe impl Sync for Egl {}
 
     impl SymTrait for ffi::egl::Egl {
-        fn load_with<F>(loadfn: F) -> Self
+        fn load_with<F>(lib: &libloading::Library, _: F) -> Self
```
ZeGentzy: I'd rather we change `SymTrait` to not pass a fn, then ignore the fn passed.

glutin/src/api/egl/mod.rs
```diff
@@ -21,11 +22,55 @@ mod egl {
...
+                        > = sym;
+                        let ptr = unsafe {
+                            sym(std::ffi::CString::new(s.as_bytes())
+                                .unwrap()
+                                .as_bytes_with_nul()
+                                .as_ptr()
+                                as *const std::os::raw::c_void)
+                        };
+                        if ptr.is_null() {
+                            panic!("unable to load symbol {:}", s)
```
ZeGentzy: What if the proc address is not there cause it is not supported?

glutin/src/api/egl/mod.rs
```diff
@@ -21,11 +22,55 @@ mod egl {
...
+                } {
+                    Ok(sym) => return *sym,
+                    Err(_) => (),
+                };
+
+                // The symbol was not available in the library, so ask
+                // eglGetProcAddress for it. Note that eglGetProcAddress was
+                // only able to look up extension functions prior to EGL 1.5,
+                // hence this two-part dance.
+                match unsafe { lib.get(b"eglGetProcAddress") } {
```
ZeGentzy: Can we please cache this?

glutin/src/lib.rs
```diff
@@ -750,3 +752,13 @@ impl<S> Default for GlAttributes<S> {
         }
     }
 }
+
+// Rectangles to submit as buffer damage.
+#[allow(dead_code)]
```
ZeGentzy: Hold up, why is this dead_code?

glutin/src/windowed.rs
```diff
@@ -136,6 +136,27 @@ impl<W> ContextWrapper<PossiblyCurrent, W> {
...
+    /// override your vsync settings, which means that you can't know in
+    /// advance whether `swap_buffers` will block or not.
+    pub fn swap_buffers_with_damage(
+        &self,
+        rects: &[Rect],
+    ) -> Result<(), ContextError> {
+        self.context.context.swap_buffers_with_damage(rects)
+    }
+
+    pub fn swap_buffers_with_damage_supported(&self) -> bool {
```
ZeGentzy: needs docs.

glutin_examples/examples/damage.rs
```diff
@@ -0,0 +1,105 @@
...
+    let el = EventLoop::new();
+    let wb = WindowBuilder::new().with_title("A fantastic window!");
+
+    let windowed_context =
+        ContextBuilder::new().build_windowed(wb, &el).unwrap();
+
+    let windowed_context = unsafe { windowed_context.make_current().unwrap() };
+
+    if !windowed_context.swap_buffers_with_damage_supported() {
+        panic!("Damage not supported, this demo will instead perform full buffer swaps");
```
ZeGentzy: No it doesn't. It panics.

glutin_examples/examples/damage.rs
```diff
@@ -0,0 +1,105 @@
...
+                    // composited to screen.
+                    //
+                    // Panics if damage is not supported due to the unwrap.
+                    color = color.next();
+                    gl.draw_frame([color.red, color.green, color.blue, 1.0]);
+                    if windowed_context.swap_buffers_with_damage_supported() {
+                        windowed_context.swap_buffers_with_damage(&[Rect{
+                            x: 0,
+                            y: 0,
+                            height: 100,
```
ZeGentzy: What if the window is smaller then 100x100 px?

glutin/src/lib.rs
```diff
@@ -750,3 +752,13 @@ impl<S> Default for GlAttributes<S> {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Rect {
+    pub x: u32,
+    pub y: u32,
```
ZeGentzy: Can we switch this to use winit's `PhysicalSize` and physical position?

glutin/src/api/egl/mod.rs
kennylevinsen: Done

glutin/src/api/egl/mod.rs
kennylevinsen: Woops, fixed.

glutin/src/lib.rs
kennylevinsen: Eh, it's not. Sorry.

glutin/src/windowed.rs
kennylevinsen: Done

glutin_examples/examples/damage.rs
kennylevinsen: Hmm, the docs doesn't seem to mention if damage larger than the surface or buffer is an error. Sway accepts it, and just damages the full surface.

If you want, I can add the logic to handle resizes and cap damage to the current buffer size, but it doesn't appear to be necessary.

glutin/src/lib.rs
kennylevinsen: Is the winit PhysicalPosition normally used for GL coordinates (origin lower left)? If so, I guess there should be no problem. Not super happy about using floats, but oh well.

glutin/src/api/egl/mod.rs
kennylevinsen: Added cache in a global. I'd have preferred to keep it in the Egl struct, but I don't think that would work with egl_bindings.

glutin/src/lib.rs
ZeGentzy: Hmmm, apparently `PhysicalSize` is getting changed, but `PhysicalPosition` isn't.... I'd bring up this inconsistency here so it gets "fixed": https://github.com/rust-windowing/winit/issues/939

kennylevinsen: Inconsistency brought up. I can convert Rect to contain a PhysicalSize/PhysicalPosition, although people might be surprised that PhysicalPosition is used with origin bottom left, when it seems to usually be used for origin top left.

ZeGentzy:
Okay, so I'd hate if this PR sits here for eternity over `PhysicalSize`/`PhysicalPosition` shenangians. I'm merging it as is. Thank you very much @kennylevinsen for this PR.

kennylevinsen:
Thanks!

PR #1221, opened 2019-09-27
MacOSX's raw_handle trait method returns CGLContext object

ceyusa:
Righ now MacOS' raw_handler() from ContextTraitExt returns NSOpenGLContext object, from cocoa-rs appkit. It's address is not useful for sharing the real GL context address, which is a CGLContextObj.

This patch propose to return the return value of CGLContextObj()


ceyusa:
I wonder if it is possible to backport this patch to 0.21

ceyusa:
The CI errors seems unrelated with this pull request.

ZeGentzy:
Thanks! merging. If you want it on 0.21, there's the branch `backport21`.

PR #1224, opened 2019-10-12
Fix not linking against OpenGLES.framework for iOS

aleksijuvani:


glutin_gles2_sys/build.rs
```diff
@@ -8,6 +8,8 @@ fn main() {
     let dest = PathBuf::from(&env::var("OUT_DIR").unwrap());
 
     println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rustc-link-lib=framework=GLKit");
+    println!("cargo:rustc-link-lib=framework=OpenGLES");
```
goddessfreya: These need to be moved under the `if target.contains` bellow. 

Is this also necessary for macos, or just iOS?

aleksijuvani: I think glutin_gles2_sys is only ever used by the iOS backend, so the if-statement is redundant.

goddessfreya: Yes, you are correct, I guess it is not used with macos.

Anyways, as it stands, CI builds every crate for every platform, and keeping this out of the if statement breaks the CI. 

I imagine it's possible to stop the CI from doing this, but I think the simplest solution would be just to move those two lines in. This would be consistent with all the other *_sys which all no-op if built on a non-supporting platform.

aleksijuvani: I see. I'll make this change in a bit when I get the chance.

PR #1225, opened 2019-10-12
Add Xcode project for iOS examples

aleksijuvani:
This pull request adds an example Xcode project that allows building the "window" example for iOS and the iOS Simulator. I'm not entirely sure if this Xcode project is portable between machines, so I'd appreciate it if anyone with an access to a Mac could give it a try.

A few notes about code signing. Building for the Simulator doesn't require code signing. Building for a physical iOS device requires setting up code signing by clicking "Automatically manage signing" on the "Signing & Capabilities" tab of the project and selecting the correct "team":

Screenshot 2019-10-12 at 9 47 23

This requires a free Apple Developer account, and for the account to be added to Xcode. See [this post](https://9to5mac.com/2016/03/27/how-to-create-free-apple-developer-account-sideload-apps/) for how to create an Apple Developer account how to add it to Xcode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ target/
 *~
 #*#
 headless.png
+xcuserdata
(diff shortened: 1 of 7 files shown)


goddessfreya:
Apologies, some how I missed this PR. Is this different from https://github.com/rust-windowing/glutin/pull/1233 ?

aleksijuvani:
This pull request takes a different approach than #1233. Instead of building Rust code into a universal library and then using Objective-C to invoke the library, this pull request simply invokes `cargo build` as a build step.

goddessfreya:
I see. Is it possible to move these project files under `glutin_examples/ios-example2`? I wouldn't want example code being sprinkled around the repository, it makes it hard to find them.

Also, how about adding this project to the CI? I'd hate if this example inadvertently broke, and no one noticed.

A small (readme)[https://github.com/rust-windowing/glutin/blob/master/glutin_examples/ios-example/README.md] explaining the differences between this and the other example would be helpful I image. And some limited instructions on how to build it.

aleksijuvani:
I don't think we'll want to have both of these examples in the repository. We might as well just keep #1233 since it's already been merged.

goddessfreya:
Fair enough. Closing.

PR #1226, opened 2019-10-12
Disable caching on Travis

aleksijuvani:
Travis builds are stalling due to the caches growing too large. See rust-windowing/winit#918 for the same issue in the winit repository. This pull request solves the problem by disabling caching.


goddessfreya:
Thanks!

PR #1230, opened 2019-10-18
Changed type of vsync from bool to Option<bool>

Tacklebox:
Hi,

This is to address issue #631 

I am not really sure about  glutin/src/api/egl/mod.rs line 1089, if the Option is none, should it be treated as if the bool was false? The comment `// VSync defaults to enabled; disable it if it was not requested.` seems to be at odds with the ones in lib.rs `/// By default, vsync is not enabled.

### DIFF
--- a/glutin/src/api/egl/mod.rs
+++ b/glutin/src/api/egl/mod.rs
@@ -1086,19 +1086,21 @@ impl<'a> ContextPrototype<'a> {
 
         if let Some(surface) = surface {
             // VSync defaults to enabled; disable it if it was not requested.
-            if !self.opengl.vsync {
-                let _guard = MakeCurrentGuard::new(
-                    self.display,
-                    surface,
-                    surface,
-                    context,
-                )
-                .map_err(|err| CreationError::OsError(err))?;
+            if let Some(vsync) = self.opengl.vsync {
+                if !vsync {
+                    let _guard = MakeCurrentGuard::new(
+                        self.display,
+                        surface,
+                        surface,
+                        context,
+                        )
+                        .map_err(|err| CreationError::OsError(err))?;
 
-                let egl = EGL.as_ref().unwrap();
-                unsafe {
-                    if egl.SwapInterval(self.display, 0) == ffi::egl::FALSE {
-                        panic!("finish_impl: eglSwapInterval failed: 0x{:x}", egl.GetError());
+                    let egl = EGL.as_ref().unwrap();
+                    unsafe {
+                        if egl.SwapInterval(self.display, 0) == ffi::egl::FALSE {
+                            panic!("finish_impl: eglSwapInterval failed: 0x{:x}", egl.GetError());
+                        }
                     }
                 }
             }
@@ -1294,8 +1296,12 @@ where
     }
 
     // We're interested in those configs which allow our desired VSync.
-    let desired_swap_interval = if opengl.vsync {
-        1
+    let desired_swap_interval = if let Some(vsync) = opengl.vsync {
+        if vsync {
+            1
+        } else {
+            0
+        }
     } else {
         0
     };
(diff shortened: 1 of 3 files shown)


glutin/src/lib.rs
```diff
@@ -181,7 +181,7 @@ impl<'a, T: ContextCurrentState> ContextBuilder<'a, T> {
     ///
     /// By default, vsync is not enabled.
```
goddessfreya: Docs should be updated to say that `None` equals don't care.

glutin/src/lib.rs
```diff
@@ -705,7 +705,7 @@ pub struct GlAttributes<S> {
     /// screen tearing.
     ///
     /// The default is `false`.
```
goddessfreya: Docs should be updated to say that `None` equals don't care.

glutin/src/api/egl/mod.rs
```diff
@@ -1086,19 +1086,21 @@ impl<'a> ContextPrototype<'a> {
 
         if let Some(surface) = surface {
             // VSync defaults to enabled; disable it if it was not requested.
```
goddessfreya: EGL by default has VSync enabled, and you have to explicitly disable it. Hope that makes sense.

glutin/src/api/egl/mod.rs
```diff
@@ -1294,8 +1296,12 @@ where
     }
 
     // We're interested in those configs which allow our desired VSync.
-    let desired_swap_interval = if opengl.vsync {
```
goddessfreya: This code as it exists contradicts the prior code. The prior code does not disable vsync when `None` is received, yet sets the `desired_swap_interval` to 0 (no vsync).

If we don't care about `VSync`, then it doesn't matter what the `MIN_SWAP_INTERVAL`/`MAX_SWAP_INTERVAL` equals, so we don't even need the filter.

glutin/src/api/glx/mod.rs
```diff
@@ -413,51 +413,53 @@ impl<'a> ContextPrototype<'a> {
...
+                if check_ext(&self.extensions, "GLX_EXT_swap_control")
+                    && extra_functions.SwapIntervalEXT.is_loaded()
+                    {
+                        // this should be the most common extension
+                        unsafe {
+                            extra_functions.SwapIntervalEXT(
+                                self.xconn.display as *mut _,
+                                window,
+                                1,
+                                );
```
goddessfreya: Nit: formating?

glutin/src/api/glx/mod.rs
```diff
@@ -413,51 +413,53 @@ impl<'a> ContextPrototype<'a> {
...
-                    );
-                }
+                        let mut swap = unsafe { std::mem::zeroed() };
+                        unsafe {
+                            glx.QueryDrawable(
+                                self.xconn.display as *mut _,
+                                window,
+                                ffi::glx_extra::SWAP_INTERVAL_EXT as i32,
+                                &mut swap,
+                                );
```
goddessfreya: Nit: formating?

glutin/src/api/glx/mod.rs
```diff
@@ -413,51 +413,53 @@ impl<'a> ContextPrototype<'a> {
...
+                        } else if check_ext(&self.extensions, "GLX_SGI_swap_control")
+                            && extra_functions.SwapIntervalSGI.is_loaded()
+                            {
+                                unsafe {
+                                    extra_functions.SwapIntervalSGI(1);
+                                }
+                            } else {
+                                return Err(CreationError::OsError(
+                                        "Couldn't find any available vsync extension".to_string(),
+                                        ));
```
goddessfreya: Nit: formating?

goddessfreya:
Merry Christmas. Closing this PR due to inactivity. Cheers!

PR #1233, opened 2019-11-06
Added an xcode project for an ios build

simlay:
Hi friends,

I dunno if it's wanted or not but I added an example ios xcode project.

Let me know if you'd like anything changed or added.

Thanks,


simlay:
> Not familiar with ios, but lgtm. One question: is the example built by the CI? If not, can you make it so?

I’ll look into making work in CI.

simlay:
@goddessfreya i added ci integration. Though, honestly I’m a little worried that come the next release of iOS, it may break due to the Xcode build target simulator.

goddessfreya:
I hope not. Anyways, we'll burn that bridge when we cross it. Merging! Thanks.

Issue #926, opened 2017-08-22
Cannot disable vysnc.


thelearnerofcode:
```rust
let context = glutin::ContextBuilder::new() .with_vsync(false);
```
does not disable vsync as expected.
It seems to stall during the line:
```rust
self.window.swap_buffers().unwrap();
```
On gnome/wayland, if that makes a difference.

EDIT: tried with x-server instead of wayland, problem still continues.

jansol:
Make sure your driver or compositor settings don't force vsync on.

thelearnerofcode:
Thank you!

LukasKalbertodt:
For future visitors, this might be useful: try executing your application with `vblank_mode=0`. It worked for me on Ubuntu 18.04.

```
$ vblank_mode=0 cargo run
```

PR #1252, opened 2020-01-01
[0.21] Fix VSync not being disabled when requested

2xsaiko:
By default, the swap interval is set to 1 (see below). Since glutin does nothing when vsync=false, this value never gets set to 0, and vsync is enabled by default regardless of the value set.

The problem occurs because of the environment variable vblank_mode. When it is unset, it defaults to 2, which causes the swap interval to default to 1.

Currently, this can be worked around by starting the application with vblank_mode=1 (default swap interval to 0) or vblank_mode=0 (force swap interval to 0).

This patch causes glutin to always set the swap interval.

See #926

### DIFF
--- a/glutin/src/api/glx/mod.rs
+++ b/glutin/src/api/glx/mod.rs
@@ -431,52 +431,52 @@ impl<'a> ContextPrototype<'a> {
         let (extra_functions, context) = self.create_context()?;
 
         // vsync
-        if self.opengl.vsync {
-            let _guard = MakeCurrentGuard::new(&self.xconn, window, context)
-                .map_err(|err| CreationError::OsError(err))?;
+        let swap_mode = if self.opengl.vsync { 1 } else { 0 };
 
-            if check_ext(&self.extensions, "GLX_EXT_swap_control")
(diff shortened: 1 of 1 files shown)


goddessfreya:
Is there a reason this change is not being done against master?

2xsaiko:
Yeah, I'm using glium which uses 0.21, so that's what I had to use to test. Patch should work the same for master though, I'll rebase it in a sec.

2xsaiko:
I've opened a new PR in addition to this one for master.

goddessfreya:
Hey @2xsaiko, I just remembered I forgot to publish 0.21.3. Unfortunately, we've switched from Travis CI to Github actions. Can you open a PR that adds [this file](https://github.com/rust-windowing/glutin/blob/master/.github/workflows/publish.yml#L5) to the `backport21` branch, while changing [this line](https://github.com/rust-windowing/glutin/blob/master/.github/workflows/publish.yml#L5) to include the branch name? Can you also bump glutin's version in the `Cargo.toml` too? Thanks.

PR #1260, opened 2020-01-06
Update to winit 0.20.0

chrisduerr:
I've taken the liberty to bump winit to version 0.20.0, since the recent DPI changes all have been merged.

I think the only potentially controversial thing is that I've blindly made all `PhysicalSize` instances to `PhysicalSize<u32>`, since most usecases I saw in the code used `let _: (u32, u32) = size.into();` to immediately destructure the physical size into its width/height. However it's possible that there is an instance where u32 is not the optimal choice, I can't fully judge that myself.

### DIFF
--- a/glutin/Cargo.toml
+++ b/glutin/Cargo.toml
@@ -18,7 +18,7 @@ serde = ["winit/serde"]
 
 [dependencies]
 lazy_static = "1.3"
-winit = "=0.20.0-alpha4" #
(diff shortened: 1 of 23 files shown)


chrisduerr:
I've tried to go through the clunky github CI and hopefully got it all now. Should all be working now hopefully.

goddessfreya:
Sweet. Thanks!

PR #1274, opened 2020-02-05
Glutin0.23 android

dvc94ch:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,31 +64,32 @@ jobs:
       matrix:
         rust_version: [stable, nightly]
         platform:
-          - { target: x86_64-pc-windows-msvc,   os: windows-latest,  }
-          - { target: i686-pc-windows-msvc,     os: windows-latest,  }
-          - { target: x86_64-pc-windows-gnu,    os: windows-latest, host: -x86_64-pc-windows-gnu }
-          - { target: i686-pc-windows-gnu,      os: windows-latest, host: -i686-pc-windows-gnu }
+         # - { target: x86_64-pc-windows-msvc,   os: windows-latest,  }
+         # - { target: i686-pc-windows-msvc,     os: windows-latest,  }
+         # - { target: x86_64-pc-windows-gnu,    os: windows-latest, host: -x86_64-pc-windows-gnu }
+         # - { target: i686-pc-windows-gnu,      os: windows-latest, host: -i686-pc-windows-gnu }
           - { target: i686-unknown-linux-gnu,   os: ubuntu-latest,   }
           - { target: x86_64-unknown-linux-gnu, os: ubuntu-latest,   }
-          - { target: x86_64-apple-darwin,      os: macos-latest,    }
-          - { target: x86_64-apple-ios,         os: macos-latest,    }
-          - { target: armv7-apple-ios,          os: macos-latest,    }
-          - { target: aarch64-apple-ios,        os: macos-latest,    }
+          - { target: aarch64-linux-android,    os: ubuntu-latest, cmd: 'apk --' }
+         # - { target: x86_64-apple-darwin,      os: macos-latest,    }
+         # - { target: x86_64-apple-ios,         os: macos-latest,    }
+         # - { target: armv7-apple-ios,          os: macos-latest,    }
+         # - { target: aarch64-apple-ios,        os: macos-latest,    }
           # We're using Windows rather than Ubuntu to run the wasm tests because caching cargo-web
           # doesn't currently work on Linux.
-          #- { target: wasm32-unknown-unknown,   os: windows-latest, features: stdweb,  web: web }
-          #- { target: wasm32-unknown-unknown,   os: windows-latest, features: web-sys, web: web }
+          #- { target: wasm32-unknown-unknown,   os: windows-latest, features: stdweb,  cmd: web }
+          #- { target: wasm32-unknown-unknown,   os: windows-latest, features: web-sys, cmd: web }
 
     env:
       RUST_BACKTRACE: 1
       CARGO_INCREMENTAL: 0
       RUSTFLAGS: "-C debuginfo=0"
       FEATURES: ${{ format(',{0}', matrix.platform.features ) }}
-      WEB: ${{ matrix.platform.web }}
+      CMD: ${{ matrix.platform.cmd }}
 
     runs-on: ${{ matrix.platform.os }}
     steps:
-        
+
     # FIXME
     - uses: actions/checkout@v1
       with:
@@ -142,6 +143,10 @@ jobs:
       if: matrix.platform.os == 'ubuntu-latest'
       run: sudo apt-get update && sudo apt-get install libgbm-dev libdrm-dev
 
+    - name: Install cargo-apk
+      if: contains(matrix.platform.target, 'android')
+      run: cargo install cargo-apk
+
     - name: Install cargo-web
       continue-on-error: true
       if: contains(matrix.platform.target, 'wasm32')
@@ -150,19 +155,22 @@ jobs:
     - name: Check documentation
       shell: bash
       if: matrix.platform.target != 'wasm32-unknown-unknown'
-      run: cargo doc --no-deps --target ${{ matrix.platform.target }} --features $FEATURES
+      run: cargo $CMD doc --no-deps --target ${{ matrix.platform.target }} --features $FEATURES
 
-    - name: Build 
+    - name: Build
       shell: bash
-      run: cargo $WEB build --verbose --target ${{ matrix.platform.target }} --features $FEATURES
+      run: cargo $CMD build --verbose --target ${{ matrix.platform.target }} --features $FEATURES
 
     - name: Build tests
       shell: bash
-      run: cargo $WEB test --no-run --verbose --target ${{ matrix.platform.target }} --features $FEATURES
+      run: cargo $CMD test --no-run --verbose --target ${{ matrix.platform.target }} --features $FEATURES
     - name: Run tests
       shell: bash
-      if: (!contains(matrix.platform.target, 'ios') && !contains(matrix.platform.target, 'wasm32'))
-      run: cargo $WEB test --verbose --target ${{ matrix.platform.target }} --features $FEATURES
+      if: (
+        !contains(matrix.platform.target, 'android') &&
+        !contains(matrix.platform.target, 'ios') &&
+        !contains(matrix.platform.target, 'wasm32'))
+      run: cargo $CMD test --verbose --target ${{ matrix.platform.target }} --features $FEATURES
 
     - name: iOS example
       shell: bash
--- a/examples/window.rs
+++ b/examples/window.rs
@@ -7,45 +7,70 @@ use winit::event::{Event, WindowEvent};
 use winit::event_loop::{ControlFlow, EventLoop};
 use winit::window::WindowBuilder;
 
+#[cfg(target_os = "android")]
+ndk_glue::ndk_glue!(main);
+
 fn main() {
     simple_logger::init().unwrap();
     let el = EventLoop::new();
     let wb = WindowBuilder::new().with_title("A fantastic window!");
 
     let confs = unsafe { ConfigsFinder::new().find(&*el).unwrap() };
-    let conf = &confs[0];
+    let conf = confs[0].clone();
     println!("Configeration chosen: {:?}", conf);
 
-    let ctx = unsafe { ContextBuilder::new().build(conf).unwrap() };
-    let (win, surf) = unsafe { Surface::new_window(conf, &*el, wb).unwrap() };
+    let ctx = unsafe { ContextBuilder::new().build(&conf).unwrap() };
+    let win = unsafe { Surface::build_window(&conf, &*el, wb).unwrap() };
 
-    unsafe { ctx.make_current(&surf).unwrap() }
-    let gl = support::Gl::load(|s| ctx.get_proc_address(s).unwrap());
-
+    // On android the surface can only be created after the resume event
+    // was received.
+    let mut surf = None;
+    let mut gl = None;
     el.run(move |event, _, control_flow| {
         println!("{:?}", event);
         *control_flow = ControlFlow::Wait;
 
         match event {
+            Event::Resumed => {
+                let surface = unsafe { Surface::new_from_existing_window(&conf, &win).unwrap() };
+                unsafe {
+                    ctx.make_current(&surface).unwrap();
+                }
+                if gl.is_none() {
+                    gl = Some(support::Gl::load(|s| ctx.get_proc_address(s).unwrap()));
+                }
+                surf = Some(surface)
+            }
+            Event::Suspended => surf = None,
             Event::LoopDestroyed => return,
-            Event::MainEventsCleared => {
-                win.request_redraw();
-            }
+            Event::MainEventsCleared => win.request_redraw(),
             Event::RedrawRequested(_) => {
-                gl.draw_frame([1.0, 0.5, 0.7, 1.0]);
-                surf.swap_buffers().unwrap();
+                if let (Some(gl), Some(surf)) = (&gl, &surf) {
+                    gl.draw_frame([1.0, 0.5, 0.7, 1.0]);
+                    surf.swap_buffers().unwrap();
+                }
             }
-            Event::WindowEvent { ref event, .. } => match event {
-                WindowEvent::Resized(size) => {
+            Event::WindowEvent { ref event, .. } => {
+                let size = match event {
+                    WindowEvent::ScaleFactorChanged {
+                        new_inner_size: size,
+                        ..
+                    } => size,
+                    WindowEvent::Resized(size) => size,
+                    WindowEvent::CloseRequested => {
+                        *control_flow = ControlFlow::Exit;
+                        return;
+                    }
+                    _ => return,
+                };
+                if let (Some(gl), Some(surf)) = (&gl, &surf) {
                     ctx.update_after_resize();
                     surf.update_after_resize(*size);
                     unsafe {
                         gl.gl.Viewport(0, 0, size.width as _, size.height as _);
                     }
                 }
-                WindowEvent::CloseRequested => *control_flow = ControlFlow::Exit,
-                _ => (),
-            },
+            }
             _ => (),
         }
     });
--- a/src/platform_impl/android/android.rs
+++ b/src/platform_impl/android/android.rs
@@ -1,4 +1,224 @@
 #![cfg(target_os = "android")]
 
-pub use crate::api::android::*;
-pub use winit::event_loop::{EventLoop, EventLoopWindowTarget};
+use crate::api::egl;
+use crate::config::{ConfigAttribs, ConfigWrapper, ConfigsFinder, SwapInterval};
+use crate::context::ContextBuilderWrapper;
+pub use crate::platform::android::ConfigPlatformAttributes;
+use crate::surface::{PBuffer, Pixmap, SurfaceTypeTrait, Window};
+use glutin_interface::{
+    AndroidWindowParts, NativeDisplay, NativePixmap, NativePixmapSource, NativeWindow,
+    NativeWindowSource, RawWindow, Seal,
+};
+use std::ops::Deref;
+use std::os::raw;
+use winit_types::dpi;
+use winit_types::error::{Error, ErrorType};
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Config(egl::Config);
+
+impl Deref for Config {
+    type Target = egl::Config;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Config {
+    pub fn new<ND: NativeDisplay>(
+        cf: &ConfigsFinder,
+        nd: &ND,
+    ) -> Result<Vec<(ConfigAttribs, Config)>, Error> {
+        Ok(
+            egl::Config::new(cf, nd, |confs, _| confs.into_iter().map(Ok).collect())?
+                .into_iter()
+                .map(|(attribs, config)| (attribs, Config(config)))
+                .collect(),
+        )
+    }
+
+    pub fn raw_config(&self) -> *const raw::c_void {
+        (**self).raw_config()
+    }
+
+    pub fn raw_display(&self) -> *mut raw::c_void {
+        (**self).raw_display()
+    }
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub(crate) struct Surface<T: SurfaceTypeTrait>(egl::Surface<T>);
+
+impl<T: SurfaceTypeTrait> Deref for Surface<T> {
+    type Target = egl::Surface<T>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl<T: SurfaceTypeTrait> Surface<T> {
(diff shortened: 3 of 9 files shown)


src/context.rs
```diff
@@ -145,7 +145,7 @@ impl Context {
         if self.get_config() != read_surf.get_config()
             || self.get_config() != write_surf.get_config()
         {
-            warn!("[glutin] `make_current`: Your surfaces' and context's configurations don't match. Are you sure this is intentional?")
+            log::warn!("[glutin] `make_current`: Your surfaces' and context's configurations don't match. Are you sure this is intentional?")
```
goddessfreya: Should probably get rid of the platform-guard here instead, no? https://github.com/rust-windowing/glutin/blob/e6b71859a19c63deac7996cef964cea8ce8c949a/src/lib.rs#L335

src/platform_impl/android/android.rs
```diff
@@ -1,4 +1,304 @@
...
+            if suspended {
+                if let Some(surface) = inner2.lock().take() {
+                    surface.make_not_current().ok();
+                }
+            } else {
+                let display = Arc::clone(&config2.config.display);
+                let config = config2.config.config;
+                let egl = egl::EGL.as_ref().unwrap();
+                let surface =
+                    egl.CreateWindowSurface(**display, config, *a_native_window, std::ptr::null());
```
goddessfreya: What about the descriptors? https://github.com/rust-windowing/glutin/blob/e6b71859a19c63deac7996cef964cea8ce8c949a/src/api/egl.rs#L1200

Can't we just reuse that `new` function?

src/platform_impl/android/android.rs
```diff
@@ -1,4 +1,304 @@
...
+            }
+        })));
+        Ok((nw, Surface { inner, config }))
+    }
+
+    pub unsafe fn new_existing<NW: NativeWindow>(
+        _conf: ConfigWrapper<&Config, &ConfigAttribs>,
+        _nw: &NW,
+    ) -> Result<Self, Error> {
+        Err(make_error!(ErrorType::NotSupported(
```
goddessfreya: If `set_suspend_callback` is added to `NW` instead, couldn't we support `new_existing` and then implement `new` in terms of `new_existing`, as x11 and wayland do?

src/platform_impl/android/android.rs
```diff
@@ -1,4 +1,304 @@
...
+        _conf: ConfigWrapper<&Config, &ConfigAttribs>,
+        _nps: &NPS,
+        _pb: NPS::PixmapBuilder,
+    ) -> Result<(NPS::Pixmap, Self), Error> {
+        return Err(make_error!(ErrorType::NotSupported(
+            "pixmaps not supported on android".into()
+        )));
+        // avoid unused code warning
+        let conf = _conf.map_config(|conf| &**conf);
+        egl::Surface::<Pixmap>::new(conf, std::ptr::null_mut()).ok();
```
goddessfreya: Can't we just add a platform guard to that function to avoid doing this?

examples/window.rs
```diff
@@ -19,28 +19,34 @@ fn main() {
...
-    unsafe { ctx.make_current(&surf).unwrap() }
-    let gl = support::Gl::load(|s| ctx.get_proc_address(s).unwrap());
-
+    let mut gl = None;
     el.run(move |event, _, control_flow| {
         println!("{:?}", event);
         *control_flow = ControlFlow::Wait;
 
         match event {
+            Event::Resumed => {
```
goddessfreya: Shouldn't all the other examples change too? Also, I didn't get any `Resume` event on X11, so this would have to be `Event::NewEvents(Init) | Event::Resumed`. I think we should also add a comment explaining that this is for android support.

src/platform_impl/android/android.rs
dvc94ch: I'm not sure that android supports any. It crashed with the default implementation with x3004 invalid attributes or something like that.
The other difference is using `CreateWindowSurface` vs `CreatePlatformWindowSurface`.

examples/window.rs
dvc94ch: This callback stuff is probably not such a good idea. How about adding on_resumed and on_suspended functions to `Surface` and `Context` instead?

dvc94ch: the `a_native_window` ptr may be null if it's called before `Event::Resumed` on android. we may have to add some special casing. Do you think people would be open to emitting a `Event::Resumed` on desktop platforms?

src/platform_impl/android/android.rs
goddessfreya: > I'm not sure that android supports any. It crashed with the default implementation with x3004 invalid attributes or something like that.

I noticed something similar to that w/ gbm on some srgb configs but didn't look too far into it. Good to know.

> The other difference is using CreateWindowSurface vs CreatePlatformWindowSurface.

Afaik, using `CreatePlatformWindowSurface` should be fine, no? `EGL_KHR_platform_android` mentions using `eglCreatePlatformWindowSurface`, then later on mentions the following:
```
    1. Should this extension even exist? Android devices only support one
       window system.

       RESOLUTION: Yes. Although the Android Open Source Project master branch 
       only supports one window system, customized versions of Android could
       extend that to support other window systems. More importantly, having a
       platform extension allows EGL 1.5 applications to use the platform and
       non-platform Get*Display and Create*WindowSurface calls interchangeably. 
       As a user of the API it would be confusing if that didn't work.
```

src/api/egl.rs
```diff
@@ -1284,6 +1236,36 @@ impl Surface<Window> {
...
+        nwin: *mut raw::c_void,
+    ) -> Result<Self, Error> {
+        let display = Arc::clone(&conf.config.display);
+        let egl = EGL.as_ref().unwrap();
+        let surface = unsafe {
+            let surf =
+                egl.CreateWindowSurface(**display, conf.config.config, nwin, std::ptr::null());
+            if surf.is_null() {
+                return Err(make_oserror!(OsError::Misc(format!(
+                    "eglCreatePlatformWindowSurface failed with 0x{:x}",
```
goddessfreya: Outdated.

src/api/egl.rs
```diff
@@ -1284,6 +1236,36 @@ impl Surface<Window> {
         })
     }
 
+    #[cfg(target_os = "android")]
```
goddessfreya: How about instead of making a second function, we go:
```
#[cfg(not(target_os = "android"))]
let desc = Self::assemble_desc(conf.clone(), None);
#[cfg(not(target_os = "android"))]
let desc = desc.as_ptr();

#[cfg(target_os = "android")]
let desc = std::ptr::null();
```

This is assuming `CreatePlatformWindowSurface`  works fine on Android.

examples/window.rs
goddessfreya: > This callback stuff is probably not such a good idea. How about adding on_resumed and on_suspended functions to Surface and Context instead?

Would that be like the `update_after_resize` functions? I hope not to add anything else like them.

goddessfreya: > Do you think people would be open to emitting an Event::Resumed on desktop platforms?

Sounds fine for me, if it will make the lives of downstream users easier.

src/api/egl.rs
dvc94ch: `CreatePlatformWindowSurface` gives me a not loaded error, otherwise I'd have gone with that.

goddessfreya: Are the systems EGL 1.4 or lacking EGL_KHR_platform_android? Shame.

goddessfreya:
To handle the android lifetime cycle thing, can't we just have users drop the `Surface` and recreate it with `new_existing` when their application resumes? Is that not like the whole point of `new_existing`? Why does glutin have to drop it internally when the users can drop it themselves? I think this would be a better solution than platform-specific functions or weird callback setting functions on `glutin_interface`.

examples/window.rs
```diff
@@ -19,28 +19,51 @@ fn main() {
...
+                surf.on_resume().unwrap();
+                unsafe { ctx.make_current(&surf).unwrap() }
+                gl = Some(support::Gl::load(|s| ctx.get_proc_address(s).unwrap()));
+            }
+            Event::Suspended => {
+                surf.on_suspend().unwrap();
+                gl = None
```
goddessfreya: Why do we need to reload the pointers? Is this something android specific?

Going off the egl docs:
>  Client API function pointers returned by eglGetProcAddress are independent of the display and the currently bound client API context, and may be used by any client API context which supports the function.

examples/window.rs
```diff
@@ -19,28 +19,51 @@ fn main() {
...
+            }
             Event::LoopDestroyed => return,
             Event::MainEventsCleared => {
                 win.request_redraw();
             }
             Event::RedrawRequested(_) => {
-                gl.draw_frame([1.0, 0.5, 0.7, 1.0]);
+                gl.as_ref().unwrap().draw_frame([1.0, 0.5, 0.7, 1.0]);
```
goddessfreya: Is it possible for `RedrawRequested`/`MainEventsCleared` to arrive before the initial resume on some devices? If so, I don't think we should panic here.

examples/window.rs
dvc94ch: The reason for that is to make sure no drawing is done after a suspend. It's only ok to do it between the InitWindow/TermWindow events, which emit Resumed/Suspended events in winit.

examples/window.rs
dvc94ch: Don't know if this can happen, that's why I set gl to None. I guess being more defensive can't hurt

examples/window.rs
goddessfreya: I see. Well, we can have a bool tracking if we are currently resumed or suspended, and we can load the symbols on the first `Resume` only. If the bool is unset when `RedrawRequested` is triggered we can just no-op instead of panicing.

dvc94ch:
> To handle the android lifetime cycle thing, can't we just have users drop the Surface and recreate it with new_existing when their application resumes?

Are you suggesting something like this? I'm not very pleased with this solution, but maybe I missunderstood. `glutin::platform_impl::Config` and `winit::platform_impl::EventLoop` must be `Clone`.

```diff
diff --git a/examples/window.rs b/examples/window.rs
index 5560af8..7f87ec5 100644
--- a/examples/window.rs
+++ b/examples/window.rs
@@ -13,62 +13,70 @@ fn main() {
     let wb = WindowBuilder::new().with_title("A fantastic window!");
 
     let confs = unsafe { ConfigsFinder::new().find(&*el).unwrap() };
-    let conf = &confs[0];
+    let conf = confs[0].clone();
     println!("Configeration chosen: {:?}", conf);
 
-    let ctx = unsafe { ContextBuilder::new().build(conf).unwrap() };
-    let (win, surf) = unsafe { Surface::new_window(conf, &*el, wb).unwrap() };
+    let ctx = unsafe { ContextBuilder::new().build(&conf).unwrap() };
 
+    let mut win = None;
+    let mut surf = None;
     let mut gl = None;
-    el.run(move |event, _, control_flow| {
+    el.clone().run(move |event, _, control_flow| {
         println!("{:?}", event);
         *control_flow = ControlFlow::Wait;
 
         match event {
             Event::Resumed => {
-                surf.on_resume().unwrap();
-                unsafe { ctx.make_current(&surf).unwrap() }
-                gl = Some(support::Gl::load(|s| ctx.get_proc_address(s).unwrap()));
+                let surface = if let Some(win) = &win {
+                    unsafe { Surface::new_from_existing_window(&conf, win).unwrap() }
+                } else {
+                    let (window, surf) = unsafe { Surface::new_window(&conf, &*el, wb.clone()).unwrap() };
+                    win = Some(window);
+                    gl = Some(support::Gl::load(|s| ctx.get_proc_address(s).unwrap()));
+                    surf
+                };
+                unsafe { ctx.make_current(&surface).unwrap(); }
+                surf = Some(surface);
             }
             Event::Suspended => {
-                surf.on_suspend().unwrap();
-                gl = None
+                if let Some(surf) = &surf {
+                    unsafe { surf.make_not_current() };
+                }
+                surf = None;
             }
             Event::LoopDestroyed => return,
             Event::MainEventsCleared => {
-                win.request_redraw();
+                if let Some(win) = &win {
+                    win.request_redraw();
+                }
             }
             Event::RedrawRequested(_) => {
-                gl.as_ref().unwrap().draw_frame([1.0, 0.5, 0.7, 1.0]);
-                surf.swap_buffers().unwrap();
+                if let (Some(gl), Some(surf)) = (&gl, &surf) {
+                    gl.draw_frame([1.0, 0.5, 0.7, 1.0]);
+                    surf.swap_buffers().unwrap();
+                }
             }
-            Event::WindowEvent { ref event, .. } => match event {
-                WindowEvent::ScaleFactorChanged {
-                    new_inner_size: size,
-                    ..
-                } => {
-                    ctx.update_after_resize();
-                    surf.update_after_resize(**size);
-                    unsafe {
-                        gl.as_ref()
-                            .unwrap()
-                            .gl
-                            .Viewport(0, 0, size.width as _, size.height as _);
+            Event::WindowEvent { ref event, .. } => {
+                let size = match event {
+                    WindowEvent::ScaleFactorChanged {
+                        new_inner_size: size,
+                        ..
+                    } => size,
+                    WindowEvent::Resized(size) => size,
+                    WindowEvent::CloseRequested => {
+                        *control_flow = ControlFlow::Exit;
+                        return;
                     }
-                }
-                WindowEvent::Resized(size) => {
+                    _ => return,
+                };
+                if let (Some(gl), Some(surf)) = (&gl, &surf) {
                     ctx.update_after_resize();
                     surf.update_after_resize(*size);
                     unsafe {
-                        gl.as_ref()
-                            .unwrap()
-                            .gl
-                            .Viewport(0, 0, size.width as _, size.height as _);
+                        gl.gl.Viewport(0, 0, size.width as _, size.height as _);
                     }
                 }
-                WindowEvent::CloseRequested => *control_flow = ControlFlow::Exit,
-                _ => (),
-            },
+            }
             _ => (),
         }
     });
```

dvc94ch:
actually if all platforms emit `Suspend` `Resume` in winit and the glutin api gains a `Surface::<Window>::build_window() -> Window`, not sure about other platform's requirements, I think it would be neat.

goddessfreya:
> Are you suggesting something like this? I'm not very pleased with this solution, but maybe I missunderstood. `glutin::platform_impl::Config` and `winit::platform_impl::EventLoop` must be `Clone`. 

Ah, well, `glutin::platform_impl::Config` should be clone. I marked the ones in `src/api/{glx,egl}.rs` as such, but forgot to mark the types in `src/platform_impl/unix/*`, so oops. I've updated the `sendsync` test to also check for clone.

Meanwhile, you don't need an `winit::platform_impl::EventLoop`, but instead an `&winit::platform_impl::EventLoopWindowTarget`, which is the second param that winit passes to the `run` closure, so no need to clone the eventloop.

I should also note that dropping `Surf` automatically makes it not current, so you don't need the `make_not_current` in `Suspended`.

> I guess if there was a `Surface::<Window>::build_window() -> Window` and `Surface::<Window>::new()` was implemented as `(Self::build_window(), Self::new_existing())` it would be less weird.

This sounds like an excellent idea, now that I come to think of it. Yeah, we should do that and then platforms don't have to implement `new` and instead only `new_existing`.

> actually if all platforms emit Suspend Resume in winit and the glutin api gains a Surface::<Window>::build_window() -> Window, not sure about other platform's requirements, I think it would be neat.

Yeah, should be fine on GLX and EGL, as, like I mentioned above, that is what was intended. Off my limited knowledge of how WGL works, I can't imagine anything will go wrong. No clue about cocoa though.

src/api/egl.rs
dvc94ch: > Are the systems EGL 1.4 or lacking EGL_KHR_platform_android?
Looks like both.

```sh
Configeration chosen: ConfigWrapper { attribs: ConfigAttribs { swap_interval_ranges: [Wait(1..2), DontWait], version: (OpenGlEs, Version(2, 0)), hardware_accelerated: true, color_bits: 24, alpha_bits: 0, depth_bits: 0, stencil_bits: 0, stereoscopy: false, double_buffer: true, multisampling: None, srgb: false, supports_pbuffers: true, supports_pixmaps: false, supports_windows: true, supports_surfaceless: true, float_color_buffer: false }, config: Config(Config { display: Display { display: 0x1, egl_version: (1, 4), extensions: ["EGL_KHR_get_all_proc_addresses", "EGL_ANDROID_presentation_time", "EGL_KHR_swap_buffers_with_damage", "EGL_ANDROID_get_native_client_buffer", "EGL_ANDROID_front_buffer_auto_refresh", "EGL_ANDROID_get_frame_timestamps", "EGL_EXT_surface_SMPTE2086_metadata", "EGL_EXT_surface_CTA861_3_metadata", "EGL_KHR_image", "EGL_KHR_image_base", "EGL_EXT_image_gl_colorspace", "EGL_KHR_lock_surface", "EGL_KHR_gl_colorspace", "EGL_KHR_gl_texture_2D_image", "EGL_KHR_gl_texture_3D_image", "EGL_KHR_gl_texture_cubemap_image", "EGL_KHR_gl_renderbuffer_image", "EGL_KHR_reusable_sync", "EGL_KHR_fence_sync", "EGL_KHR_create_context", "EGL_KHR_surfaceless_context", "EGL_EXT_create_context_robustness", "EGL_ANDROID_image_native_buffer", "EGL_KHR_wait_sync", "EGL_ANDROID_recordable", "EGL_KHR_partial_update", "EGL_EXT_pixel_format_float", "EGL_KHR_create_context_no_error", "EGL_KHR_mutable_render_buffer", "EGL_EXT_yuv_surface", "EGL_EXT_protected_content", "EGL_IMG_context_priority", "EGL_KHR_no_config_context", ""], client_extensions: [] }, config: 0x7dc43773e0 }) }
```

src/surface.rs
```diff
@@ -335,8 +335,37 @@ impl Surface<Window> {
...
+        let surface = platform_impl::Surface::<Window>::new_existing(conf.as_ref(), &window)?;
+        Ok((window, Surface(surface)))
+    }
+
+    /// Takes an `NWS` and its `NWS::WindowBuilder` type, returning a
+    /// `NWS::Window`.
+    ///
+    /// On Wayland, the [`Config`]'s `ND` must provide the same Wayland
+    /// connection as `NWS`. X11 is more lenient on this matter, allowing
+    /// different connections to the same display and screen. Other platforms
```
goddessfreya: Can you test this on android, and update the docs here, above at `new_window` and below at `new_from_existing_window`. Also, listing any other restrictions you might encounter on `new_from_existing_window` would be nice.

Honestly, to avoid doc duplication, maybe we should just have `new_window` and `build_window`'s docs link to `new_from_existing_window`'s docs?

src/surface.rs
```diff
@@ -335,8 +335,37 @@ impl Surface<Window> {
...
+    /// have not been tested.
+    ///
+    /// `EGL_EXT_platform_device` and `EGL_MESA_platform_surfaceless` do not
+    /// support windows.
+    ///
```
goddessfreya: Missing `Saftey` section.

src/surface.rs
```diff
@@ -335,8 +335,37 @@ impl Surface<Window> {
...
+    /// [`Window`]: crate::surface::Window
+    /// [`Config`]: crate::config::Config
+    #[inline]
+    pub unsafe fn build_window<NWS: NativeWindowSource>(
```
goddessfreya: I think we should do something similar to pixmaps, if only for consistency's sake.

src/api/egl.rs
```diff
@@ -1261,10 +1212,22 @@ impl Surface<Window> {
...
+        let desc = std::ptr::null();
+
         let surface = unsafe {
-            let surf =
-                egl.CreatePlatformWindowSurface(**display, conf.config.config, nwin, desc.as_ptr());
+            let surf = if egl.CreatePlatformWindowSurface.is_loaded() {
+                egl.CreatePlatformWindowSurface(**display, conf.config.config, nwin, desc)
+            } else {
+                // Used on android
+                egl.CreateWindowSurface(**display, conf.config.config, nwin, desc as _)
```
goddessfreya: This function might end up on use on non-android platforms where `desc` will not be null. Last I checked, `desc as _` would not be a valid cast as `EGLint` and `EGLAttrib` have different sizes, `i32` and `isize`, respectively.

src/surface.rs
```diff
@@ -215,19 +240,18 @@ impl Surface<Pixmap> {
     ///
     /// [`Pixmap`]: crate::surface::Pixmap
     #[inline]
-    pub unsafe fn new_pixmap<NPS: NativePixmapSource>(
+    pub unsafe fn build_pixmap<NPS: NativePixmapSource>(
```
goddessfreya: This function's `Safety` section should be copied to `new_pixmap`, as is. There is no surface being returned, so this function's safety should be something like `The returned NPS::Pixmap should not outlive the Config's ND`.

You should do a similar thing for the `build_window` and `new_window` functions.

Also, I just noticed that I misspelt safety everywhere. My bad. Could you fix that?

dvc94ch:
I rebased the changes and I think you can merge all except the last commit which is specific to the winit changes. That way I don't hold up other platform implementations.

goddessfreya:
Okay, one last thing (I swear): is this comment still up to date?

https://github.com/rust-windowing/glutin/pull/1274/files#diff-5ce72bac836d0ca6fa8f3082f38d28d2L170-L172

dvc94ch:
I can't find the comment, I think the commit you linked is outdated?

goddessfreya:
https://github.com/dvc94ch/glutin/blob/glutin0.23-android/src/api/egl.rs#L178-L180

.github/workflows/ci.yml
```diff
@@ -41,9 +41,10 @@ jobs:
         ref: master
     - uses: actions/checkout@v1
       with:
-        repository: goddessfreya/winit
+        # FIXME
+        repository: dvc94ch/winit
         path: winit
-        ref: winit-glutin-next
+        ref: winit-glutin-next-android
```
dvc94ch: NOTE to self

.github/workflows/ci.yml
```diff
@@ -142,6 +144,11 @@ jobs:
       if: matrix.platform.os == 'ubuntu-latest'
       run: sudo apt-get update && sudo apt-get install libgbm-dev libdrm-dev
 
+    - name: Install cargo-apk
+      if: contains(matrix.platform.target, 'android')
+      # FIXME
+      run: cargo install --git https://github.com/dvc94ch/android-ndk-rs --branch android-glue
+
```
dvc94ch: NOTE to self

examples/window.rs
```diff
@@ -7,45 +7,83 @@ use winit::event::{Event, WindowEvent};
 use winit::event_loop::{ControlFlow, EventLoop};
 use winit::window::WindowBuilder;
 
+#[cfg(target_os = "android")]
```
goddessfreya: We should move this to the support module so it can be shared by the other examples, if we decide to change them.

dvc94ch: Mmh the only way we could move this to the support module is turning it into a macro, since it needs to be in the same scope as main. This needing to be included in every file and having to specify `cdylib` manually are the downsides of the new `cargo-apk`. The advantage is a hughe reduction in complexity, not relying on native-app-glue and more flexibility/ease of adding features.

goddessfreya: Could we not use `super::main` and require that main is `pub`? If it has to be a macro, then so be it, still beats duplicating it between multiple different examples.

dvc94ch:
Closing due to inactivity

PR #1283, opened 2020-02-28
Don't unnecessarily instantiate GLX or EGL

dabreegster:
On my system, this removes a 1-4s delay. I think this delay was introduced sometime 0.21.0. I only tested on my one Ubuntu 18.04 box.

- [X ] Compilation warnings were addressed
- [X] `cargo fmt` has been run on this branch
- [X] `cargo doc` builds successfully
- [n/a] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [n/a] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [n/a] Created or updated an example program if it would help users understand this functionality

### DIFF
--- a/glutin/src/platform_impl/unix/x11.rs
+++ b/glutin/src/platform_impl/unix/x11.rs
@@ -401,10 +401,14 @@ impl Context {
                         "both libGL and libEGL are not present".to_string(),
                     ));
                 } else {
-                    match (&*GLX, &*EGL, prefer_egl) {
-                        (Some(_), Some(_), true) => return egl(builder_egl_u),
-                        (Some(_), Some(_), false) => return glx(builder_glx_u),
-                        _ => (),
+                    if prefer_egl {
+                        if let Some(_) = &*EGL {
+                            return egl(builder_egl_u);
+                        }
+                    } else {
+                        if let Some(_) = &*GLX {
+                            return glx(builder_glx_u);
+                        }
                     }
 
                     return Err(CreationError::NotSupported(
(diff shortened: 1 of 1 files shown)


dabreegster:
> Could you do that here as well
> 
> https://github.com/rust-windowing/glutin/blob/33aec609b6da89afc3c59b8b6e294b00c52fc12b/glutin/src/platform_impl/unix/x11.rs#L400
> .
> 
> The problem is that glutin lazily searches for libraries on Deref, which I don't like but I guess it's fine to keep it the way it's right now. Just doing `if let` in a way you did should fix it as well.

Done. I slightly changed the logic; before the `match` would demand both existed, even though it only used one. But if I'm reading the comments on `force_prefer_unless_only` correctly, I don't think it was correct before.

kchibisov:
>Done. I slightly changed the logic; before the match would demand both existed, even though it only used one. But if I'm reading the comments on force_prefer_unless_only correctly, I don't think it was correct before.

I'm not certain about the old code either, but logically the thing you're doing with if let is correct and can't break anything, since it just doesn't make any sense to query both libraries in the same time. yeah, it's not that obvious that deref is actually doing something non trivial, but it's what it's now.

Like if the code is actually wrong here even before, it'll transform naturally to a wrong code now as you can see. So there's a question whether this comment correct or whether the code is correct. And I start feeling that there's something wrong about this comment, but to actually understand that better it'll be nice to understand why this comment is here and why the code looks like it's now.

I'll try to do that when I'll have time, but you can also try to investigate if you want so.

As for PR, it seems fine from 'no new bugs were added' point of view.

PR #1285, opened 2020-03-09
Bump winit to 0.22.0

chrisduerr:


chrisduerr:
According to their GitHub status, gentz doesn't seem to be available at the moment.

Maybe @murarth or @Osspial can take over at least for these basic maintenance tasks for now? I don't think there's any glutin specific knowledge required to bump winit and release another version.

chrisduerr:
@murarth You've approved this but haven't merged it, I'd assume you're not comfortable merging updates/releasing new versions on your own since you're primarily responsible for the Linux winit backend?

I'd honestly think that if neither @goddessfreya nor @Osspial are currently available, it would be best to move forward without them. I don't think it does anyone any favors to leave crates unmaintained for any amount of time, at least as a consumer suffering from these major bugs I'd prefer to apologize later rather than waiting endlessly for approval.

Though the best scenario of course would be if someone capable of maintaining glutin signs off on this, but unfortunately I don't think there's anyone other than gentz.

PR #1287, opened 2020-03-12
Support resize for fullscreen example

kaimast:
I slightly modified the fullscreen example so the window_context resizes properly. Would be cool if you could merge this, if not nbd.


kchibisov:
The issue it raised were on Wayland right?

kaimast:
> The issue it raised were on Wayland right?

Yes

kchibisov:
it makes sense to follow the rest of the examples here, I guess. I'm not really sure that Wayland handling of such things in winit is a bit weird, but I don't see any issue with more universal example (you can at least unfullscreen it correctly now).

PR #1289, opened 2020-03-29
Make gles work when not specifying version

derekdreery:
This PR makes glutin work with GLES when a requested version is not given. I think ignoring the version is ok because we don't know it until we try to create a context. I've copied the oldest gles version code to be conservative.

### DIFF
--- a/glutin/src/api/egl/mod.rs
+++ b/glutin/src/api/egl/mod.rs
@@ -1180,7 +1180,14 @@ where
                     out.push(ffi::egl::OPENGL_ES_BIT as raw::c_int);
                 }
             }
-            (Api::OpenGlEs, _) => unimplemented!(),
+            (Api::OpenGlEs, None) => {
+                if egl_version >= &(1, 3) {
+                    out.push(ffi::egl::RENDERABLE_TYPE as raw::c_int);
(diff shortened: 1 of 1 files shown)


derekdreery:
The test failure is spurious I think.

derekdreery:
ping

glutin/src/api/egl/mod.rs
```diff
@@ -1180,7 +1180,14 @@ where
                     out.push(ffi::egl::OPENGL_ES_BIT as raw::c_int);
                 }
             }
-            (Api::OpenGlEs, _) => unimplemented!(),
+            (Api::OpenGlEs, None) => {
```
kchibisov, line 1183: It seems like you can merge it with previous match arm?

kchibisov (changes requested):
Could also add changelog entry for that. If you're not interested in that patch I could take care of it...

derekdreery:
Hi @kchibisov would you mind finishing the patch - I've forgotten what I did it for. If there's an issue with you doing it I'd still be happy to, but probably easier for you.

kchibisov:
Closing in favor of #1404.

Issue #1292, opened 2020-04-15
[Windows] Glutin loads an unrelated libEGL.dll from the other app


syrel:
Hi 😃

A glutin based app was crashing for one of the users on Windows 10.
While debugging it, we noticed that there were unrelated library loads:
```
C:\Program Files\OtherUserApp\bin\pc-win64\libEGL.dll
C:\Program Files\OtherUserApp\bin\pc-win64\libGLESv2.dll
```
The flow which causes this load is:
`build_headless (glutin_sys)` -> `new_headless` -> `EGL lazy static` -> `LoadLibrary`

In `new_headless`, if no windowed sharing was specified, it tries to load `libEGL.dll` from the load path, which in user's case was in PATH from another program, and it led to a crash later on.

The library is defined here:
https://github.com/rust-windowing/glutin/blob/33aec609b6da89afc3c59b8b6e294b00c52fc12b/glutin/src/api/egl/mod.rs#L93

Windows guides suggest that when performing a dynamic loading we should search dynamic libraries within the app (relative) and/or rely on absolute paths for safety (https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order).

Would it be possible to update library path in [egl/mod.rs](https://github.com/rust-windowing/glutin/blob/master/glutin/src/api/egl/mod.rs) to relative within the glutin app?

Thank you 👍

PR #1293, opened 2020-04-15
[Windows]: Don't search for libEGL on PATH

NightRa:
Fixes https://github.com/rust-windowing/glutin/issues/1292.

Currently, Glutin in `new_headless -> EGL lazy static initializer` tries to load `libEGL.dll` **optimistically** from anywhere in the user's search to decide if to use it or not, which includes the user's PATH - which may include the DLL from unrelated programs on the user's PC, causing unwanted behavior and crashes.

This change avoids loading `libEGL.dll` from the user's PATH, but still uses the other default search paths windows uses.


kchibisov:
r @Osspial ?

Osspial:
Looks good. Thanks for pulling this together!

PR #1299, opened 2020-05-16
Bump macOS core depedencies

chrisduerr:


chrisduerr:
@Osspial It seems like @goddessfreya has taken a temporary vacation from glutin. Would you be willing to help out managing some of the basic PRs while he's gone? Or do you know of anyone else who's able to do so?

Last time I bothered @murarth until he created a release for me, but I don't think he's really comfortable making any changes to glutin either.

kchibisov:
Bumping to match winit and reduce the amount of deps. Also I don't see any breaking changes in those crates, just bug fixes + new functions.

PR #1310, opened 2020-06-22
Add license symlink

returntrip:
To make it easier for Linux distributions to ship the licenses text within the crate directory.


kchibisov:
Thanks for letting me know. Since glutin had some other issues with LICENSE and you had some wrong symlinks I've decided to open #1311, since the change is very trivial.

PR #1312, opened 2020-06-29
Bump gl_generator

returntrip:


glutin_glx_sys/Cargo.toml
```diff
@@ -10,7 +10,7 @@ build = "build.rs"
 edition = "2018"
 
 [build-dependencies]
-gl_generator = "0.13"
+gl_generator = "0.14"
 
 [target.'cfg(any(target_os = "linux", target_os = "freebsd", target_os="dragonfly", target_os="netbsd", target_os="openbsd"))'.dependencies]
-x11-dl = "2.18.3"
+x11-dl = "2.18.5"
```
kchibisov, line 16: There's no reason for that change, the '2.18.3' implies that crates can use '2.18.5', so revert it.

returntrip: Sorry, thanks for the explanation. Done.

Issue #1307, opened 2020-06-10
Error compiling to Android with cargo apk


naomijub:
Hi, I have tried to compile a project to android and I get this error.
Rust nightly:
```
error[E0599]: no method named `set_suspend_callback` found for reference `&winit::event_loop::EventLoopWindowTarget<T>` in the current scope
  --> /Users/julia.boeira/.cargo/registry/src/github.com-1ecc6299db9ec823/glutin-0.24.1/src/api/android/mod.rs:86:12
   |
86 |         el.set_suspend_callback(Some(Box::new(move |suspended| {
   |            ^^^^^^^^^^^^^^^^^^^^ method not found in `&winit::event_loop::EventLoopWindowTarget<T>`

error[E0615]: attempted to take value of method `SwapBuffersWithDamageKHR` on type `&api::egl::egl::Egl`
   --> /Users/julia.boeira/.cargo/registry/src/github.com-1ecc6299db9ec823/glutin-0.24.1/src/api/egl/mod.rs:694:17
    |
694 |         if !egl.SwapBuffersWithDamageKHR.is_loaded() {
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^ method, not a field
    |
help: use parentheses to call the method
    |
694 |         if !egl.SwapBuffersWithDamageKHR(_, _, _, _).is_loaded() {
    |                                         ^^^^^^^^^^^^

error[E0615]: attempted to take value of method `SwapBuffersWithDamageKHR` on type `&api::egl::egl::Egl`
   --> /Users/julia.boeira/.cargo/registry/src/github.com-1ecc6299db9ec823/glutin-0.24.1/src/api/egl/mod.rs:740:13
    |
740 |         egl.SwapBuffersWithDamageKHR.is_loaded()
    |             ^^^^^^^^^^^^^^^^^^^^^^^^ method, not a field
    |
help: use parentheses to call the method
    |
740 |         egl.SwapBuffersWithDamageKHR(_, _, _, _).is_loaded()
    |                                     ^^^^^^^^^^^^

error: aborting due to 3 previous errors

Some errors have detailed explanations: E0599, E0615.
For more information about an error, try `rustc --explain E0599`.
error: could not compile `glutin`.
```

rust stable:
```
error[E0599]: no associated item named `EPSILON` found for type `f64` in the current scope
[...]

PR #1313, opened 2020-07-07
[WIP] Android support

katyo:


katyo:
According to current winit's event handling [implementation details](/rust-windowing/winit/blob/3d5d05eac7d23c78e4e66789563b3d4285e3a4eb/src/platform_impl/android/mod.rs#L108-L127) there is no way directly hook Android's pause/resume events which can be used to create/destroy surface.

katyo:
To get it work we need solve at least two issues.

1. Create context after window actually created.
    This happens when we get `WindowCreated` event from *ndk-glue*.
    In winit this event corresponds to `Resumed` event.
2. Manage creating and destroying rendering surface.
    This should be done when we get `Resume/Suspend` events from *ndk-glue*.
    Winit currently missing corresponding events.

For now I see the next possible solutions:

1. Change *winit* API by introducing additional event: `WindowEvent::Created` related to corresponding event from *ndk-glue*.
    This event should be place for actual context initialization. On other platforms this event can be triggered synchronous when window created. In this case `Resume`/`Suspend` events may corresponds to Android`s one and can be used to manage rendering surface.
2. Override needed (low-level) callbacks directly in `NativeActivity` at runtime.
    In that case the `main` function should be called after `WindowCreated` event actually received (which requires `ndk-glue` changes or additional wrapper for it). This solution does not require *winit* API changes but looks too dirty for me and possibly may have less obvious flaws.

#1307

Kaiser1989:
Here we go, some running example based on your changes:

https://github.com/Kaiser1989/rust-android-example

dvc94ch:
> Change winit API by introducing additional event: WindowEvent::Created related to corresponding event from ndk-glue.
This event should be place for actual context initialization. On other platforms this event can be triggered synchronous when window created. In this case Resume/Suspend events may corresponds to Android`s one and can be used to manage rendering surface.

Those were removed because @goddessfreya and me came to the conclusion that glutin didn't need to know about those after the improvements to glutin 0.23. I'd still like to see those happen as it makes integrating with for example `libflutter_engine.so` in flutter-rs much more natural and less error prone. But it looks like no one is willing to update the macos/windows backends :/

PR #1326, opened 2020-10-02
Fix several crashes on Windows

mooman219:
Winit pushed https://github.com/rust-windowing/winit/commit/b4c6cdf9a33530a2116f37a21547341c34537be1 fixing several crashes that I'm currently experiencing in Windows, then pushed version 0.22.2. This PR updates the dependency version to 0.22.2 from 0.22.0.

### DIFF
--- a/glutin/Cargo.toml
+++ b/glutin/Cargo.toml
@@ -18,7 +18,7 @@ serde = ["winit/serde"]
 
 [dependencies]
 lazy_static = "1.3"
-winit = "0.22.0"
+winit = "0.22.2"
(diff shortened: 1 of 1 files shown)


kchibisov:
patch version isn't a breaking change for downstream and so you can use 0.22.2 by running `cargo update -p winit` in your binary, there's no point in requiring specific patch version.

mooman219:
I apologize, I was not aware that I could do that through a cargo command and I've updated my version locally. Thanks for the notice!

PR #1329, opened 2020-10-03
Make x11 and wayland features

ogoffart:
winit now has a x11 and wayland feature that can be used to disable one of them, make uses of these feature also in this crate.

My goal is to reduce binary size of my binaries that does not use wayland.

- [X] Compilation warnings were addressed  (i made sure not to introduce new warnings)
- [X] `cargo fmt` has been run on this branch
- [X] `cargo doc` builds successfully

### DIFF
--- a/glutin/Cargo.toml
+++ b/glutin/Cargo.toml
@@ -15,10 +15,13 @@ features = ["serde"]
 
 [features]
 serde = ["winit/serde"]
+x11 = ["winit/x11", "glutin_glx_sys"]
+wayland = ["winit/wayland", "wayland-client", "wayland-egl"]
+default = ["x11", "wayland"]
 
 [dependencies]
 lazy_static = "1.3"
-winit = "0.23.0"
+winit = { version = "0.23.0", default-features = false }
 
 [target.'cfg(target_os = "android")'.dependencies]
 android_glue = "0.2"
@@ -54,10 +57,10 @@ parking_lot = "0.11"
 
 [target.'cfg(any(target_os = "linux", target_os = "freebsd", target_os = "dragonfly", target_os = "netbsd", target_os = "openbsd"))'.dependencies]
 osmesa-sys = "0.1"
-wayland-client = { version = "0.28", features = ["dlopen"] }
-wayland-egl = "0.28"
+wayland-client = { version = "0.28", features = ["dlopen"], optional = true }
+wayland-egl = { version = "0.28", optional = true }
 libloading = "0.6.1"
 glutin_egl_sys = { version = "0.1.4", path = "../glutin_egl_sys" }
-glutin_glx_sys = { version = "0.1.6", path = "../glutin_glx_sys" }
+glutin_glx_sys = { version = "0.1.6", path = "../glutin_glx_sys", optional = true }
 parking_lot = "0.11"
 log = "0.4"
--- a/glutin/src/platform_impl/unix/mod.rs
+++ b/glutin/src/platform_impl/unix/mod.rs
@@ -6,17 +6,25 @@
     target_os = "openbsd",
 ))]
 
+#[cfg(not(any(feature = "x11", feature = "wayland")))]
+compile_error!(
+    "at least one of the 'x11' or 'wayland' features must be enabled"
+);
+
 mod wayland;
 mod x11;
 
+#[cfg(feature = "x11")]
 use self::x11::X11Context;
 use crate::api::osmesa;
 use crate::{
     Api, ContextCurrentState, ContextError, CreationError, GlAttributes,
     NotCurrent, PixelFormat, PixelFormatRequirements, Rect,
 };
+#[cfg(feature = "x11")]
 pub use x11::utils as x11_utils;
 
+#[cfg(feature = "x11")]
 use crate::platform::unix::x11::XConnection;
 use crate::platform::unix::EventLoopWindowTargetExtUnix;
 use winit::dpi;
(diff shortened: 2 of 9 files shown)


glutin/Cargo.toml
```diff
@@ -15,10 +15,13 @@ features = ["serde"]
...
 [features]
 serde = ["winit/serde"]
+x11 = ["winit/x11", "glutin_glx_sys"]
+wayland = ["winit/wayland", "wayland-client", "wayland-egl"]
+default = ["x11", "wayland"]
 
 [dependencies]
 lazy_static = "1.3"
-winit = "0.23.0"
+winit = { version="0.23.0", default-features = false }
```
chrisduerr, line 24:
```suggestion
winit = { version = "0.23.0", default-features = false }
```

glutin/src/platform_impl/unix/mod.rs
```diff
@@ -6,17 +6,22 @@
     target_os = "openbsd",
 ))]
 
+#[cfg(feature = "wayland")]
 mod wayland;
```
chrisduerr, line 10: For things like glx you define the feature inside the module while defining it outside the module here. Would probably be best to be consistent. I generally prefer having it as external annotation at the top of the module.

chrisduerr (changes requested):

glutin/src/platform_impl/unix/mod.rs
```diff
@@ -97,24 +108,28 @@ impl Context {
...
             let gl_attr = gl_attr.clone().map_sharing(|ctx| match *ctx {
                 Context::X11(ref ctx) => ctx,
                 _ => unreachable!(),
             });
-            x11::Context::new(wb, el, pf_reqs, &gl_attr)
-                .map(|(win, context)| (win, Context::X11(context)))
+            return x11::Context::new(wb, el, pf_reqs, &gl_attr)
+                .map(|(win, context)| (win, Context::X11(context)));
         }
+        unreachable!()
```
chrisduerr, line 132: This is **not** unreachable.

glutin/src/platform_impl/unix/mod.rs
```diff
@@ -133,29 +148,37 @@ impl Context {
...
             let gl_attr = gl_attr.clone().map_sharing(|ctx| match *ctx {
                 Context::X11(ref ctx) => ctx,
                 _ => unreachable!(),
             });
-            x11::Context::new_headless(&el, pf_reqs, &gl_attr, size)
-                .map(|ctx| Context::X11(ctx))
+            return x11::Context::new_headless(&el, pf_reqs, &gl_attr, size)
+                .map(|ctx| Context::X11(ctx));
         }
+        unreachable!()
```
chrisduerr, line 173: This is **not** unreachable.

chrisduerr (changes requested):

glutin/src/platform_impl/unix/mod.rs
ogoffart: Is it not?
I am under the impression that it was either x11 or wayland and that this should be unreachable.
I've added a if just for symmetry.  But   `is_x11()` is defined as `!is_wayland()`.
What are you suggesting to do instead?  Is a `panic!` better than a `unreachable!` ?

chrisduerr: What stops you from compiling it with neither x11 or wayland?

ogoffart: Right, this could happen if, for example, winit was compiled with wayland support (via another dependency), but glutin was not compiled with wayland support, and that we run on wayland anyway.

So the question is what do you think should happen in that case? `panic!("glutin was not compiled to support this backend")` ?

chrisduerr: I'd advocate for an error during compilation, not runtime. See https://doc.rust-lang.org/std/macro.compile_error.html.

glutin/src/platform_impl/unix/mod.rs
```diff
@@ -129,7 +132,7 @@ impl Context {
             return x11::Context::new(wb, el, pf_reqs, &gl_attr)
                 .map(|(win, context)| (win, Context::X11(context)));
         }
-        unreachable!()
+        panic!("Unsupported backend")
```
chrisduerr, line 135: Now unreachable would have been the correct choice! Since it should be actually unreachable now. Though I think you can still pass messages to it so feel free to add the "at least one of.." message there too. But just unreachable should be fine too.

chrisduerr (changes requested):

glutin/src/platform_impl/unix/mod.rs
ogoffart: I just tested that and it is possible to reach this if the winit backend is compiled with more features:

To reproduce, imagine this in the application's Carto.toml
```
winit = { version = "0.23" }  # default feature enabled so wayland is enabled when building winit
glutin = { version = "0.25.0", default-features = false, feature="x11"}
```
And then run the resulting program  on wayland

chrisduerr: Ah, because at that point you're running into a runtime check, yeah.

I'd like to print a more specific error though, since `Unsupported backend` is far too generic to be helpful. Maybe something more like `glutin was not compiled with support for this display server`, just as a rough suggestion. Maybe that's a bit long.

chrisduerr (changes requested):

chrisduerr (approved):
Thanks for your contribution.

@kchibisov Do you know of anyone else who can review this? Could you maybe check that Wayland still works at least? Otherwise I'd just merge it myself.

kchibisov:
>@kchibisov Do you know of anyone else who can review this? Could you maybe check that Wayland still works at least? Otherwise I'd just merge it myself.

I'm not aware that anyone can review this, however given that it's basically adding a bunch of cfg's anyone can review it. it works on Wayland, so if it works on X11 (can't test it right now). You're free to merge it.

Issue #899, opened 2017-06-29
Context sharing not implemented for many platforms


mitchmindtree:
This is a tracking issue for the state of context display list sharing for each backend. 

- [x] Linux | Wayland | EGL
- [x] Linux | X11 | EGL
- [x] Linux | X11 | GLX
- [x] Windows | WGL
- [x] Windows | EGL
- [x] macOS
- [ ] iOS
- [ ] Android

ZeGentzy:
Linux support + windows w/ egl has been added in #1101.

kchibisov:
Fixed in #1439.

PR #1333, opened 2020-10-13
Allow context sharing for headless contexts on MacOS

bschwind:

Original issue: #899

This was originally implemented in https://github.com/rust-windowing/glutin/pull/1132, however sharing support was never added for headless contexts. This is essentially the same change as before, just in `new_headless`.

I tested this in a company application which uses multi-threaded OpenGL contexts with texture sharing, and this change fixed an issue I was having. However, in `master` before this change, `cargo run --example sharing` just showed a black screen. It also showed a black screen after this change, so I can't claim the example runs correctly. There's some other issue for MacOS at play for that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+- On macOS, add support for sharing of headless contexts.
+
 # Version 0.28.0 (2021-12-02)
 
 - On Windows, fixed a panic for headless contexts because of active drag-and-drop (OleInitialize failed! Result was: `RPC_E_CHANGED_MODE`)
--- a/glutin/src/platform_impl/macos/mod.rs
+++ b/glutin/src/platform_impl/macos/mod.rs
@@ -150,6 +150,8 @@ impl Context {
     ) -> Result<Self, CreationError> {
         let gl_profile = helpers::get_gl_profile(gl_attr, pf_reqs)?;
         let attributes = helpers::build_nsattributes(pf_reqs, gl_profile)?;
+        let share_ctx = gl_attr.sharing.map_or(nil, |c| *c.get_id());
+
         let context = unsafe {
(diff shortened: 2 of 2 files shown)


strohel:
Hi, I've rebased the PR on current `master`, added a Changelog entry, and @bschwind was kind to test it on a Mac.

Could somebody with permissions please approve the CI run and consider merging this? Thanks!

maroider:
I don't have the means to test this myself, so I'd be merging partially blind right now.
As you said, @bschwind, the PR is pretty much identical to #1132, so I'd expect it to work based on that, but I'd really like more than one person to confirm that the fix works.

If either of @mcginty @xiaopengli89 could get a repro on the bug and fix (I see you two have given this PR a "👍"), then that would help soothe my worries.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On Windows, fixed a panic for headless contexts because of active drag-and-drop (OleInitialize failed! Result was: `RPC_E_CHANGED_MODE`)
+- Context sharing of headless contexts is now supported on macOS.
```
maroider, line 4:
```suggestion
- On macOS, add support for sharing of headless contexts.
```

strohel:
Full disclosure: me, @bschwind and @mcginty work together in the same company (tonari).

CHANGELOG.md
strohel: Thanks, that is clearer, applied.

bschwind:
Hi @maroider I've rebased on the latest `master` branch and can confirm this still works on MacOS, though I understand if you want others to test too.

kchibisov:
The context sharing is enabled on macOS, however headless was removed in some sense, but you can create a context without a window, so it might be fine?

bschwind:
@kchibisov I see things have changed a bit since I submitted this PR.

I'll test out shared contexts on MacOS when I get the chance. If I can create a context without a window, I think that will suffice for my usage.

PR #1338, opened 2020-11-21
Support for unreleased winit @ (0.24.0?)

abowerman:

See upcoming change here: https://github.com/rust-windowing/winit/pull/1738


chrisduerr (changes requested):
Please don't just open PRs that provide no benefit and will just cause maintenance overhead.

There is no new winit version and glutin will not be built against a git version.

PR #1341, opened 2020-12-04
Add Apple Silicon support

samdoiron:
Apple Silicon support was added to winit in the latest (unreleased)
master version. It also renamed some types, so we'll need to do the
same here to keep compatible.

The dependency update won't be the `git` version ultimately, of course,
but I'm adding it in for now in case anyone else wants to know how
to build on M1 in the meantime.

### DIFF
--- a/glutin/Cargo.toml
+++ b/glutin/Cargo.toml
@@ -21,7 +21,7 @@ default = ["x11", "wayland"]
 
 [dependencies]
 lazy_static = "1.3"
-winit = { version = "0.23.0", default-features = false }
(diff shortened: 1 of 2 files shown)


chrisduerr (changes requested):
How does this add Apple silicon support? It just updates winit to the latest version.

Please do not update winit unless a new version is actually released. That's not conductive and just wastes a lot of time for reviewers.

samdoiron:
Understood, apologies. As described in the commit support was added in winit in the unreleased version. I did not intend this to ship nor be reviewed as is, hence its draft status.

Happy to leave this closed of course until the release is cut.

PR #1342, opened 2020-12-10
Bump winit to 0.24.0

chrisduerr:


glutin/src/platform/mod.rs
```diff
@@ -30,7 +30,7 @@ pub mod desktop {
         target_os = "netbsd",
         target_os = "openbsd",
     ))]
-    pub use winit::platform::desktop::*;
+    pub use winit::platform::run_return::*;
```
kchibisov, line 33: You should properly update this method, since it should include `android`. Also you should update docs, and rename it to expose that it just brings winit's run_return extension.

kchibisov (changes requested):
Also, while we're at it, could we fix the warnings. I can send a patch myself if you don't want to.

chrisduerr:
> Also, while we're at it, could we fix the warnings. I can send a patch myself if you don't want to.

I intentionally didn't fix the warnings because the code needs some serious cleanup in that regard. It's not just a simple fix.

kchibisov (approved):

PR #1343, opened 2020-12-10
Usa stable version of rustfmt for formatting

kchibisov:
This changes the way rustfmt runs, since right now it uses nightly
versions and it's not properly checked on CI, while we can use nightly
on CI, the stable rustfmt should work fine.

Also the 'small' heuristics was changed to 'Max' to make code more
compact.

### DIFF
--- a/glutin/src/api/android/mod.rs
+++ b/glutin/src/api/android/mod.rs
@@ -1,12 +1,8 @@
 #![cfg(target_os = "android")]
 
-use crate::api::egl::{
-    Context as EglContext, NativeDisplay, SurfaceType as EglSurfaceType,
-};
+use crate::api::egl::{Context as EglContext, NativeDisplay, SurfaceType as EglSurfaceType};
 use crate::CreationError::{self, OsError};
-use crate::{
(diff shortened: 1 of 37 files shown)


chrisduerr (approved):
I'd assume all options used by glutin are stable in rustfmt? If that's the case, there's no reason to use nightly since that's just more unreliable.

kchibisov:
>I'd assume all options used by glutin are stable in rustfmt? If that's the case, there's no reason to use nightly since that's just more unreliable.

Previous options required nightly actually, so that's why your CI passed, while rustfmt was passing for me, however I don't really trust github actions when it comes to rust toolchains and I don't want everything fail due to rustfmt or really require nightly if it was barely used in a first place.

chrisduerr:
> I don't really trust github actions when it comes to rust toolchains

Github Actions has absolutely nothing to do with rust toolchains, it's just a linux system with a rustup install. You're free to pick whatever you want.

> I don't want everything fail due to rustfmt or really require nightly if it was barely used in a first place.

If you set it up properly, it won't. Just like it won't for Alacritty. Removing options and making the formatting style worse just so you don't have to bother with setting things up properly seems like an awful idea to me.

PR #1345, opened 2020-12-28
Fix broken link in documentation

atsheehan:
The link to the `Window` struct on the [glutin index page is
broken][0]. It looks for the struct in the top-level module but
it's defined in the `window` module.

This change prefixes the link with the window module name.

[0]: https://docs.rs/glutin/0.26.0/glutin/


chrisduerr (approved):
Works @murarth.

PR #1350, opened 2021-03-18
fix: share_context in headless mode (macos)

xiaopengli89:
I added `share_ctx` parameter to `initWithFormat_shareContext_` method.


xiaopengli89:
related #1333

PR #1354, opened 2021-05-19
Update winit to 0.25 and cocoa to 0.24

alexheretic:
Updates _winit_ to the latest version: https://github.com/rust-windowing/winit/releases/tag/v0.25.0 I also updated _cocoa_ to match the version used in _winit_ (closes #1346).

It would be great to see a new glutin release to go with this.

- ~Tested on all platforms changed~ I've tested Linux examples.
- ~Compilation warnings were addressed~ There are loads and are not related.
- [x] `cargo fmt` has been run on this branch
- [x] `cargo doc` builds successfully
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- ~Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior~
- ~Created or updated an example program if it would help users understand this functionality~

PR #1355, opened 2021-05-20
Update libloading

ogoffart:
PR #1354 updates winit, I'm also taking the opportunity to update another outdated dependency.


maroider:
Thanks

PR #1356, opened 2021-05-21
Fix CI warnings

maroider:
- [x] ~~Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users~~ This shouldn't be noticeable for downstream users
- [x] ~~Created or updated an example program if it would help users understand this functionality~~

### DIFF
--- a/glutin/src/lib.rs
+++ b/glutin/src/lib.rs
@@ -84,15 +84,6 @@ extern crate lazy_static;
 #[cfg(any(target_os = "macos", target_os = "ios"))]
 #[macro_use]
 extern crate objc;
-#[cfg(any(
-    target_os = "linux",
-    target_os = "dragonfly",
-    target_os = "freebsd",
-    target_os = "netbsd",
-    target_os = "openbsd",
-))]
-#[macro_use]
-extern crate log;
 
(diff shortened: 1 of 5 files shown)


glutin/src/lib.rs
```diff
@@ -394,28 +381,23 @@ pub enum ContextError {
...
+            ContextError::ContextLost => "Context lost".to_string(),
+            ContextError::FunctionUnavailable => "Function unavailable".to_string(),
         }
     }
 }
 
 impl std::fmt::Display for ContextError {
     fn fmt(&self, formatter: &mut std::fmt::Formatter) -> Result<(), std::fmt::Error> {
-        formatter.write_str(self.to_string())
+        formatter.write_str(&self.to_string())
```
francesca64, line 396: I know this is just a warning cleanup, but since it's touching this anyway... if we move the work from `to_string` into `Display::fmt`, we can eliminate `to_string` and use `ToString::to_string` instead.

francesca64 (changes requested):
Thanks! Looks great.

francesca64 (approved):
Thanks!

PR #1357, opened 2021-05-22
Fix ios glutin example viewport size

mkieblesz:
Adding launch screen image fixes the issue. Basically without it viewport does not match actual screen size (there are black edges to on the top and bottom of the screen). Here is an example: https://stackoverflow.com/questions/32641240/ios-9-xcode-7-application-appears-with-black-bars-on-top-and-bottom

- [x] ~~Compilation warnings were addressed~~
- [x] ~~`cargo fmt` has been run on this branch~~
- [x] ~~`cargo doc` builds successfully~~
- [x] ~~Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users~~


Nazariglez:
Hey! I tried this fix using the example on the ios_example folder, and it seems that it doesn't work, I cannot see the triangle only the clear color. The example works for you?

kchibisov:
There's no ios for now.

PR #1379, opened 2021-12-01
Update contact links

maroider:
- ~~Tested on all platforms changed~~
- ~~Compilation warnings were addressed~~
- ~~`cargo fmt` has been run on this branch~~
- ~~`cargo doc` builds successfully~~
- ~~Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users~~
- ~~Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior~~
- ~~Created or updated an example program if it would help users understand this functionality~~

See rust-windowing/winit#1966

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,9 +15,8 @@ glutin = "0.27.0"
 
 Join us in any of these:
 
-[![Freenode](https://img.shields.io/badge/freenode.net-%23glutin-red.svg)](http://webchat.freenode.net?channels=%23glutin&uio=MTY9dHJ1ZSYyPXRydWUmND10cnVlJjExPTE4NSYxMj10cnVlJjE1PXRydWU7a)
-[![Matrix](https://img.shields.io/badge/Matrix-%23Glutin%3Amatrix.org-blueviolet.svg)](https://matrix.to/#/#Glutin:matrix.org)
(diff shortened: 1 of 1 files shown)


maroider:
It was suggested that we may want separate rooms/channels for winit and glutin.

I don't really have any strong feelings one way or the other.

chrisduerr:
> It was suggested that we may want separate rooms/channels for winit and glutin.
>
> I don't really have any strong feelings one way or the other.

Honestly with the amount of activity it's probably easier to just keep them combined. The naming itself might be a bit discouraging but people will figure it out.

kchibisov:
>It was suggested that we may want separate rooms/channels for winit and glutin.

There's no reason to do so, glutin isn't a separated project to be fair, since it reexports winit and exposes it in its public API to some point. But that doesn't really matter, all rust-windowing projects should be discussed in one channel for now due to low number of community members. Could rename '#winit' to '#rust-windowing'.

madsmtm (approved):
I think I suggested the separate channels, but I agree that we should keep them together.

These links are to the `#rust-windowing` "room", so I think we can safely merge this PR whether or not we should rename the `#winit` channel.

PR #1382, opened 2021-12-08
Fix sRGB support for WGL and EGL

derezzedex:
This PR fixes sRGB support for the `wgl` and `egl` backends (which I encountered while trying sRGB with OpenGL ES).

### DIFF
--- a/glutin_egl_sys/build.rs
+++ b/glutin_egl_sys/build.rs
@@ -38,6 +38,7 @@ fn main() {
                 "EGL_EXT_platform_wayland",
                 "EGL_EXT_platform_device",
                 "EGL_KHR_swap_buffers_with_damage",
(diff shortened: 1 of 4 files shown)


glutin_egl_sys/build.rs
```diff
@@ -38,6 +38,7 @@ fn main() {
                 "EGL_EXT_platform_wayland",
                 "EGL_EXT_platform_device",
                 "EGL_KHR_swap_buffers_with_damage",
+                "EGL_KHR_gl_colorspace",
```
kchibisov, line 41: What will happen with new EGL path when this extension isn't supported?

kchibisov (changes requested):
Should add a note about that in changelog entry.

kchibisov:
That should be fixed in #1439.

Issue #1262, opened 2020-01-08
[Linux] Window creation fails on platforms without VSYNC extensions


syrel:
The pull request https://github.com/rust-windowing/glutin/pull/1253 prevents Glutin from creating windowed contexts on systems that do not have vsync extensions. It fails even if `vsync` is set to `false`.

```
CreationErrors([NoAvailablePixelFormat, OsError("Couldn't find any available vsync extension")])
```
In my opinion instead of throwing an error at [mod.rs#L459](https://github.com/2xsaiko/glutin/blob/3acad49efa2325230d9479e665104c5d68600719/glutin/src/api/glx/mod.rs#L459 ) we shall ignore vsync if there are no available extensions.

Here is an example of such limited system:

```
name of display: :0
display: :0  screen: 0
direct rendering: Yes
server glx vendor string: Parallels Inc
server glx version string: 1.4
server glx extensions:
    GLX_ARB_get_proc_address, GLX_EXT_texture_from_pixmap, GLX_SGIX_fbconfig
client glx vendor string: Parallels Inc
client glx version string: 1.4
client glx extensions:
    GLX_ARB_get_proc_address, GLX_EXT_texture_from_pixmap, GLX_SGIX_fbconfig
GLX version: 1.4
GLX extensions:
    GLX_ARB_get_proc_address, GLX_EXT_texture_from_pixmap, GLX_SGIX_fbconfig
OpenGL vendor string: Parallels and ATI Technologies Inc.
OpenGL renderer string: Parallels using AMD Radeon Pro 460 OpenGL Engine
OpenGL version string: 2.1 ATI-3.2.24
OpenGL shading language version string: 1.20
OpenGL extensions:
    GL_ARB_color_buffer_float, GL_ARB_depth_buffer_float, GL_ARB_depth_clamp, 
    GL_ARB_depth_texture, GL_ARB_draw_buffers, 
    GL_ARB_draw_elements_base_vertex, GL_ARB_draw_instanced, 
    GL_ARB_fragment_program, GL_ARB_fragment_program_shadow, 
    GL_ARB_fragment_shader, GL_ARB_framebuffer_object, 
    GL_ARB_framebuffer_sRGB, GL_ARB_half_float_pixel, 
    GL_ARB_half_float_vertex, GL_ARB_imaging, GL_ARB_instanced_arrays, 
    GL_ARB_map_buffer_range, GL_ARB_multisample, GL_ARB_multitexture, 
[...]

PR #1387, opened 2022-01-27
Fix error on systems without vsync support (#1262)

ch-sy:
If vsync was not requested and it is not supported on the system, then this feature can safely be ignored. This makes the program runnable on systems without vsync support like the Raspberry Pi Zero. 


Ar37-rs:
Any update? why not being merged yet?

kchibisov (approved):

PR #1390, opened 2022-02-06
Add `buffer_age` method on WindowedContext

kchibisov:
This uses `EGL_EXT_buffer_age` and `GLX_EXT_buffer_age` API to query
window's back buffer age.


glutin/src/windowed.rs
```diff
@@ -170,6 +170,16 @@ impl<W> ContextWrapper<PossiblyCurrent, W> {
...
+
+    /// Query the underlying surface back's buffer age.
+    ///
+    /// Returns `Some(n)` where `n` is the number of frames elapsed since it was most recently
+    /// drawn.
+    ///
+    /// Return `None` when query fails.
+    pub fn buffer_age(&self) -> Option<u32> {
+        self.context.context.buffer_age()
+    }
```
chrisduerr, line 182: As far as I understand it, a buffer age of `0` is basically to be handled like an empty buffer, right?

Is there anything downstream could do with the `Option` other than just `unwrap_or(0)`? Like realistically how is someone going to handle that other than just assuming they need to fully damage anyway? Seems like we can get rid of that complexity?

chrisduerr (changes requested):

glutin/src/windowed.rs
kchibisov: Yeah, I think returning always `0` makes sense. Was also thinking about doing so while implementing, but went for `Option` instead, don't know why.

PR #1394, opened 2022-04-13
Fix CI failed to run

Ar37-rs:
- [ ✅] Tested on all platforms changed
- [ ✅] `cargo fmt` has been run on this branch
- [ ✅] `cargo doc` builds successfully


kchibisov:
Nah, I think I'll queue this with buffer age patch, since it touches the same file.

Issue #1393, opened 2022-04-13
[BUG] Assertion `!xcb_xlib_threads_sequence_lost' failed (glutin <= v0.28.0)


Ar37-rs:
reopen this issue: the fixes is here: https://github.com/rust-windowing/glutin/pull/1387

Issue #1034, opened 2018-06-24
Assertion `!xcb_xlib_threads_sequence_lost' failed


kuviman:
Trying to run on Ubuntu 14.04

```
kuviman@UbuntuBox:~/Temp/glutin$ cargo run --example window
    Finished dev [unoptimized + debuginfo] target(s) in 0.21s
     Running `target/debug/examples/window`
[xcb] Unknown sequence number while processing queue
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
window: ../../src/xcb_io.c:274: poll_for_event: Assertion `!xcb_xlib_threads_sequence_lost' failed.
Aborted (core dumped)
```

kuviman:
According to `Cargo.lock`, I'm on `winit` 0.15.1, still have the issue.

I'll take a look at the linked issue, maybe it should be reopened.

kuviman:
Running [multithread_window](https://github.com/azriel91/multithread_window) with winit 0.15.1 is ok, can't reproduce the issue with that.

Running glutin's `--example window` reproduces this reliably though.

Also, it runs fine on my Arch Linux laptop, but reproduces every time with both Ubuntu 14.04 and 18.04 in VirtualBox, if that helps.

kuviman:
Some more details: my VM was set to use only 1 CPU, when I changed it to more than that, it runs successfully like 70% of times.

azriel91:
Hiya, I accidentally created a way to reproduce this issue. I've put it on the `glutin-xlib-sequence-lost` branch of [multithread_window](https://github.com/azriel91/multithread_window). Cut and paste commands:

```bash
git clone -b glutin-xlib-sequence-lost --depth 1 git@github.com:azriel91/multithread_window.git
cd multithread_window
cargo run # to compile it

# Limited it to 5 because it can be hard to stop when the windows keep popping up.
for i in {1..5}; do target/debug/multithread_window; done
```

The sample output:

```
$ for i in {1..5}; do target/debug/multithread_window; done
Spawned threads.
Xlib: sequence lost (0x1015e > 0x162) in reply type 0x0!
[xcb] Unknown sequence number while processing queue
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
multithread_window: ../../src/xcb_io.c:259: poll_for_event: Assertion `!xcb_xlib_threads_sequence_lost' failed.
[...]

PR #1395, opened 2022-04-15
Ignore vsync safely on unsupported platforms.

Ar37-rs:
- [✅] Tested on all platforms changed
- [✅] Compilation warnings were addressed
- [✅] `cargo fmt` has been run on this branch
- [✅] `cargo doc` builds successfully
- [✅] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users

improvement of this PR: https://github.com/rust-windowing/glutin/pull/1387
thanks to: @ch-sy

### DIFF
--- a/glutin/src/api/glx/mod.rs
+++ b/glutin/src/api/glx/mod.rs
@@ -364,7 +364,13 @@ impl<'a> ContextPrototype<'a> {
     }
 
     pub fn finish(self, window: ffi::Window) -> Result<Context, CreationError> {
-        let glx = GLX.as_ref().unwrap();
+        let glx = match GLX.as_ref() {
+            Some(glx) => glx,
+            _ => {
+                return Err(CreationError::OsError("Can't get GLX as reference.".into()));
+            }
+        };
+
         let (extra_functions, context) = self.create_context()?;
 
         // vsync
@@ -410,9 +416,7 @@ impl<'a> ContextPrototype<'a> {
                 extra_functions.SwapIntervalSGI(swap_mode);
             }
         } else if self.opengl.vsync {
-            return Err(CreationError::OsError(
-                "Couldn't find any available vsync extension".to_string(),
-            ));
+            eprintln!("Couldn't find any available vsync extension on this platform.")
         }
 
         Ok(Context {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+- Ignore vsync safely on unsupported platforms (e.g: llvmpipe software render).
+
 # Version 0.28.0 (2021-12-02)
 
 - On Windows, fixed a panic for headless contexts because of active drag-and-drop (OleInitialize failed! Result was: `RPC_E_CHANGED_MODE`)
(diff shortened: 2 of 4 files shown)


Ar37-rs:
it really useful for anyone like myself who's running their GL apps on software render mode (e.g: mesa llvmpipe) on their VM, so the next version of glutin there will be no more such thing as this issue anymore https://github.com/rust-windowing/glutin/issues/1393
thank you @kchibisov

glutin/src/api/glx/mod.rs
```diff
@@ -410,9 +416,7 @@ impl<'a> ContextPrototype<'a> {
                 extra_functions.SwapIntervalSGI(swap_mode);
             }
         } else if self.opengl.vsync {
-            return Err(CreationError::OsError(
-                "Couldn't find any available vsync extension".to_string(),
-            ));
+            eprintln!("Couldn't find any available vsync extension on this platform.")
```
kchibisov, line 419: Should we `log` it?

I'm also not sure if we've asked for `vsync`?

I think you want to have some fallback, right? Could change API `with_vsync(Option<bool>)`, and when passing `None` it means you don't care about whether vsync is used or not?

kchibisov (changes requested):

glutin/src/api/glx/mod.rs
Ar37-rs: exactly, for fallback purpose.  on third party apps the end user have no control over. like egui for example, it uses vsync by default, and there's no way to change the fallback config on the fly. if the GPU doesn't support vsync the apps will be crashing like so many issue has been occured like this: https://github.com/rust-windowing/glutin/issues/1393

Ar37-rs: @kchibisov After this PR being merged i believe this issue: https://github.com/rust-windowing/glutin/issues/1034  can be closed also.

kchibisov: As I said, would you mind changing the API to `Option<bool>`? I think it'll make sense for the time being, since some applications want to force vsync.

Ar37-rs: there's no need to change bool to Option<bool>, because the logic uses statement if {..} else if {..}. so if the GPU driver support the vsync extension it will use vsync, otherwise the error will be printed

Ar37-rs:
@kchibisov if i changed the bool to ```Option<bool>``` obviously it will be breaking change, and i don't want to do that since the only issue which cause the apps with glutin dependecies crashing so often is that the current version of glutin inability to ignore vsync on the machine with no HW acceleration support. after modified glutin the same as on this PR i swear i've tested on windows with Intel GPU HW acceleration and linux with software render (llvmpipe). both are working fine, no more crashing even on software render mode unless with a specific GL option, let's say using glow crate with multisampling > 4 (of course, software render can't comprehend pixelFormat with multisampling large than that). and GPU HW acceleration there's no issue at all, thank you. @kchibisov would you mind to merge this PR? otherwise i'll close this myself.

kchibisov:
>@kchibisov if i changed the bool to Option<bool> obviously it will be breaking change

I'm fine with doing breaking change, given that the next winit release is a glutin breaking change anyway, so given that the next release will be breaking anyway, I'm all for adding that kind of API.

PR #1397, opened 2022-04-16
Return Err instead of panic on surfaceless GLX contexts on Linux

unlimitedbacon:
Returns an `Err` instead of panicking in this specific case when surfaceless context creation fails on Linux. This allows my program to gracefully fall back to an OsMesa context when it fails.


kchibisov (changes requested):
Could you add changelog entry and run rustfmt. Other than that it should be good to go.

unlimitedbacon:
Done

kchibisov (approved):

MarijnS95 (approved):

PR #1385, opened 2021-12-28
Error compiling to Android with cargo apk #1307 

Kaiser1989:
- [x] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [x] `cargo doc` builds successfully
- [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [x] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [ ] Created or updated an example program if it would help users understand this functionality

This merge request addresses the endless unresolved build errors when trying to compile with cargo apk for android.
https://github.com/rust-windowing/glutin/issues/1307

There is an open merge request on winit crate:
https://github.com/rust-windowing/winit/pull/2118

If that request is merged, we can finally resolve this bug. But before merging, we need to adjust the winit version in cargo toml. I just want to created this request to address this also on winit side.

Working examples with both fixes can be found here:
https://github.com/Kaiser1989/rust-android-example
https://github.com/Kaiser1989/game-gl

PR #1398, opened 2022-04-22
Use non-static `StructGenerator` on Android

jamienicol:
This is an attempt to fix the current compilation errors on Android. I know there is already #1313 and #1385, but I thought if I try to just tackle part of the problem, and do so properly rather than just commenting out the broken code, it might be easier to push things forward.

On of the compilation errors we get on android is:
```
error[E0615]: attempted to take value of method `SwapBuffersWithDamageKHR` on type `&api::egl::egl::Egl`
   --> glutin/src/api/egl/mod.rs:621:17
    |
621 |         if !egl.SwapBuffersWithDamageKHR.is_loaded() {
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^ method, not a field
```

This is due to using the `StaticStructGenerator` rather than `StructGenerator`. From my testing using `StructGenerator` appears to work correctly, so this patch switches us to use that on Android. Are there any issues with using StructGenerator on Android that I'm not aware about?

Closes #1407


kchibisov:
I don't know anything about using glutin on android, but the patch looks good to me given that it just switches to what other unix platforms do here, and I think libloading should work on android fine. Have you tried running something with glutin on android with your patch applied?

Though, if you're in doubt, I'd suggest asking other folks using glutin on android, since I think there're some in gamedev engines communities.

jamienicol:
Note this doesn't fix all of the compilation errors on Android. But yes, I am successfully using Glutin with this patch and another to fix the rest, and our app runs perfectly.

Added a changelog entry.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
+
+- On Android, switched from StaticStructGenerator to StructGenerator, fixing some compilation errors.
```
MarijnS95, line 3:
```suggestion
- On Android, switched from `StaticStructGenerator` to `StructGenerator`, fixing some compilation errors.
```

MarijnS95 (approved):
> but I thought if I try to just tackle part of the problem, and do so properly rather than just commenting out the broken code, it might be easier to push things forward.

Massive, massive thanks for this. It just makes PRs so much easier to review, have shorter life/churn-time, and is overal a lot friendlier for maintainers. If we go ahead with this and a followup we can finally start closing all the other noise :tada: 

Regarding the contents, this seems like a sensible solution. It removes Android-specific bloat (`libloading` definitely works on Android, it's "just Unix") and statically linking against an optional extension function is wrong (looking at you, recent PRs that just comment out `is_loaded()`). Then again I don't know if Android devices are required to ship with `EGL_KHR_swap_buffers_with_damage`, seems like they provide some fallback that diverts into `eglSwapBuffers`: https://android.googlesource.com/platform/frameworks/native/+/a894d08%5E%21/ (so perhaps it's possible to just always static-link into this one).

The only alternative solution I can think of is statically _and dynamically_ link, and make a clear decision in every place which one to use. Perhaps the EGL bindings generator should be able to make this distinction by generating the core statically and the extensions dyanamically.

MarijnS95:
> But yes, I am successfully using Glutin with this patch and another to fix the rest, and our app runs perfectly.

Do you want to open a PR [with that](https://github.com/jamienicol/glutin/commit/03285da9c14ec56296c2400c781d2c32b80d745a) as soon as we merge this one? (If you fix the changelog conflict I'll click the button :smile:). Then we can finally close a whole bunch of noise on lots of repositories that all tried to address the compiler errors in mysterious ways.

I've cooked up a little Android example _within the `glutin` repo_ on top of your branch to demonstrate it working: https://github.com/MarijnS95/glutin/pull/new/wr. We could possibly combine those.

Finally we'll have to think up what to do about the `Suspended`/`Resumed` situation. Somewhat intentionally `winit` models these as events instead of callbacks, but they're not cross-platform and I'd hope to change that some day (to make all windows work the same way). The way I implemented it in the example works, but is expensive as we only need to recreate the surface and not the entire GL context. Let's discuss that elsewhere though, I have a few ideas :+1:

MarijnS95:
@kchibisov I tested this too, if you don't mind I'll click the squash+merge button :)

kchibisov:
>@kchibisov I tested this too, if you don't mind I'll click the squash+merge button :)

I'm not familiar with android at all, so I was kind of more style/don't break anything without testing by third parties. Given that you know what android is and can likely test stuff on it, you can go ahead without my approvals on pure android backend stuff.

MarijnS95:
> I'm not familiar with android at all, so I was kind of more style/don't break anything without testing by third parties. Given that you know what android is and can likely test stuff on it, you can go ahead without my approvals on pure android backend stuff.

Lovely, thanks!

jamienicol:
Thanks for merging this @MarijnS95!

> Do you want to open a PR [with that](https://github.com/jamienicol/glutin/commit/03285da9c14ec56296c2400c781d2c32b80d745a) as soon as we merge this one? (If you fix the changelog conflict I'll click the button smile). Then we can finally close a whole bunch of noise on lots of repositories that all tried to address the compiler errors in mysterious ways.

Would be delighted to! I started with just this one as I thought it was the more production ready of the two, given the remaining issues you mentioned with Suspend/Resume. But a PR is a good starting point to discuss those further, and even landing as is would be an improvement on the current situation of it not compiling.

MarijnS95:
> I started with just this one as I thought it was the more production ready of the two

Trunk-based development is perfect here, makes for a much more pleasant experience getting things in, one isolated feature/fix at a time.

> But a PR is a good starting point to discuss those further, and even landing as is would be an improvement on the current situation of it not compiling.

Perhaps it's fine to get your removal in so that it compiles and people manually use `Suspend`/`Resume` events, then I PR my example on top for that, and subsequently finish+PR additional changes to handle those events automatically again, as well as open the door for a "custom" `NativeWindow` (ie. from an Android `Surface`) instead of forcing this to use `ndk_glue::native_window()` in a `NativeActivity`.

Does that sound sensible to you or should we tackle the migration from `set_suspend_callback` to a proper solution at once?

jamienicol:
That sounds very sensible to me. This will allow a lot of people to finally use glutin on android again. Let's not let perfect be the enemy of good.

jamienicol:
Opened #1411

MarijnS95:
> Let's not let perfect be the enemy of good.

For reference I brought this up not to be perfect on the first try, but because I don't feel like removing "valid" (yet ancient) functionality in hopes of adding it back some time in the future. But alas, it's coming anyway so no need to make a fuss about this one.

PR #1402, opened 2022-05-03
Update parking_lot

emilio:
- [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users (N/A)
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior (N/A)
- [ ] Created or updated an example program if it would help users understand this functionality (N/A)


emilio:
The formatting issue is unrelated to my change.

kchibisov:
I'll bump parking_lot together with winit, after winit release, so they'll have a matching version.

Bumping glutin right now isn't really a good thing, so we'd have a missmatch of deps with winit.

PR #1403, opened 2022-05-03
Initial kmsdrm support

StratusFearMe21:

(Warning, this PR lists my PR to winit as a dependency)

See [my PR to winit](https://github.com/rust-windowing/winit/pull/2272)


StratusFearMe21:
Okay, the performance issue is fixed, this PR is ready to be reviewed.

StratusFearMe21:
I'm pretty confident with this PR at this point. As a functionality test, I ported Alacritty to this PR and now it runs straight from the tty https://github.com/StratusFearMe21/alacritty

StratusFearMe21:
The only thing I can think of to add is a cursor, but it's not a must

StratusFearMe21:
Yeah, at this point, I think this is ready to be merged. Im very happy with this PR in its current state.

madsmtm:
FYI: I think https://github.com/rust-windowing/glutin/pull/1435 is going to hit first

kchibisov:
Given that #1439 got merged you'd need to update it, however EGL can create context from DRM device, not sure what needs here?

kchibisov:
I'll close it, since I don't understand what else is required for KMS. I was able to use `glutin` with `gbm` in the context of smithay just fine to draw my triangle. Maybe `vsync` handling needs a bit special treatment, but it could be done separately.

PR #1407, opened 2022-05-21
Fixes android symbol loader not having `is_loaded` (build error)

lattice0:
Fixes https://github.com/rust-windowing/glutin/issues/1307

### DIFF
--- a/glutin/src/api/egl/mod.rs
+++ b/glutin/src/api/egl/mod.rs
@@ -617,7 +617,11 @@ impl Context {
     #[cfg(not(target_os = "windows"))]
     pub fn swap_buffers_with_damage(&self, rects: &[Rect]) -> Result<(), ContextError> {
         let egl = EGL.as_ref().unwrap();
-        
+
+        if !self.swap_buffers_with_damage_supported() {
+            return Err(ContextError::FunctionUnavailable);
+        }
+
         let surface = self.surface.as_ref().unwrap().lock();
         if *surface == ffi::egl::NO_SURFACE {
             return Err(ContextError::ContextLost);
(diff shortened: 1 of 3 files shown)


lattice0:
Hello, before I fix the CI/CD stuff and the missing checkboxes, can I know if something's missing? I think these changes fix the problem. Android seems to not have dynamic loading of dynamically loadable libs so there's no `.is_loaded()` for it.

lattice0:
PS: all testes failed because I pointed to a local winit for my stuff. Gonna fix it soon but currently maybe leave like this so my local project works, is there a problem? just want to keep things simple

glutin/src/api/egl/mod.rs
```diff
@@ -617,11 +617,7 @@ impl Context {
     #[cfg(not(target_os = "windows"))]
     pub fn swap_buffers_with_damage(&self, rects: &[Rect]) -> Result<(), ContextError> {
         let egl = EGL.as_ref().unwrap();
-
-        if !egl.SwapBuffersWithDamageKHR.is_loaded() {
```
MarijnS95, line 621: Perhaps just call `self.swap_buffers_with_damage_supported()` here so that original functionality is retained on non-Android?

glutin/src/api/egl/mod.rs
```diff
@@ -658,12 +654,18 @@ impl Context {
     }
 
     #[inline]
-    #[cfg(not(target_os = "windows"))]
+    #[cfg(all(not(target_os = "windows"), not(target_os = "android")))]
```
MarijnS95, line 657: Note: you can use De Morgan's law to turn this into `not(any(target_os = "windows", target_os = "android"))` :)

glutin/src/api/egl/mod.rs
lattice0: was it calling it originally? I thought not, it was just a check so it calls later on line 638

lattice0: hmmm got it

PR #1408, opened 2022-05-26
Fix `clippy::inherent_to_string_shadow_display` on `CreationError`

MarijnS95:
Clippy throws:

    error: type `CreationError` implements inherent method `to_string(&self) -> String` which shadows the implementation of `Display`
       --> glutin/src/lib.rs:317:5
        |
    317 | /     fn to_string(&self) -> String {
    318 | |         match self {
    319 | |             CreationError::OsError(text) | CreationError::NotSupported(text) => text.clone(),
    320 | |             CreationError::NoBackendAvailable(_) => "No backend is available".to_string(),
    ...   |
    333 | |         }
    334 | |     }
        | |_____^
        |
        = note: `#[deny(clippy::inherent_to_string_shadow_display)]` on by default
        = help: remove the inherent method from type `CreationError`
        = help: for further information visit
        https://rust-lang.github.io/rust-clippy/master/index.html#inherent_to_string_shadow_display

To solve this, the contents of the function - which were only consumed by the `impl Display for CreationError` anyway - have been moved into that implementation.  The function has been revised to only format a nested error if the enum variant has one.  Regardless the default implementation for `source()` returns `None` and doesn't forward to the deprecated `fn cause()` implementation, which we should probably update too.

- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users (internal change not relevant to users)

### DIFF
--- a/glutin/src/lib.rs
+++ b/glutin/src/lib.rs
@@ -337,7 +337,11 @@ impl std::fmt::Display for CreationError {
                 return write!(f, "{}", err);
             }
             CreationError::CreationErrors(ref es) => {
-                return write!(f, "Received multiple errors: `{:?}`", es);
+                writeln!(f, "Received multiple errors:")?;
+                for e in es {
+                    writeln!(f, "\t{}", e)?;
+                }
+                return Ok(());
             }
         })
     }
--- a/glutin/src/lib.rs
+++ b/glutin/src/lib.rs
@@ -313,42 +313,33 @@ impl CreationError {
             _ => CreationError::CreationErrors(vec![Box::new(err), Box::new(self)]),
         }
     }
-
-    fn to_string(&self) -> String {
-        match self {
-            CreationError::OsError(text) | CreationError::NotSupported(text) => text.clone(),
-            CreationError::NoBackendAvailable(_) => "No backend is available".to_string(),
-            CreationError::RobustnessNotSupported => {
-                "You requested robustness, but it is not supported.".to_string()
-            }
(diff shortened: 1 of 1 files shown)


MarijnS95:
I feel like we should remove the `cargo fmt` and `cargo doc` steps from the PR template, the CI already checks this so it's only extra noise to fill in (or I hope contributors look at "CI failed" notifications and make sure their contribution adheres to the standards after initially submitting a PR!).

glutin/src/lib.rs
```diff
@@ -313,42 +313,33 @@ impl CreationError {
...
-
-        if let Some(err) = std::error::Error::source(self) {
-            write!(formatter, ": {}", err)?;
-        }
-        Ok(())
+            CreationError::Window(err) => {
+                return write!(f, "{}", err);
+            }
+            CreationError::CreationErrors(ref es) => {
+                return write!(f, "Received multiple errors: `{:?}`", es);
```
MarijnS95, line 340: Note that this uses the debug implementation on the nested `CreationError`s instead of this `Display` implementation, because `Vec` doesn't implement that. Perhaps we should loop over the nested errors and debug-print them one by one.

kchibisov (approved):

glutin/src/lib.rs
MarijnS95: Sneaked in a little change to call the `Display` implementation recursively here, and spread it out over multiple lines. Let me know if that's okay by you @kchibisov.

Issue #1324, opened 2020-09-29
Use `glutin` to render to `Surface`/`NativeWindow` without `winit`/`ndk_glue`


lattice0:
I want to render OpenGL inside Flutter. I've already made it work in Java and I know it's possible in C++. 

Flutter has this external texture concept where it creates a `SurfaceTexture surfaceTexture` that I can render into. Given the `surfaceTexture`, this is how I render things in Java:

```
eglSurface = egl.eglCreateWindowSurface(eglDisplay, eglConfig, surfaceTexture, null);
egl.eglMakeCurrent(eglDisplay, eglSurface, eglSurface, eglContext)
```

after that, it's all a matter of calling opengl functions like `glTexImage` to send textures, `glDraw` to draw, etc.

Can I use this lib or even glium to simply create an `eglSurface` from a `surfaceTexture` and then draw things on it? 

I don't think `glium` or `glutin` have the power to do anything with a `surfaceTexture`. But maybe I can create the eglSurface (from a surfaceTexture) in java calling from Rust, then it returns and then I can issue opengl calls using glutin? Something like this:

flutter engine calls java code
java code creates the surfaceTexture from flutter call and gives it an id
java starts rust code
rust code starts glutin or glium and fills vertex buffers, etc
rust code calls java with the id and thus java code can create the egl surface and make it current
java code call returns, now glium can issue opengl calls like `glTexImage` and `glDraw`, since the `eglSurface` is current, things will be drawn to it

MarijnS95:
@lattice0 Wow, I didn't expect you to have an issue open on this for well over 1.5 years :)

Is it resolved now that you've submitted a `SurfaceTexture` implementation to `android-ndk-rs`? If so, please document here how you've solved it (little codesnippets using that implementation) for posterity, and feel free to close. Thanks!

lattice0:
@MarijnS95 I didn't solve it yet. I'm implementing `SurfaceTexture` rendering for `glium` right now: https://github.com/lattice0/glium/blob/surface_texture/src/backend/surface_texture/mod.rs, using my SurfaceTexture implementation from android-ndk-rs and also using other small PRs I made, but having lots of OpenGL errors since I'm not an OpenGL expert. 

I also have a Kotlin Android POC of my SurfaceTexture backend: https://github.com/lattice0/glium_surface_texture_poc/blob/master/app/src/main/rust/surface_texture_glium/src/lib.rs

[...]

Issue #1320, opened 2020-08-06
Do not panic when swap_buffer (egl) fails


Kaiser1989:
I notices some bug in running on android. When resuming to the app, sometimes the swap_buffer method is throwing an egl error:

>... /? E/BufferQueueProducer: [rust.rust_android_example/android.app.NativeActivity#0] queueBuffer: BufferQueue has been abandoned
.../rust.rust_android_example E/Surface: queueBuffer: error queuing buffer to SurfaceTexture, -19
.../? E/BufferQueueProducer: [rust.rust_android_example/android.app.NativeActivity#0] dequeueBuffer: BufferQueue has been abandoned
.../rust.rust_android_example E/EGL_emulation: tid 21300: swapBuffers(552): error 0x300d (EGL_BAD_SURFACE)

The problem is located under _glutin/src/api/egl/mod.rs_ line 663:

```
    #[inline]
    pub fn swap_buffers(&self) -> Result<(), ContextError> {
        let egl = EGL.as_ref().unwrap();
        let surface = self.surface.as_ref().unwrap().lock();
        if *surface == ffi::egl::NO_SURFACE {
            return Err(ContextError::ContextLost);
        }

        let ret = unsafe { egl.SwapBuffers(self.display, *surface) };

        if ret == 0 {
            match unsafe { egl.GetError() } as u32 {
                ffi::egl::CONTEXT_LOST => {
                    return Err(ContextError::ContextLost)
                }
                err => panic!(
                    "swap_buffers: eglSwapBuffers failed (eglGetError returned 0x{:x})",
                    err
                ),
            }
        } else {
            Ok(())
        }
    }
```
Calling the _panic!_ kills the app. There is no chance to recover from this state. I noticed, that this problem can easily be fixed by recreating the graphics context. Therefore we need to have a chance to handle that problem:

My simple hack is following:
```
    #[inline]
    pub fn swap_buffers(&self) -> Result<(), ContextError> {
        let egl = EGL.as_ref().unwrap();
        let surface = self.surface.as_ref().unwrap().lock();
        if *surface == ffi::egl::NO_SURFACE {
            return Err(ContextError::ContextLost);
        }

        let ret = unsafe { egl.SwapBuffers(self.display, *surface) };

        if ret == 0 {
            Err(ContextError::ContextLost)
        } else {
            Ok(())
        }
    }
```

[...]

PR #1411, opened 2022-05-27
Switch from android_glue to ndk-glue

jamienicol:
Additionally remove the Android lifecycle handling code, as winit no
longer provides the required callback meaning it does not compile.

This fixes compilation on Android. However, there is a caveat: Glutin
is no longer able to automatically destroy and create the Surface in
response to Android lifecycle events. The application must therefore
ensure they only create a Context following a winit Event::Resumed,
and they destroy the context when receiving an Event::Suspended.


MarijnS95:
> However, there is a caveat: Glutin
> is no longer able to automatically destroy and create the Surface in
> response to Android lifecycle events. The application must therefore
> ensure they only create a Context following a winit Event::Resumed,
> and they destroy the context when receiving an Event::Suspended.

For the record: I'll PR my example as soon as I get home that showcases how this is done. And also for the record, we're working on changes to make this more ergonomic too.

glutin/src/api/android/mod.rs
```diff
@@ -26,27 +25,6 @@ pub struct Context(Arc<AndroidContext>);
...
-impl android_glue::SyncEventHandler for AndroidSyncEventHandler {
-    fn handle(&mut self, event: &android_glue::Event) {
-        match *event {
-            // 'on_surface_destroyed' Android event can arrive with some delay
-            // because multithreading communication. Because of
-            // that, swap_buffers can be called before processing
-            // 'on_surface_destroyed' event, with the native window
-            // surface already destroyed. EGL generates a BAD_SURFACE error in
-            // this situation. Set stop to true to prevent
-            // swap_buffer call race conditions.
```
MarijnS95, line 38: Ah we "fixed" this in `ndk-glue` but it relies on the caller/user (winit) to hold a lock on the `window` and release it after its done all its cleanup upon receiving `WindowDestroyed`. Looks like at least `RawWindowHandle` doesn't do that yet, but `new_windowed()` here calls directly into `ndk_glue::native_window()` where we can implement that :tada: 

It should perhaps be part of this PR though (to store `nwin` inside the context).

MarijnS95: It looks like #1320 reported this too, which we can probably close as "outdated" when this is merged.

glutin/src/api/android/mod.rs
```diff
@@ -57,41 +35,20 @@ impl Context {
     ) -> Result<(winit::window::Window, Self), CreationError> {
         let win = wb.build(el)?;
         let gl_attr = gl_attr.clone().map_sharing(|c| &c.0.egl_context);
-        let nwin = unsafe { android_glue::get_native_window() };
-        if nwin.is_null() {
+        let nwin = ndk_glue::native_window();
+        if nwin.is_none() {
             return Err(OsError("Android's native window is null".to_string()));
         }
```
MarijnS95, line 41:
```suggestion
        let nwin = nwin.ok_or_else(|| OsError("Android's native window is null".to_string()))?;
```

Then you can remove the `.as_ref().unwrap()` below too.

MarijnS95 (approved):
Other than the `unwrap()` nit and request to hold the `NativeWindow` lock I'm totally happy with this PR!

glutin/src/api/android/mod.rs
```diff
@@ -26,27 +25,6 @@ pub struct Context(Arc<AndroidContext>);
 #[derive(Debug)]
 struct AndroidSyncEventHandler(Arc<AndroidContext>);
```
MarijnS95, line 26: Drop this line as well :)

(I'll have CI up for Android shortly, too - that'll catch remaining unused variables)

glutin/src/api/android/mod.rs
MarijnS95: Fwiw the API is currently annoying with an `Option` inside the lock, I'm thinking about changing `ndk-glue` to `parking_lot` which allows mapping - in this case specifically for transmuting - the `Option` out of the `RWLockReadGuard`.

jamienicol: Putting the `RwLockGuard` in `AndroidContext` fails to compile because it is not `Sync`:

```
   Compiling glutin v0.28.0 (/home/jamie/src/glutin/glutin)
error[E0277]: `std::sync::RwLockReadGuard<'static, Option<NativeWindow>>` cannot be sent between threads safely
   --> glutin/src/context.rs:204:6
    |
204 | impl FailToCompileIfNotSendSync for Context<NotCurrent> {}
    |      ^^^^^^^^^^^^^^^^^^^^^^^^^^ `std::sync::RwLockReadGuard<'static, Option<NativeWindow>>` cannot be sent between threads safely
    |
    = help: within `AndroidContext`, the trait `Send` is not implemented for `std::sync::RwLockReadGuard<'static, Option<NativeWindow>>`
    = note: required because it appears within the type `Option<std::sync::RwLockReadGuard<'static, Option<NativeWindow>>>`
note: required because it appears within the type `AndroidContext`
   --> glutin/src/api/android/mod.rs:18:8
    |
18  | struct AndroidContext {
    |        ^^^^^^^^^^^^^^
    = note: required because of the requirements on the impl of `Sync` for `Arc<AndroidContext>`
note: required because it appears within the type `api::android::Context`
   --> glutin/src/api/android/mod.rs:25:12
    |
25  | pub struct Context(Arc<AndroidContext>);
    |            ^^^^^^^
note: required because it appears within the type `context::Context<context::NotCurrent>`
   --> glutin/src/context.rs:34:12
    |
34  | pub struct Context<T: ContextCurrentState> {
    |            ^^^^^^^
note: required by a bound in `FailToCompileIfNotSendSync`
   --> glutin/src/context.rs:201:18
    |
199 | trait FailToCompileIfNotSendSync
    |       -------------------------- required by a bound in this
200 | where
201 |     Self: Send + Sync,
    |                  ^^^^ required by this bound in `FailToCompileIfNotSendSync`
```

Unless I'm misunderstanding your suggestion? Any ideas how to fix?

glutin/src/api/android/mod.rs
jamienicol: I had tried this too, but got this error: `move occurs because value has type `Option<NativeWindow>`, which does not implement the `Copy` trait`

and I couldn't figure out how to appease the borrow checker

MarijnS95: Apologies for editing this suggestion in the last minute; it compiles for me with `as_ref()` :)

glutin/src/api/android/mod.rs
MarijnS95: You're not misunderstanding my suggestion, it's just a little oversight that I haven't checked further than "this lock guard must be held in some way" and now that you actually try it, see that `glutin` requires its context to be `Send`/`Sync`.

My gut feeling says that `glutin` relies on the (EGL) context is internally synchronized (hence safe to not only `Send` to a different thread, but also access from multiple threads concurrently (`Sync`) without locking on the rust size). A lock guard is obviously not, so it seems like we'll have to wrap the lock guard inside a `Mutex` :hand_over_mouth:. Logically that doesn't solve the requirement for `Send` though, maybe those locks from `parking_lot` _can_ be moved into a different thread?

MarijnS95: For reference, the nomicon nicely explains why lock guards are not `Send`:

https://doc.rust-lang.org/nomicon/send-and-sync.html
> A nice example where this does not happen is with a MutexGuard: notice how [it is not Send](https://doc.rust-lang.org/std/sync/struct.MutexGuard.html#impl-Send). The implementation of MutexGuard [uses libraries](https://github.com/rust-lang/rust/issues/23465#issuecomment-82730326) that require you to ensure you don't try to free a lock that you acquired in a different thread.

Looks like `parking_lot` supports sending a lock guard though, as long as we enable `send_guard`: https://github.com/Amanieu/parking_lot#usage

At this point it feels like we might put this issue on ice, and solve it when we get to it - as I'd like to address it in `winit` as well and anyway need `parking_lot` to be able to map a lock guard and get rid of the inner `Option<>`.

MarijnS95:
@jamienicol As part of this PR, can you also update the `### Android` section in `README.md`? Just spotted that it still links to the old `android-rs-glue` repo.

glutin/src/api/android/mod.rs
MarijnS95: @jamienicol Here you go:

https://github.com/rust-windowing/android-ndk-rs/pull/282
https://github.com/jamienicol/glutin/compare/ndk-glue...MarijnS95:ndk-glue-native-window-lock

MarijnS95: We'll need to figure out something with the versioning - I had to do some `patch` hacks to get everything in the dependency chain (`winit`) to use the same `ndk` and `ndk-glue`. But that's _after_ everyone is content with this idea :)

glutin/Cargo.toml
```diff
@@ -25,9 +25,9 @@ lazy_static = "1.3"
 winit = { version = "0.26", default-features = false }
 
 [target.'cfg(target_os = "android")'.dependencies]
-android_glue = "0.2"
 glutin_egl_sys = { version = "0.1.5", path = "../glutin_egl_sys" }
 libloading = "0.7"
+ndk-glue = "0.5"
```
MarijnS95, line 30:
```suggestion
ndk-glue = "0.5" # Keep in sync with winit
```

glutin/src/api/android/mod.rs
MarijnS95: Fwiw I've added this to `winit` in https://github.com/rust-windowing/winit/pull/2307 to make the lock held at least as long as until after `winit` users processed and returned from `Event::Suspended`. With that in this becomes less of an issue but it still feels right to hold the lock at exactly the same place where the "consumer" of the window - an EGL window surface - is lifecycle-managed. However, given that we don't really help the user yet with lifetime management given the removal of `set_suspend_callback`, we should just get your PR merged without solving this one and I'll take it on later.

rib:
If you're looking at switching the glue layer, I wonder if it could be worth evaluating these glue crates that I've been working on recently, while I've been investigating how to support native Rust applications that aren't always based on NativeActivity.

https://github.com/rib/agdk-rust

The glue API was initially based on `ndk-glue` but it notably handles native <-> Java synchronization internally so there's no need to e.g. expose a need to hold a lock over the internal window state.

The API also supports input events, and exposes things like the `sdk_version`, `internal/external/obb_data_dir` without needing to depend on ANativeActivity or AInputQueue so there's more opportunity to support alternative Activity base classes.

MarijnS95:
@rib Let's tackle one thing at a time. This patch and some subsequent changes are working well for its purpose, and are simply necessary to accompany `glutin`s upgrade to newer `winit`, which wasn't tested on Android nor is caught by the CI (but I will change that once this patch lands).

`winit` in its current form uses `ndk-glue`, and `winit` is tied "heavily" into `glutin`, hence it is not possible to randomly use a different native/game-activity wrapper here.

Leave the sales-pich for `winit` and `glutin` can follow.

glutin/src/api/android/mod.rs
MarijnS95: I've been having _even more_ thoughts about this (https://github.com/rust-windowing/winit/pull/2307#issuecomment-1140520631), and I don't think `glutin` should hold the lock for this bit of the API but depend on `winit` to do it (https://github.com/rust-windowing/winit/pull/2307).

Then, with that squared away (though we can just "trust" the surface is safe for now since that's what everyone is doing and works "okay enough", until the next `winit` release with my PR in) `glutin` doesn't - shouldn't - use `ndk_glue` directly at all. `winit` uses `raw-window-handle` to expose these details, and `glutin` can simply use it directly:

https://github.com/MarijnS95/glutin/commit/baaf31a

(I've added the locks to `winit` with that in mind: people will use that `raw-window-handle` interface and never see nor know about the Android lock directly).

In fact we can use this to our advantage to unify `glutin` to use this "generic" way to communicate window-handle pointers around and create GL contexts on it, that's how it's done for Vulkan at least: https://github.com/ash-rs/ash/tree/master/ash-window (out of scope for this PR).

rib:
> @rib Let's tackle one thing at a time. This patch and some subsequent changes are working well for its purpose, and are simply necessary to accompany `glutin`s upgrade to newer `winit`, which wasn't tested on Android nor is caught by the CI (but I will change that once this patch lands).

Yep, absolutely, makes sense. Definitely not meaning to disrupt fixing issues that exist now.

> 
> `winit` in its current form uses `ndk-glue`, and `winit` is tied "heavily" into `glutin`, hence it is not possible to randomly use a different native/game-activity wrapper here.
> 
> Leave the sales-pich for `winit` and `glutin` can follow.

haha, :) sorry for sounding like a sales pitch. I mainly just wanted to float the idea, with the specific technical context that the API in `native-activity` and `game-activity` offers a potential solution to this synchronization problem.

There's also room for incorporating a similar approach in `ndk-glue` and my intention behind suggesting to "evaluate" my work was to try and tone down being hard sale pitch vs pointing to a potential technical solution too (but yeah when I read it back it does look like a sales pitch, sorry about that!)

jamienicol:
Perhaps I should combine the two Android changelog entries to emphasise what the users will actually care about, eg:

> * Fix compilation on Android:
>   * Switch from `StaticStructGenerator` to `StructGenerator` to dynamically load symbols.
>   * Replace `android_glue` dependency with `ndk-glue`, and remove broken lifecycle event handling.
>   * Glutin can now be used on Android, however, the application must ensure it only creates the Context following a winit `Event::Resumed` event, and destroy the context in response to a `Event::Suspended` event.

How does that sound?

README.md
```diff
@@ -50,7 +50,7 @@ Glutin is only officially supported on the latest stable version of the Rust com
 
 To compile the examples for android, you have to use the `cargo apk` utility.
 
-See [the `android-rs-glue` repository](https://github.com/rust-windowing/android-rs-glue) for instructions.
+See [the `android-nd-rs` repository](https://github.com/rust-windowing/android-ndk-rs/cargo-apk) for instructions.
```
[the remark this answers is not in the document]
MarijnS95: (I'll extend this readme when submitting my example)

MarijnS95 (approved):

MarijnS95:
@jamienicol Sounds good on the suggested changelog regrouping, I love some sub-`ul`'s :)

MarijnS95:
> Do we need to do anything with the native window lock in this PR? or do we need to wait for your winit PR to land, and update the winit dep in glutin first?

Let's leave it like this for now. I'll probably plan a larger change that removes `ndk-glue` again in favour of using `raw_window_handle`, which should work across all platforms and isn't wrapped in a lock anyway.

CHANGELOG.md
```diff
@@ -2,8 +2,11 @@
 
 - Fix crash when creating OpenGLES context without explicit version.
 - Add `buffer_age` method on `WindowedContext`.
-- On Android, switched from `StaticStructGenerator` to `StructGenerator`, fixing some compilation errors.
 - Return an `Err` instead of panicking when surfaceless GLX context creation fails on Linux.
+- Fix compilation on Android:
+  - Switch from `StaticStructGenerator` to `StructGenerator` to dynamically load symbols.
+  - Replace `android_glue` dependency with `ndk-glue`, and remove broken lifecycle event handling.
+  - Glutin can now be used on Android, however, the application must ensure it only creates the Context following a winit `Event::Resumed` event, and destroys the context in response to a `Event::Suspended` event.
```
MarijnS95, line 9:
```suggestion
  - Glutin can now be used on Android, however, the application must ensure it only creates the `Context` following a winit `Event::Resumed` event, and destroys the `Context` in response to a `Event::Suspended` event.
```
?

jamienicol: done

MarijnS95:
> haha, :) sorry for sounding like a sales pitch.

Hehe you may have noticed I have a habit for spamming ideas and walls of text in PRs too (side-eye https://github.com/rust-windowing/winit/pull/2307) :grin: 

> I mainly just wanted to float the idea, with the specific technical context that the API in `native-activity` and `game-activity` offers a potential solution to this synchronization problem.

It's a good idea to conjure up something that solves this broadly for the generic `RawWindowHanle` case, tying into the discussion of making `winit` more consistent across all platforms wrt `Event::Resumed`/`Event::Suspended` etc.

> There's also room for incorporating a similar approach in `ndk-glue` and my intention behind suggesting to "evaluate" my work was to try and tone down being hard sale pitch vs pointing to a potential technical solution too (but yeah when I read it back it does look like a sales pitch, sorry about that!)

It's all good, just that `glutin` doesn't - shouldn't - use any of the input queues, native activity, data dir stuff, that's more `winit`s place. As mentioned in the long suggestion thread above, if we use `RawWindowHandle` in `glutin` it can be made oblivious to what's backing the window - and we can even use "generic" Java `Surface`s (oddly named `NativeWindow` in the NDK) that don't need to represent the entire window crated by/for `NativeActivity`: https://github.com/rust-windowing/glutin/issues/1324#issuecomment-1140547588

MarijnS95 (approved):
:partying_face:

MarijnS95:
> It's all good, just that `glutin` doesn't - shouldn't - use any of the input queues

How embarrassing; I just read the title of this project in the README and `IN` in `glutIN` apparently stands for `INPUT` :hand_over_mouth: 

I thought `glutin` merely juggled the various OpenGL contexts around, and it's probably only exposing input through the reexported `EventLoop` from `winit` nowadays :grimacing:

PR #1412, opened 2022-05-28
Run clippy in the CI with warnings disallowed; address most warnings

MarijnS95:
Depends on #1413 + #1408.

This is a draft mainly to look at how `cargo clippy --fix` and myself have addressed warnings for most issues. There are still a few remaining, and clippy found some odd/fishy things we should look into.

  - [x] Android + EGL;
  - [x] Wayland + EGL;
  - [ ] Wayland + EGL on `sharing` example (doesn't run at all, but has nontrivial `clippy` warnings);
  - [ ] Windows.

### DIFF
--- a/glutin/src/api/egl/mod.rs
+++ b/glutin/src/api/egl/mod.rs
@@ -235,7 +235,7 @@ fn get_native_display(native_display: &NativeDisplay) -> *const raw::c_void {
             vec![]
         } else {
             let p = CStr::from_ptr(p);
-            let list = String::from_utf8(p.to_bytes().to_vec()).unwrap_or_else(|_| String::new());
+            let list = String::from_utf8(p.to_bytes().to_vec()).unwrap_or_default();
             list.split(' ').map(|e| e.to_string()).collect::<Vec<_>>()
         }
     };
@@ -397,7 +397,7 @@ impl Context {
         let extensions = if egl_version >= (1, 2) {
             let p =
                 unsafe { CStr::from_ptr(egl.QueryString(display, ffi::egl::EXTENSIONS as i32)) };
-            let list = String::from_utf8(p.to_bytes().to_vec()).unwrap_or_else(|_| String::new());
+            let list = String::from_utf8(p.to_bytes().to_vec()).unwrap_or_default();
             list.split(' ').map(|e| e.to_string()).collect::<Vec<_>>()
         } else {
             vec![]
--- a/glutin/src/api/egl/mod.rs
+++ b/glutin/src/api/egl/mod.rs
@@ -7,101 +7,18 @@
     target_os = "netbsd",
     target_os = "openbsd",
 ))]
-#![allow(unused_variables)]
 
-mod egl {
-    use super::ffi;
-    use crate::api::dlloader::{SymTrait, SymWrapper};
-    use libloading;
-    use std::sync::{Arc, Mutex};
-
-    #[cfg(unix)]
-    use libloading::os::unix as libloading_os;
-    #[cfg(windows)]
-    use libloading::os::windows as libloading_os;
-
-    #[derive(Clone)]
-    pub struct Egl(pub SymWrapper<ffi::egl::Egl>);
-
-    /// Because `*const raw::c_void` doesn't implement `Sync`.
-    unsafe impl Sync for Egl {}
-
-    type EglGetProcAddressType = libloading_os::Symbol<
-        unsafe extern "C" fn(*const std::os::raw::c_void) -> *const std::os::raw::c_void,
-    >;
-
-    lazy_static! {
-        static ref EGL_GET_PROC_ADDRESS: Arc<Mutex<Option<EglGetProcAddressType>>> =
-            Arc::new(Mutex::new(None));
-    }
-
-    impl SymTrait for ffi::egl::Egl {
-        fn load_with(lib: &libloading::Library) -> Self {
-            let f = move |s: &'static str| -> *const std::os::raw::c_void {
-                // Check if the symbol is available in the library directly. If
-                // it is, just return it.
-                match unsafe {
-                    lib.get(std::ffi::CString::new(s.as_bytes()).unwrap().as_bytes_with_nul())
-                } {
-                    Ok(sym) => return *sym,
-                    Err(_) => (),
-                };
-
-                let mut egl_get_proc_address = (*EGL_GET_PROC_ADDRESS).lock().unwrap();
-                if egl_get_proc_address.is_none() {
-                    unsafe {
-                        let sym: libloading::Symbol<
-                            unsafe extern "C" fn(
-                                *const std::os::raw::c_void,
-                            )
-                                -> *const std::os::raw::c_void,
-                        > = lib.get(b"eglGetProcAddress").unwrap();
-                        *egl_get_proc_address = Some(sym.into_raw());
-                    }
-                }
-
-                // The symbol was not available in the library, so ask
-                // eglGetProcAddress for it. Note that eglGetProcAddress was
-                // only able to look up extension functions prior to EGL 1.5,
-                // hence this two-part dance.
-                unsafe {
-                    (egl_get_proc_address.as_ref().unwrap())(
-                        std::ffi::CString::new(s.as_bytes()).unwrap().as_bytes_with_nul().as_ptr()
-                            as *const std::os::raw::c_void,
-                    )
-                }
-            };
-
-            Self::load_with(f)
-        }
-    }
-
-    impl Egl {
-        pub fn new() -> Result<Self, ()> {
-            #[cfg(target_os = "windows")]
-            let paths = vec!["libEGL.dll", "atioglxx.dll"];
-
-            #[cfg(not(target_os = "windows"))]
-            let paths = vec!["libEGL.so.1", "libEGL.so"];
-
-            SymWrapper::new(paths).map(Egl)
-        }
-    }
-}
-
-mod make_current_guard;
-
-pub use self::egl::Egl;
-use self::make_current_guard::MakeCurrentGuard;
-#[cfg(not(target_os = "windows"))]
-use crate::Rect;
-use crate::{
-    Api, ContextError, CreationError, GlAttributes, GlRequest, PixelFormat,
-    PixelFormatRequirements, ReleaseBehavior, Robustness,
-};
+use std::ffi::{CStr, CString};
+use std::ops::{Deref, DerefMut};
+use std::os::raw;
+use std::sync::{Arc, Mutex};
 
 use glutin_egl_sys as ffi;
-use parking_lot::Mutex;
+use libloading;
+#[cfg(unix)]
+use libloading::os::unix as libloading_os;
+#[cfg(windows)]
+use libloading::os::windows as libloading_os;
 #[cfg(any(
     target_os = "android",
     target_os = "windows",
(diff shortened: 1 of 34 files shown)


glutin_examples/examples/sharing.rs
```diff
@@ -61,10 +61,9 @@ fn main() {
         glw.gl.BindFramebuffer(gl::DRAW_FRAMEBUFFER, 0);
         glw.gl.Viewport(0, 0, size.width as _, size.height as _);
     }
-    std::mem::drop(windowed_context);
```
MarijnS95, line 64: Clippy points out here that all these `mem::drop`s on the result of `get_current()` are useless, because that function returns borrows. Dropping a borrow has no effect. Now, with this borrow gone, `clippy` complains about some unused `let` bindings so we should see if the function call itself is just "getting" something or if it's really doing something under the hood of which the result somehow needs to be "kept alive".

glutin/src/api/glx/mod.rs
```diff
@@ -8,51 +8,49 @@
...
-            Self::load_with(|sym| unsafe {
-                lib.get(std::ffi::CString::new(sym.as_bytes()).unwrap().as_bytes_with_nul())
-                    .map(|sym| *sym)
-                    .unwrap_or(std::ptr::null_mut())
-            })
-        }
+
+use crate::api::dlloader::{SymTrait, SymWrapper};
+use glutin_glx_sys as ffi;
+use std::ops::{Deref, DerefMut};
```
MarijnS95, line 14: Clippy didn't like the submodule here being named the same as the containing module, and I don't see any reason for it to exist. Perhaps I should reorder the `use` statements to all be at the top of the file, perhaps a subsequent PR on top that cleans up everything automatically once using `cargo fmt -- --config imports_granularity=crate --config group_imports=stdexternalcrate`? Let me know if that's okay and/or we should pick different settings for it.

glutin/src/api/egl/mod.rs
```diff
@@ -7,91 +7,85 @@
...
-#![allow(unused_variables)]
 
-mod egl {
-    use super::ffi;
-    use crate::api::dlloader::{SymTrait, SymWrapper};
-    use libloading;
-    use std::sync::{Arc, Mutex};
+use crate::api::dlloader::{SymTrait, SymWrapper};
+use libloading;
+use std::sync::{Arc, Mutex};
```
MarijnS95, line 13: Same here, I didn't see a need for this submodule to exist besides the use of `std::sync::Mutex` within versus `parking_lot::Mutex` outside. Same question applies regarding moving (preferably though `cargo fmt -- --config imports_granularity=crate --config group_imports=stdexternalcrate`) imports back together.

glutin/src/api/glx/mod.rs
kchibisov: >cargo fmt -- --config imports_granularity=crate --config group_imports=stdexternalcrate? Let me know if that's okay and/or we should pick different settings for it.

In general I'd change the whole formatting for glutin/winit all together, given that it's hard to read code with rustfmt defaults. Though, lots of good options are not available on stable rustfmt and requires nightly for it...

In general the entire glutin should be rewritten to be decoupled from winit, but it's another long story...

MarijnS95: In contrary, I've always found `rustfmt` code with the defaults extremely pleasant to read, especially in contrast to the horrible abomination `clang-format` usually comes up with for C/C++ code. Yes, it is a pain that we cannot use `rustfmt` to automatically regroup and recombine `use` statements on `stable`, but it's "coming™" at least.

So the question is basically, should we:
- Reformat it just once with this command;
- Force this formatting in the CI, even if contributors can't easily reproduce it locally;
- Don't do anything for now.

I don't think it matters what `winit` does, I'm contributing to `glutin` here and now and am not looking at `winit` code for the time being :) - we can tackle that another day unless you feel like both projects should do things in unison (which I doubt, given that you want to _decouple it_).

One important mention: these features are unstable and don't work in `rustfmt.toml` yet. _However_, on a stable Rust toolchain they _do_ work when passed on the commandline as showcased above. There's still the warning that it's `only available in nightly channel`, but it works :shrug:

MarijnS95: One more comment on decoupling: that's exactly what I'll need to support "random" `Surface`s in `glutin`, without a window loop nor events (ie. #1324). I think I can make it work, but it's odd to have the `winit` dependency and usage chained through the API at that point. Perhaps rewriting it in a more layered system would help, but that's for another issue.

kchibisov: I'd suggest to look into https://github.com/rust-windowing/glutin/tree/v0.23_reworks. It should have like the most of the thing done iirc. just the author and the past maintainer disappeared...

But we can pick up that branch and try moving forward with it, given that glutin didn't had that much changes and most can ported to it without much issues.

If you feel like, I think I can pick up that rework, though, I'm not sure I'll move forward that much with macOS/Windows bits, but at least we can try...

kchibisov: >In contrary, I've always found rustfmt code with the defaults extremely pleasant to read

The defaults are probably nice, however I like it more with `use_small_heuristics = "Max"`. Since it's less lines of code to read and having one liners is easier for me to grasp other than 5 lines of basically `.ok()` lines.

MarijnS95: @kchibisov Just a heads-up that I don't use OpenGL/EGL, but rolled into the maintainer role of `android-ndk-rs` some time ago and kept getting overloaded with strange attempts to fix "glutin not compiling for Android" in strange ways on the ndk/winit side (ie. adding unimplemented callback interfaces just to make things compile, not speaking of working just yet - why not fix it in glutin :shrug:).

Now I've taken it upon myself to fix this situation and close a significant amount of issues and stale PRs.

With that come a few drive-by fixes like addressing all `clippy` lints and rolling them up into the CI, to make my own development/contributing experience a little nicer. We'll get Android added to the CI when open PR(s, #1411) land too, so that it doesn't invisibly break in the future.

---

As such I'd love for you to pick up the rework as a more active maintainer. I'll probably stick around to aid on Android-specific issues, but I don't think supporting "window-less" surface support (#1324) are undoable before such a rework lands (so no hurry!).

---

> The defaults are probably nice, however I like it more with `use_small_heuristics = "Max"`. Since it's less lines of code to read and having one liners is easier for me to grasp other than 5 lines of basically `.ok()` lines.

I hadn't really thought about that one, but yes this has been annoying me on some occasions. `rustfmt` isn't absolutely perfect and I'll keep this one in mind, thanks for bringing it to my attention :+1:

MarijnS95:
Looks like `nightly` clippy is throwing some false-positives on `objc::msg_send![]`'s return-type annotation https://docs.rs/objc/latest/objc/#messaging-objects.

That raises the question: is it desired to run `nightly` clippy? I'm personally all in favour of having `clippy` in the CI (some call it low signal-to-noise ratio, I find it making the code more pleasant to read), but `nightly` clippy is a bit overzealous. It has a high amount of false-positives compared to the actual future-proofing one might get; better wait for the stabilized version.

So for now this PR will only run `stable` clippy :)

kchibisov:
Running on nightly doesn't make much sense, except if the target requires nightly for whatever reason. Just stable is enough, I guess.

MarijnS95:
Doing that now, and all checks succeed :tada: - what's your thought about getting the other PRs and this one in?

glutin/src/api/ios/mod.rs
```diff
@@ -384,13 +384,13 @@ impl Context {
     pub fn get_proc_address(&self, proc_name: &str) -> *const core::ffi::c_void {
         let proc_name_c = CString::new(proc_name).expect("proc name contained interior nul byte");
         let path = b"/System/Library/Frameworks/OpenGLES.framework/OpenGLES\0";
-        let addr = unsafe {
+
+        // debug!("proc {} -> {:?}", proc_name, addr);
```
kchibisov, line 388: Can remove this.

kchibisov (approved):

glutin_examples/examples/sharing.rs
MarijnS95: I haven't been able to test this example yet, it fails to find any config in `eglChooseConfig` and I have yet to dissect what is going on/wrong :)

kchibisov (changes requested):
Actually, could you add into lib.rs of each crate.
```rust
#![deny(clippy::all)
#![cfg_attr(feature = "cargo-clippy", deny(warnings))]
```
Instead of denying on CI?

MarijnS95:
@kchibisov What would be the advantage of those annotations? I personally don't like disallowing all clippy warnings during development, it doesn't allow contributors to compile code when perhaps having one or two testing variables / functions commented out.

Secondly, what's the advantage of disallowing normal `rustc` warnings through a `cargo-clippy` feature? I don't think I've seen such a setup yet.

kchibisov:
>@kchibisov What would be the advantage of those annotations?

I mean that with those two you can compile normally, but when you run clippy everything errors out.

MarijnS95:
> I mean that with those two you can compile normally, but when you run clippy everything errors out. The intention is to error out on clippy, while allowing checks/builds to throw warnings as usual.

Right, yes. I use `clippy` in `rust-analyzer` (to see `clippy` lints directly in my editor, instead of only `rustc` warnings) and it typically stops linting on the first crate that encounters errors (`deny`, `-Dwarnings`) meaning it was not working on `glutin_examples` until #1408/#1413 were addressed (both are `deny` lints).

Strangely enough it seems to be fine now, without those commits :shrug: 

> That's what we're using in alacritty for years. Basically it means: "when running clippy, treat all rustc warnings as errors".

Ack, understood. I have previously seen but forgot that `cargo-clippy` is an actual feature (not `cfg`) that's enabled inside crates when "compiling" code under `clippy`.

Do you mind if I leave `deny(clippy::all)` out? That group consists of warn/deny lints only, and is already unambiguously covered by `#![cfg_attr(feature = "cargo-clippy", deny(warnings))]`. For example:

`` note: `#[deny(clippy::missing_safety_doc)]` implied by `#[deny(warnings)]` ``

> Instead of denying on CI?

Do you mind if I leave this? In the hypothetical case that a new crate is added? Otherwise this should also be added to every binary (test/example) target.

kchibisov:
>Do you mind if I leave deny(clippy::all) out? That group consists of warn/deny lints only, and is already unambiguously covered by #![cfg_attr(feature = "cargo-clippy", deny(warnings))]. For example:

That's fine. You can go ahead.

kchibisov:
@MarijnS95 sharing is also failing for me. It could be failing that way for a long time on Wayland, and I won't be surprised. Not sure that we should bother with it here since it's not something new.

glutin/src/api/egl/mod.rs
```diff
@@ -246,12 +235,12 @@ fn get_native_display(native_display: &NativeDisplay) -> *const raw::c_void {
             vec![]
         } else {
             let p = CStr::from_ptr(p);
-            let list = String::from_utf8(p.to_bytes().to_vec()).unwrap_or_else(|_| format!(""));
+            let list = String::from_utf8(p.to_bytes().to_vec()).unwrap_or_else(|_| String::new());
```
kchibisov, line 238:
```suggestion
            let list = String::from_utf8(p.to_bytes().to_vec()).unwrap_or_default();
```

glutin/src/api/egl/mod.rs
```diff
@@ -408,14 +397,14 @@ impl Context {
         let extensions = if egl_version >= (1, 2) {
             let p =
                 unsafe { CStr::from_ptr(egl.QueryString(display, ffi::egl::EXTENSIONS as i32)) };
-            let list = String::from_utf8(p.to_bytes().to_vec()).unwrap_or_else(|_| format!(""));
+            let list = String::from_utf8(p.to_bytes().to_vec()).unwrap_or_else(|_| String::new());
```
kchibisov, line 400:
```suggestion
            let list = String::from_utf8(p.to_bytes().to_vec()).unwrap_or_default();
```

MarijnS95:
@kchibisov Perfect, let's go for it then. If you can review this once more and approve it, I'll fix the conflict and re-push!

kchibisov (changes requested):
Just a small nits I've found after re-review.

MarijnS95:
Nice catch, I hadn't spotted that one in the automatic `clippy --fix` diff :)

kchibisov (approved):

PR #1417, opened 2022-06-05
Add Android example

MarijnS95:
As mentioned in #1398/#1411 we should have an example that showcases how to deal with the Android activity lifecycle.  Ultimately this turns into a helper function on `ContextWrapper` that has access to the internal surface for recreation, as did the broken implementation prior to #1411.  At the same time I hope to replace `new_windowed()` with an implementation soon™ that takes a `RawWindowHandle`, making the code more generic, `winit`-agnostic, and allows users to use Android `Surface`s for specific views in their Java/Kotlin/Flutter apps without needing a fullscreen `NativeActivity`.

This PR is based on top of #1412 so that my editor doesn't go bonkers on all the lint warnings/errors :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ cargo run --example window
 Glutin is an OpenGL context creation library and doesn't directly provide
 OpenGL bindings for you.
 
-For examples, please look [here.](https://github.com/rust-windowing/glutin/tree/master/glutin_examples)
+For examples, please look [here](https://github.com/rust-windowing/glutin/tree/master/glutin_examples).
 
 Note that glutin aims at being a low-level brick in your rendering
 infrastructure. You are encouraged to write another layer of abstraction
@@ -52,6 +52,10 @@ The minimum rust version target by glutin is `1.60.0`.
 
 ### Android
 
-To compile the examples for android, you have to use the `cargo apk` utility.
+Be sure to handle Android's lifecycle correctly when using a `winit` window by only creating a GL surface after `winit` raises `Event::Resumed`, and destroy it again upon receiving `Event::Suspended`. See this in action in the [`android.rs` example](./glutin_examples/examples/android.rs).
 
-See [`cargo-apk` in the `android-ndk-rs` repository](https://github.com/rust-windowing/android-ndk-rs/tree/master/cargo-apk) for instructions.
+To compile and run the Android example on your device, install [`cargo-apk`](https://crates.io/crates/cargo-apk) and start the app using:
+
+```console
+$ cargo apk r -p glutin_examples --example android
+```
--- a/glutin_examples/Cargo.toml
+++ b/glutin_examples/Cargo.toml
@@ -23,6 +23,13 @@ glutin = { path = "../glutin", default-features = false }
 winit = { version = "0.27.2", default-features = false }
 raw-window-handle = "0.5.0"
 
+[target.'cfg(target_os = "android")'.dependencies]
+ndk-glue = "0.7" # Keep in sync with winit dependency
+
 [build-dependencies]
 gl_generator = "0.14"
 cfg_aliases = "0.1.1"
+
+[[example]]
+name = "android"
+crate-type = ["cdylib"]
--- a/glutin_examples/examples/android.rs
+++ b/glutin_examples/examples/android.rs
@@ -1,0 +1,6 @@
+#![cfg(target_os = "android")]
+
+#[ndk_glue::main(backtrace = "on")]
+fn main() {
+    glutin_examples::main()
+}
--- a/glutin_examples/src/lib.rs
+++ b/glutin_examples/src/lib.rs
@@ -28,11 +28,16 @@ pub fn main() {
 
     let raw_display = event_loop.raw_display_handle();
 
-    // We create a window before the display to accomodate for WGL, since it
-    // requires creating HDC for properly loading the WGL and it should be taken
-    // from the window you'll be rendering into.
-    let window = WindowBuilder::new().with_transparent(true).build(&event_loop).unwrap();
-    let raw_window_handle = window.raw_window_handle();
+    let mut window = cfg!(any(wgl_backend, glx_backend)).then(|| {
+        // We create a window before the display to accommodate for WGL, since it
+        // requires creating HDC for properly loading the WGL and it should be taken
+        // from the window you'll be rendering into.
+
+        // On GLX the window should be available for config_template() + find_configs()
+        // to propagate its Xlib visual ID.
+        WindowBuilder::new().with_transparent(true).build(&event_loop).unwrap()
+    });
+    let raw_window_handle = window.as_ref().map(|w| w.raw_window_handle());
 
     // Create the GL display. This will create display automatically for the
     // underlying GL platform. See support module on how it's being done.
@@ -41,73 +46,101 @@ pub fn main() {
     // Create the config we'll be used for window. We'll use the native window
     // raw-window-handle for it to get the right visual and use proper hdc. Note
     // that you can likely use it for other windows using the same config.
-    let template = config_template(window.raw_window_handle());
-    let config = unsafe {
-        gl_display
-            .find_configs(template)
-            .unwrap()
-            .reduce(|accum, config| {
-                // Find the config with the maximum number of samples.
-                //
-                // In general if you're not sure what you want in template you can request or
-                // don't want to require multisampling for example, you can search for a
-                // specific option you want afterwards.
-                //
-                // XXX however on macOS you can request only one config, so you should do
-                // a search with the help of `find_configs` and adjusting your template.
-                if config.num_samples() > accum.num_samples() {
-                    config
-                } else {
-                    accum
-                }
-            })
-            .unwrap()
-    };
+    let template = config_template(raw_window_handle);
+    let config = unsafe { gl_display.find_configs(template) }
+        .unwrap()
+        .reduce(|accum, config| {
+            // Find the config with the maximum number of samples.
+            //
+            // In general if you're not sure what you want in template you can request or
+            // don't want to require multisampling for example, you can search for a
+            // specific option you want afterwards.
+            //
+            // XXX however on macOS you can request only one config, so you should do
+            // a search with the help of `find_configs` and adjusting your template.
+            if config.num_samples() > accum.num_samples() {
+                config
+            } else {
+                accum
+            }
+        })
+        .unwrap();
 
     println!("Picked a config with {} samples", config.num_samples());
-
-    // Create a wrapper for GL window and surface.
-    let gl_window = GlWindow::from_existing(&gl_display, window, &config);
 
     // The context creation part. It can be created before surface and that's how
     // it's expected in multithreaded + multiwindow operation mode, since you
     // can send NotCurrentContext, but not Surface.
-    let context_attributes = ContextAttributesBuilder::new().build(Some(raw_window_handle));
+    let context_attributes = ContextAttributesBuilder::new().build(raw_window_handle);
 
     // Since glutin by default tries to create OpenGL core context, which may not be
     // present we should try gles.
     let fallback_context_attributes = ContextAttributesBuilder::new()
         .with_context_api(ContextApi::Gles(None))
-        .build(Some(raw_window_handle));
-    let gl_context = unsafe {
+        .build(raw_window_handle);
+    let mut not_current_gl_context = Some(unsafe {
         gl_display.create_context(&config, &context_attributes).unwrap_or_else(|_| {
             gl_display
                 .create_context(&config, &fallback_context_attributes)
                 .expect("failed to create context")
         })
-    };
+    });
 
-    // Make it current and load symbols.
-    let gl_context = gl_context.make_current(&gl_window.surface).unwrap();
+    let mut state = None;
+    let mut renderer = None;
 
-    // WGL requires current context on the calling thread to load symbols properly,
-    // so the call here is for portability reasons. In case you don't target WGL
-    // you can call it right after display creation.
-    //
-    // The symbol loading is done by the renderer.
-    let renderer = Renderer::new(&gl_display);
-
-    // Try setting vsync.
-    if let Err(res) = gl_window
-        .surface
-        .set_swap_interval(&gl_context, SwapInterval::Wait(NonZeroU32::new(1).unwrap()))
-    {
-        eprintln!("Error setting vsync: {:?}", res);
-    }
-
-    event_loop.run(move |event, _, control_flow| {
+    event_loop.run(move |event, event_loop_window_target, control_flow| {
         *control_flow = ControlFlow::Wait;
         match event {
+            Event::Resumed => {
+                // While this event is only relevant for Android, it is raised on all platforms
+                // to provide a consistent place to create windows
+
+                #[cfg(target_os = "android")]
+                println!("Android window available");
+
+                // Take a possibly early created window, or create a new one
+                let window = window.take().unwrap_or_else(|| {
+                    WindowBuilder::new().build(event_loop_window_target).unwrap()
+                });
+
+                // Create a wrapper for GL window and surface.
+                let gl_window = GlWindow::from_existing(&gl_display, window, &config);
+
+                // Make it current.
+                let gl_context = not_current_gl_context
+                    .take()
+                    .unwrap()
+                    .make_current(&gl_window.surface)
+                    .unwrap();
+
+                // The context needs to be current for the Renderer to set up shaders and
+                // buffers. It also performs function loading, which needs a current context on
+                // WGL.
+                renderer.get_or_insert_with(|| Renderer::new(&gl_display));
+
+                // Try setting vsync.
+                if let Err(res) = gl_window
+                    .surface
+                    .set_swap_interval(&gl_context, SwapInterval::Wait(NonZeroU32::new(1).unwrap()))
+                {
+                    eprintln!("Error setting vsync: {:?}", res);
+                }
+
+                assert!(state.replace((gl_context, gl_window)).is_none());
+            },
+            Event::Suspended => {
+                // This event is only raised on Android, where the backing NativeWindow for a GL
+                // Surface can appear and disappear at any moment.
+                println!("Android window removed");
+                // Destroy the GL Surface and un-current the GL Context before ndk-glue releases
+                // the window back to the system.
+                let (gl_context, _) = state.take().unwrap();
+                assert!(not_current_gl_context
+                    .replace(gl_context.make_not_current().unwrap())
+                    .is_none());
+            },
+
             Event::WindowEvent { event, .. } => match event {
                 WindowEvent::Resized(size) => {
                     if size.width != 0 && size.height != 0 {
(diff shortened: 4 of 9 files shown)


glutin/src/api/egl/mod.rs
```diff
@@ -964,7 +964,8 @@ where
         out.push(surface_type as raw::c_int);
 
         match (api, version) {
-            (Api::OpenGlEs, Some((3, _))) => {
+            // TODO: `version` is `None` when GLRequest is Latest
+            (Api::OpenGlEs, None | Some((3, _))) => {
```
MarijnS95, line 968: Not yet sure what to do about this, when requesting `Latest` instead of a specific version. Can/should we load the specific version from anywhere?

glutin_examples/examples/android.rs
```diff
@@ -0,0 +1,59 @@
...
+                println!("Android window removed");
+                // Destroy the GL context before ndk-glue releases the window back to the system
+                state = None;
+            }
+            Event::WindowEvent { event: WindowEvent::CloseRequested, .. } => {
+                *control_flow = ControlFlow::Exit
+            }
+            Event::WindowEvent { event: WindowEvent::Resized(size), .. } => {
+                if let Some((windowed_context, _)) = &state {
+                    windowed_context.resize(size);
```
MarijnS95, line 46: No-op on Android, not implemented for EGL.

kchibisov: What do you mean, we do have resize on egl, just on wayland or something like that.

MarijnS95: It's not provided/implemented/needed in the EGL API, so Android implements this as `pub fn resize(&self, _: u32, _: u32) {}`.

kchibisov: Ok then. This api is purely for consistency with other platforms, X11 also doesn't have it.

MarijnS95: Yep, as such I'm simply not sure if the example should call or omit it.

README.md
```diff
@@ -48,9 +48,13 @@ Glutin is only officially supported on the latest stable version of the Rust com
 
 ### Android
 
-To compile the examples for android, you have to use the `cargo apk` utility.
+Be sure to handle Android's lifecycle correctly when using a `winit` window by only creating a GL surface (currently entire context) after `winit` raises `Event::Resumed`, and destroy it again upon receiving `Event::Suspended`. See this in action in [the `android.rs` example](./glutin_examples/examples/android.rs).
```
MarijnS95, line 51: We should possibly reword this if/when https://github.com/rust-windowing/winit/pull/2331 lands.

glutin_examples/examples/android.rs
rib: Wouldn't it be best, ideally, to aim for an example that demonstrates running on Android _and_ hopefully most/all other platforms too?

An Android example is mainly just necessary because, compared to other platforms, it has these extra fiddly details due to the surface view coming and going but ideally most people will just want to write portable code and not have to worry about the platform until they are doing something fundamentally platform specific.

MarijnS95: That's the ideal case, but since we aren't there yet (i.e. your `winit` PR is just the beginning and I have more changes planned for `glutin`) this seems like the best temporary compromise.

Note that it also has some additional setup in `Cargo.toml` - but it breaks CI and I **thiiink** we can have a `crate-type = ["bin", "cdylib"]`.

This'll also help us toy around with reintroducing some helper API around `Suspended`/`Resumed` events that only recreate the GL Surface, instead of the entire EGL/GL Context.

MarijnS95: Nope:

```console
$ cargo r -p glutin_examples --example android
error: example target `android` is a library and cannot be executed`
$ cargo apk r -p glutin_examples --example android
   Compiling glutin_examples v0.1.2 (glutin/glutin_examples)
error: cannot mix `bin` crate type with others

error: could not compile `glutin_examples` due to previous error
Error: Command `cargo build --target aarch64-linux-android -p glutin_examples --example android` had a non-zero exit code.
```

glutin_examples/Cargo.toml
```diff
@@ -17,3 +17,10 @@ image = "0.24"
 
 [build-dependencies]
 gl_generator = "0.14"
+
+[target.'cfg(target_os = "android")'.dependencies]
+ndk-glue = "0.5" # Keep in sync with winit
+
+[[example]]
+name = "android"
+crate-type = ["cdylib"]
```
MarijnS95, line 26: I've yet to figure out how to best solve the CI failure here:

```console
$ cargo $WEB test --no-run --verbose --target x86_64-apple-ios
error: cannot produce cdylib for `glutin_examples v0.1.2 (/Users/runner/work/glutin/glutin/glutin_examples)` as the target `x86_64-apple-ios` does not support these crate types
Error: Process completed with exit code 101.
```

kchibisov:
This would need update to the latest Api.

MarijnS95:
@kchibisov Sure, I've just returned from holiday and am slowly getting back into things, here's hoping we can use this Android example to finally vet the API to be compatible/sensible in this regard.

Pieced together something that compiles at least, but running into ``thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Error { raw_code: None, raw_os_message: None, kind: BadConfig }', glutin_examples/examples/android.rs:54:73``, will throw some more time at this when it's light outside again.

kchibisov:
>Pieced together something that compiles at least, but running into thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Error { raw_code: None, raw_os_message: None, kind: BadConfig }', glutin_examples/examples/android.rs:54:73, will throw some more time at this when it's light outside again.

You should ask for GLES I guess in the template builder.

MarijnS95:
@kchibisov Right, yes, I also had to add `.with_context_api(glutin::context::ContextApi::Gles(None))` (not sure why we have two places for this, I'm not too familiar with the GL(ES) startup/setup), it's rendering (clearing) a teal background now.

kchibisov:
@MarijnS95 could we use the same example as for desktop, but for android as well? I'll look into adjusting api here in that regard.

MarijnS95:
@kchibisov That'd be nice, I copy-pasted the example from yours so it's almost identical, `#[cfg_attr(target_os = "android", ndk_glue::main(backtrace = "on"))]` already makes this a generic `fn main()` and `Resumed` is emitted on every platform now so we might get away with a single file. `crate-type = ["cdylib"]` will throw a spanner in the works, though.

kchibisov:
Yeah, if we can have a single example for each platform it could be better? But I'm not entirely sure? Maybe having a separate example for android works as well...

MarijnS95:
@kchibisov I think we can, we already have the `glutin_example` crate-library provide `support`; the more code we move in there the smaller / less overlap remains in the desktop-runnable binary and Android-runnable library (apk).

kchibisov:
You should rebase and fix the CI. Be aware that formatting is on nightly.

MarijnS95:
I'm aware, and was just waiting for the dependent PR to be merged before reevaluating where to take this.

glutin_examples/examples/android.rs
```diff
@@ -0,0 +1,126 @@
...
+
+use support::*;
+
+#[cfg_attr(target_os = "android", ndk_glue::main(backtrace = "on"))]
+fn main() {
+    let event_loop = EventLoop::new();
+
+    let raw_display = event_loop.raw_display_handle();
+
+    // TODO: On Android this is can only be retrieved _AFTER_ Event::Resumed, but
```
kchibisov, line 29: Would suggest to resolve this TODO since ti should be trivial.

glutin_examples/examples/android.rs
```diff
@@ -0,0 +1,126 @@
...
+    // it's not needed to create an EGL display. Make this an Option?
+    // let raw_window_handle = window.raw_window_handle();
+
+    // Create the GL display. This will create display automatically for the
+    // underlying GL platform. See support module on how it's being done.
+    let gl_display =
+        create_display(raw_display, RawWindowHandle::AndroidNdk(AndroidNdkWindowHandle::empty()));
+
+    let template = ConfigTemplateBuilder::new()
+        // .compatible_with_native_window(raw_window_handle)
```
kchibisov, line 39: Don't need that.

glutin_examples/examples/android.rs
```diff
@@ -0,0 +1,126 @@
...
+        .with_surface_type(ConfigSurfaceTypes::WINDOW)
```
kchibisov, line 40: The window is default as well.

glutin_examples/examples/android.rs
```diff
@@ -0,0 +1,126 @@
...
+                let raw_window_handle = window.raw_window_handle();
+
+                // Create a wrapper for GL window and surface.
+                let gl_window = GlWindow::from_existing(&gl_display, window, &config);
+
+                // The context creation part. It can be created before surface and that's how
+                // it's expected in multithreaded + multiwindow operation mode, since you
+                // can send NotCurrentContext, but not Surface.
+                let context_attributes =
+                    ContextAttributesBuilder::new().build(Some(raw_window_handle));
```
kchibisov, line 63: You don't need handle here as well. Can pass `None` on build.

kchibisov (changes requested):
I'd suggest you to create context out side of `run`. We should advertise the best we can here. The problem is symbol loading, I guess? I wonder if we can make context current, but without any surface? Make add Api for that or something to help here.

kchibisov:
>I wonder if we can make context current, but without any surface? Make add Api for that or something to help here.

Ah, we can, but it's not guaranteed to work. The issue I was trying to solve is gl functions loading, since I was confused why we need context to be current for that. And it turned out that WGL needs that to perform ICD loading. In general we don't need that if we use `RawWindowHandle` but we do support rendering without passing RawWindowHandle at all. This seems very strange to be fair, and I think that we should force `RawWindowHandle` with WGL.

Though, we can provide a method to get proc address from not current context on supported platforms.

Do you have any idea how to approach the symbol loading here and WGL situation?

MarijnS95:
> I'd suggest you to create context out side of `run`. We should advertise the best we can here. The problem is symbol loading, I guess? I wonder if we can make context current, but without any surface? Make add Api for that or something to help here.

It seems you figured out how to clean this up on the glutin API side in #1460 now, I'll rebase and see where we end up.

glutin_examples/examples/android.rs
MarijnS95: I'll add an option to `create_display()`, and should we provide a "helper" function like this in Glutin? Reason not to do it is leaving the user explicit control over the "display backend" and precedence they use.

kchibisov:
Now you should be able to reuse your context properly.

MarijnS95:
@kchibisov Having a little trouble with that, `make_current()` wants to move the `NotCurrent` context that I create outside of the `run()` loop, but it's not `Copy` nor `Clone` and I can only make it current inside `Resumed` where a `Window` is created.

I can create an `enum` that keeps either the `PossiblyCurrent` or `NotCurrent` context, and then just make it current once we have a `Window`?

MarijnS95:
Implemented it that way, by turning `NotCurrentContext` into `PossiblyCurrentContext` when a window is created, and putting it back in `Suspended`. 

This survives the Android lifecycle.

Also tested on desktop by deleting `crate-type = ["cdylib"]`. @kchibisov should we start with a cleanup patch that moves most of the code out of `window.rs` so that we have less duplication between `window.rs` and `android.rs` when I make these changes to it?

kchibisov:
>Having a little trouble with that, make_current() wants to move the NotCurrent context that I create outside of the run() loop, but it's not Copy nor Clone and I can only make it current inside Resumed where a Window is created.

That's correct, but you can `treat_as_current` if you don't need `NotCurrentContext` thread safety.

>Implemented it that way, by turning NotCurrentContext into PossiblyCurrentContext when a window is created, and putting it back in Suspended.

That's how it should be anyway since you recreate the surface, so you don't have anything current?

>Also tested on desktop by deleting crate-type = ["cdylib"]. @kchibisov should we start with a cleanup patch that moves most of the code out of window.rs so that we have less duplication between window.rs and android.rs when I make these changes to it?

I guess? I don't really mind it that way to be fair since the initialization is a bit different. We can merge it as is.

glutin_wgl_sys/src/lib.rs
```diff
@@ -1,8 +1,10 @@
...
-#![allow(clippy::missing_safety_doc)]
-#![allow(clippy::manual_non_exhaustive)]
-#![allow(clippy::unnecessary_cast)]
+#![cfg(target_os = "windows")]
+#![allow(
+    clippy::too_many_arguments,
+    clippy::missing_safety_doc,
+    clippy::manual_non_exhaustive,
+    clippy::unnecessary_cast
+)]
```
kchibisov, line 7: That seems arbitrary? The old version was intentional here, so revert it.

kchibisov (changes requested):

README.md
MarijnS95: @kchibisov This is the thread where we can unify the handling for Android and desktop.

MarijnS95:
> I guess? I don't really mind it that way to be fair since the initialization is a bit different. We can merge it as is.

This initialization is supposed to work for desktop as well, because they also receive `Event::Resumed`.

kchibisov:
>This initialization is supposed to work for desktop as well, because they also receive Event::Resumed.

Yeah, it's just symbol loading won't work on WGL, but we can load it only once on `Resume`? I wonder if we can have just one example at all or it's not possible due to `cdylib` thingy?

MarijnS95:
> Yeah, it's just symbol loading won't work on WGL, but we can load it only once on `Resume`?

We might be able to `#[cfg()]` that?

> I wonder if we can have just one example at all or it's not possible due to `cdylib` thingy?

Seems we still need two examples, but we can move 95% of the duplicate code in `{window,android}.rs` to a separate file/module (inside `glutin_examples`).

kchibisov:
>Seems we still need two examples, but we can move 95% of the duplicate code in {window,android}.rs to a separate file/module (inside glutin_examples).

That sounds good to me.

glutin_examples/examples/android.rs
```diff
@@ -0,0 +1,118 @@
...
+                    .unwrap()
+                    .make_current(&gl_window.surface)
+                    .unwrap();
+
+                // WGL requires current context on the calling thread to load symbols properly,
+                // so the call here is for portability reasons. In case you don't target WGL
+                // you can call it right after display creation.
+                //
+                // The symbol loading is done by the renderer.
+                let renderer = Renderer::new(&gl_display);
```
MarijnS95, line 63: @kchibisov Before attempting to generalize the event loop into a separate file in the `glutin_examples` lib, I tried to make the example run again.

This comment here seems to indicate that `Renderer::new()` only performs symbol loading - which only needs a current-context on WGL - yet it performs a bunch of shader and buffer setup too which require the context to be current on _all_ platforms :grimacing:

kchibisov (changes requested):
Could you not recreate the renderer on `Resume`? There's no need for that at all and it just slows you down. You also don't need to drop it.

MarijnS95:
@kchibisov It'd be nice if you actually read my questions before shooting this down immediately.

kchibisov:
I mean, I've read it. Nothing stops you from using hand rolled `Lazy`. I can send a patch if you struggle.

MarijnS95:
> I mean, I've read it.

Just asking what your design decision was here, and why the selected comment is incomplete.

> Nothing stops you from using hand rolled `Lazy`.

Yawn.

> I can send a patch if you struggle.

Pls halp I can't Rust /s

glutin_examples/examples/android.rs
kchibisov: I think you can remove this comment given it's documented on `get_proc_address`.

kchibisov:
You may base it on your recently merged PR now.

MarijnS95:
@kchibisov Done, can you recheck that this still works on WGL, especially since WGL needs a window _before_ creating a _display_ according to your comments?

glutin_examples/src/lib.rs
```diff
@@ -41,86 +44,113 @@ pub fn main() {
...
+        .build(raw_window_handle);
+    let mut not_current_gl_context = Some(
+        unsafe { gl_display.create_context(&config, &context_attributes) }.unwrap_or_else(|_| {
+            unsafe { gl_display.create_context(&config, &fallback_context_attributes) }
                 .expect("failed to create context")
-        })
-    };
-
-    // Make it current and load symbols.
-    let gl_context = gl_context.make_current(&gl_window.surface).unwrap();
```
MarijnS95, line 91: @kchibisov If I understood things correctly `make_current()` now doesn't load any functions anymore, that's done by `Renderer::new()` right? Or does it load a limited set of functions?

glutin_examples/src/lib.rs
```diff
@@ -206,10 +242,10 @@ pub fn create_display(
     let preference = DisplayApiPreference::Cgl;
 
     #[cfg(wgl_backend)]
-    let preference = DisplayApiPreference::Wgl(Some(raw_window_handle));
+    let preference = DisplayApiPreference::Wgl(Some(raw_window_handle.unwrap()));
 
     #[cfg(all(egl_backend, wgl_backend))]
-    let preference = DisplayApiPreference::WglThenEgl(Some(raw_window_handle));
+    let preference = DisplayApiPreference::WglThenEgl(Some(raw_window_handle.unwrap()));
```
MarijnS95, line 248: @kchibisov What if we were to pass `None` here?

kchibisov: This could work, but it won't load any extensions, I'd suggest to not do that on WGL. We might change that in the future, but it's sort of hard to figure out the most ergonomic API for WGL here.

tl;dr right now we must pass the right handle here.

MarijnS95: In other words: keep the Some + unwrap?

kchibisov: Yeah. Also be aware that we need raw window handle on X11 otherwise you'd need to pass visual from the config to create winit window. Yeah, that's sort of complicated. In general I think it's safer to use visual from config to create winit x11 window...

See `x11_visual` on window builder.

glutin_examples/src/lib.rs
```diff
@@ -41,86 +44,113 @@ pub fn main() {
...
     let fallback_context_attributes = ContextAttributesBuilder::new()
         .with_context_api(ContextApi::Gles(None))
-        .build(Some(raw_window_handle));
-    let gl_context = unsafe {
-        gl_display.create_context(&config, &context_attributes).unwrap_or_else(|_| {
-            gl_display
-                .create_context(&config, &fallback_context_attributes)
+        .build(raw_window_handle);
+    let mut not_current_gl_context = Some(
+        unsafe { gl_display.create_context(&config, &context_attributes) }.unwrap_or_else(|_| {
```
kchibisov, line 80: Could we have one `unsafe` for all of them? Nesting like that seems weird.

glutin_wgl_sys/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![cfg(any(target_os = "windows"))]
+#![cfg(target_os = "windows")]
```
kchibisov, line 1: You could just `#![cfg(windows)]` here.

MarijnS95: Sure, I'll do that for all `target_os = "windows"` across the repo for consistency.

glutin_examples/src/lib.rs
MarijnS95: @kchibisov Not sure I follow. Searching for `x11_visual` shows references inside GLX's `find_configs()`; in other words you want me to make sure on X11 the window is created before we call `config_template()`, but it's fine to not have it before display creation, for this:

https://github.com/rust-windowing/glutin/blob/ec74687ed2acede620a3cb0d54f85db877725f3c/glutin/src/api/glx/config.rs#L99-L103

Also noticed a little bug that `::Resumed` would open another window on these platforms, even if we already created one (but the first one seems to be invisible on X11).

kchibisov: I'd handle that myself then. I meant the winit API here, sorry for not being clear.

kchibisov (approved):
This should be good to go once CI passes. I've updated the handling of X11 to be more robust given that it was proven that it's not really working.

Not sure if I want to update internal handling of X11 config picking, but might alter it to not consider raw-window-handle at all.

kchibisov:
Thanks.

MarijnS95:
@kchibisov Thank _you_ for patching up the X11 situation and getting this in!

Issue #1271, opened 2020-02-01
Implementation tracking and API discussions for v0.24


goddessfreya:
This issue tracks the implementation of my proposed Glutin 0.23 API as well as any discussion wrt
the API. PRs implementing these changes should be made against the `v0.23_reworks` branch. Changes to winit should be made against the `winit-glutin-next` branch.

To test the changes locally, the following repositories must be 
```
git clone git@github.com:rust-windowing/glutin.git --branch v0.23_reworks
git clone git@github.com:rust-windowing/glutin_interface.git
git clone git@github.com:rust-windowing/glutin_x11_sym.git
git clone git@github.com:rust-windowing/winit_types.git
git clone git@github.com:rust-windowing/glutin_sys.git
git clone git@github.com:goddessfreya/winit.git --branch winit-glutin-next
git clone git@github.com:goddessfreya/gbm.rs.git --branch  glutin-fun
```

Changelog summary: https://github.com/rust-windowing/glutin/blob/v0.23_reworks/CHANGELOG.md
Proposed API:  https://gentz.rocks/files/1580623488/target/doc/glutin/index.html

Implemented for:
- [x] X11 - EGL (#1239)
- [x] X11 - GLX (#1239)
- [x] X11 - GLX/EGL Pixmaps (#1239)
- [x] Linux - EGLMesaSurfaceless (#1239)
- [x] Linux - EGLExtDevice (#1239)
- [x] Linux - GBM (#1239)
- [x] Wayland - EGL (#1239)
- [x] OSMesa (#1239)
- [ ] Windows - EGL
- [ ] Windows - WGL
- [ ] Windows - WGL/EGL Pixmaps
- [ ] MacOS
- [ ] iOS
- [x] Android (In progress: #1274 )

Additional tasks:
- [x] Get CI running for `glutin_sys`, `winit_types`, `glutin_x11_sym`, and `glutin_interface`.
- [ ] Publish version ones of those crates.
- [ ] https://github.com/rust-windowing/winit/pull/1403
- [ ] https://github.com/Smithay/gbm.rs/pull/2

--------

Originally intended to resolve #1140 and #1136, this set of changes has quickly spiraled into also covering #1243, #81, #501, and #466.

I wanted to have at least one implementation of the API before bugging people with it, and, well, now I got two of the major platforms down (Wayland and X11) as well as a handful of smaller targets on Linux.

[...]

PR #1418, opened 2022-06-06
Switch from `ndk-glue` to `raw-window-handle` crate

MarijnS95:
`winit` provides a `raw-window-handle` specifically for crates to process window pointers/references in a crate-agnostic way.  `glutin` can use it too, to get rid of its `ndk-glue` dependency which otherwise has a hard versioning dependency  as it should be updated in sync with `winit`.

Over time this should be worked into the other backends, in turn removing `winit` from the `new_windowed()` function signature and replacing it with `impl HasRawWindowHandle` or `dyn HasRawWindowHandle`.  This will also allow Android users to pass generic `Surface`s (first converted to a so-called `NativeWindow` in the NDK through [#272], then passed as `RawWindowHandle` thanks to [#274]) created in Java / Kotlin / Flutter to render to individual UI elements instead of the entire screen with a `NativeActivity` (`ndk-glue`).  It'll in turn make the code more generic, too, as `raw-window-handle` has a variant for every platform / windowing system currently supported by glutin.

[#272]: https://github.com/rust-windowing/android-ndk-rs/pull/272
[#274]: https://github.com/rust-windowing/android-ndk-rs/pull/274

CC @jamienicol

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 - Return an `Err` instead of panicking when surfaceless GLX context creation fails on Linux.
 - Fix compilation on Android:
   - Switch from `StaticStructGenerator` to `StructGenerator` to dynamically load symbols.
-  - Replace `android_glue` dependency with `ndk-glue`, and remove broken lifecycle event handling.
+  - Replace `android_glue` dependency with `raw-window-handle`, and remove broken lifecycle event handling.
   - Glutin can now be used on Android, however, the application must ensure it only creates the `Context` following a winit `Event::Resumed` event, and destroys the `Context` in response to a `Event::Suspended` event.
 
 # Version 0.28.0 (2021-12-02)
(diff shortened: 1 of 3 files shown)


jamienicol:
This looks good to me. Removing the direct dependency on the glue will make things easier. 👍

MarijnS95:
@jamienicol can you give an explicit review approval or should I just merge this in, given that @kchibisov gave me free reign over the Android guts of glutin?

jamienicol (approved):
LGTM

MarijnS95:
@maroider the same opinion was voiced by @kchibisov. I think we can have isolated cases where `glutin` provides `winit` helpers (ie. to deal with events) but those should perhaps reside behind a feature flag and be few and far between.

maroider:
Do we have an issue outlining these intentions? If not, then we should probably make one.

MarijnS95:
I don't think so, as I was planning to just "get it over with in a PR soon/quickly™", but feel free to create one as I don't know when I can (would like to get current open PRs and issues fleshed out, but I **_do_** have some in-progress changes to migrate to "pure" `RawWindowHandle`).

kchibisov:
@maroider https://github.com/rust-windowing/glutin/issues/1271

PR #1420, opened 2022-06-08
Move Android "api" module to `platform_impl`

MarijnS95:
As seen on the `v0.23_reworks` branch as well, this module doesn't contain a GL API but the platform-specific implementation for Android, built on top of EGL.  It should hence reside in `platform_impl`, which is currently a public reexport of every public item inside `api::android` anyway.


glutin/src/platform_impl/android/mod.rs
```diff
@@ -1,4 +1,164 @@
 #![cfg(target_os = "android")]
 
-pub use crate::api::android::*;
-pub use winit::event_loop::EventLoop;
```
MarijnS95, line 4: Note that I did remove this public reexport of `EventLoop` since it's not used in this file and no other `platform_impl` does it that way.

kchibisov (approved):

PR #1421, opened 2022-06-12
Fix cargo-apk link in README

marc2332:

This basically fixes the link to `cargo-apk` in the README. Otherwise, https://github.com/rust-windowing/android-ndk-rs/cargo-apk returns 404.


MarijnS95 (approved):
Oops!

I don't think the link is very helpful though, but I'll improve that once we have an actual example that works on Android in https://github.com/rust-windowing/glutin/pull/1417.

PR #1422, opened 2022-06-12
examples: Bump `image` dependency from `0.21` to `0.24` to address CVE-2020-35916

MarijnS95:
Dependabot showed this security alert on my fork, which is trivially addressable by bumping the version.

@kchibisov you probably want to turn on dependabot alerts for this repo :)

### DIFF
--- a/glutin_examples/Cargo.toml
+++ b/glutin_examples/Cargo.toml
@@ -13,7 +13,7 @@ publish = false
 [dependencies]
 glutin = { path = "../glutin" }
 takeable-option = "0.4"
-image = "0.21"
+image = "0.24"
 
 [build-dependencies]
 gl_generator = "0.14"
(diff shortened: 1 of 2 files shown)


kchibisov:
>@kchibisov you probably want to turn on dependabot alerts for this repo :)

I'm not sure, since it's spammy, might run cargo audit on demand from time to time. But in any case I don't have control to any settings in any rust-windowing projects, so it's not up to me. I just happened to be the one caring for glutin, since that's what we use in alacritty.

kchibisov (approved):

MarijnS95:
Dependabot is spammy but dependabot-security-alerts hopefully not that much, they only got introduced for Rust last week anyway.

Alacritty is great btw :grimacing:

Issue #1375, opened 2021-11-04
Is there a tutorial for glutin? The example is a bit complicated for me now


mingxingren:

kchibisov:
I'm not sure whether it helps, but you can try a new example. In general glutin has very little to do now other than creating Context.

kchibisov:
I'm pretty sure that the current example shows mostly all platform gotchas you may face as well as showing how to write portable applications that can work on both mobile and desktop.

PR #1427, opened 2022-07-13
docs: clarifying that examples use gl_generator

TGElder:
This would have helped me understand the examples a lot sooner. #1375 also shows that others are having trouble with the examples.

Removed the checklist since this is README only

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Join us in any of these:
 
 Warning: these are examples for master. For the [latest released version](https://github.com/rust-windowing/glutin/releases/tag/v0.27.0)
 
-The examples use [gl_generator](https://crates.io/crates/gl_generator) to generate OpenGL bindings. Try reading the documentation for this if the examples do not make sense.
+The examples use [gl_generator](https://crates.io/crates/gl_generator) to generate OpenGL bindings.
 
 ### Try it!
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -22,6 +22,8 @@ Join us in any of these:
 
 Warning: these are examples for master. For the [latest released version](https://github.com/rust-windowing/glutin/releases/tag/v0.27.0)
 
+The examples use [gl_generator](https://crates.io/crates/gl_generator) to generate OpenGL bindings. Try reading the documentation for this if the examples do not make sense.
```
MarijnS95, line 25: `Try reading the documentation for this if the examples do not make sense.` rubs with me: it doesn't really mean/say anything. Rather would it be helpful to clarify common pitfalls here? I.e. clarify _what exactly_ you had trouble with understanding? How the function pointers were loaded?

MarijnS95: Should be covered in "The Cargo Book": https://doc.rust-lang.org/cargo/reference/build-scripts.html

I don't think it's up to crates to document core Rust concepts though, but alas I'm not a maintainer to have the final say here. Regardless, I don't think there's much harm in the reworded messages, thanks!

kchibisov (approved):

PR #1432, opened 2022-07-29
Switch from `lazy_static` to `once_cell`

a1phyr:
`winit` has just done it, so `glutin` should probably do the same.

### DIFF
--- a/glutin/Cargo.toml
+++ b/glutin/Cargo.toml
@@ -23,7 +23,7 @@ wayland-csd-adwaita = ["winit/wayland-csd-adwaita"]
 wayland-csd-adwaita-notitle = ["winit/wayland-csd-adwaita-notitle"]
 
 [dependencies]
-lazy_static = "1.3"
+once_cell = "1.10"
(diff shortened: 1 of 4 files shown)


glutin/Cargo.toml
```diff
@@ -23,7 +23,7 @@ wayland-csd-adwaita = ["winit/wayland-csd-adwaita"]
 wayland-csd-adwaita-notitle = ["winit/wayland-csd-adwaita-notitle"]
 
 [dependencies]
-lazy_static = "1.3"
+once_cell = "1.10"
```
kchibisov, line 26: Could you specify the current latest version instead?

kchibisov (changes requested):

MarijnS95 (approved):
Other than the versioning issue (I'd personally write `"1"`), this seems fine to me. I thought we already did this conversion in Glutin or I forgot to submit it :sweat_smile:

kchibisov (approved):

PR #1433, opened 2022-07-30
Test rust 1.57 on CI

kchibisov:


glutin/src/context.rs
```diff
@@ -77,6 +77,7 @@ impl Context<PossiblyCurrent> {
     }
 }
 
+#[allow(rustdoc::broken_intra_doc_links)]
```
MarijnS95, line 80: Why is this? Can't we fix the links? :disappointed:

kchibisov: The rust 1.57 bugs here. The stable passes for example.

MarijnS95: That's ugly, means mistakes won't be caught on `stable` either.

MarijnS95: Why anyway do we build `rustdoc` on the MSRV CI step?

kchibisov: Well, given that's the latest release you see this code... I'm actually not sure why are we using the msrv for docs. Might run those on stable.

PR #1435, opened 2022-07-30
Decouple glutin from winit

kchibisov:
Thit commit removes direct dependency on the winit and
using raw_window_handle crate instead

--

This is a draft for now given that macOs and Windows support is missing(and also ios, but I don't have a way to add it anyway). The examples are also lacking. However the public Api are unlikely to change (besides adding more config options).

The feedback is welcome!


glutin_wgl_sys/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![cfg(any(target_os = "windows"))]
+#![allow(clippy::all)]
```
MarijnS95, line 2: :eyes:

maroider: I'm guessing this is a temporary thing?

kchibisov: I'm not sure what the approach should be for generated bindings. But in general it's common to disable all clippy lints for them.

MarijnS95: I typically disable those that are strictly causing warnings. Anything new (i.e. after a `rustup update`) and I either see if there's an update to the bindings generator, or report (if there's no report yet) that they aren't generating clippy-compliant code (such as bindgen doing a lot of debatable pointer magic in their tests that clippy didn't like).

madsmtm:
I could probably do the macOS and iOS impls

kchibisov:
I'm not quite sure how I keep the Api, but ios impl would be very welcome, since I don't really mind doing macOS one.

madsmtm:
> I'm not quite sure how I keep the Api, but ios impl would be very welcome, since I don't really mind doing macOS one.

'aight, I think I'll wait until you have - I have to get a newer device set up anyhow, which will take a little while. While I can (since https://github.com/rust-windowing/glutin/pull/1441) run `glutin` on my old iPad, it crashes for some reason I haven't (yet) figured out

kchibisov:
I've pushed one more update and kind of settled on the final Api I want to use.

The current version has stabs for macOS/Wgl with `todo!` to implement. Similar could be done to `ios`.

In general once feeling the `todo!` for both platforms in `api/wgl` and `api/cgl` they should just work in the current infra.

All the `cfg` stuff and `features` is already wired up. And the directory layout is unlikely to change.

kchibisov:
@madsmtm added macOS support, but something maybe missing here and there. Had to adjust Api a bit for it to works but at least example should show window.

glutin/src/api/egl/display.rs
```diff
@@ -0,0 +1,286 @@
...
+            {
+                (egl::PLATFORM_WAYLAND_KHR, handle.display)
+            }
+            #[cfg(feature = "x11")]
+            RawDisplayHandle::Xlib(handle) if extensions.contains("EGL_KHR_platform_x11") => {
+                attrs.push(egl::PLATFORM_X11_SCREEN_KHR as EGLAttrib);
+                attrs.push(handle.screen as EGLAttrib);
+                (egl::PLATFORM_X11_KHR, handle.display)
+            }
+            RawDisplayHandle::Gbm(handle) if extensions.contains("EGL_KHR_platform_gbm") => {
```
i509VCB, line 127: did we want to put a comment for `EGL_KHR_platform_xcb` or do you think that will never actually be implemented by any drivers?

glutin/src/lib_loading.rs
```diff
@@ -0,0 +1,46 @@
...
+use libloading::Library;
+
+#[derive(Clone)]
+pub struct SymWrapper<T> {
+    sym: T,
+    _lib: Arc<Library>,
+}
+
+pub trait SymLoading {
+    fn load_with(lib: &Library) -> Self;
```
i509VCB, line 15: This is actually unsafe?

The library could be dropped after this and unload any symbols you read from the library.

glutin/src/api/egl/display.rs
kchibisov: We can implement it here, I just haven't read extension for it and don't have a test client, but I don't see why we can avoid it.

glutin/src/lib_loading.rs
kchibisov: I haven't given much thoughts into about it, but it's internal Api that is used to load libraries into the globals. If you have suggestions on how to approach it better (or link to a code that does that, let me know).

Though in general I don't see anything wrong here in particular given that calling ffi is already `unsafe`, as well as calling `gl` apis for example, the caller should ensure that those are there.

i509VCB: Well I'd argue this should be unsafe because undefined behavior could occur without any compiler garuntees. Now calling eglCreateContext internally and etc would be safe because you have ensured the assumption that the library those functions come from is loaded.

Sure if the type had a lifetime over the library it would be fine, but it does not.

rib:
Just made a first pass through the changes here and they definitely look like a good direction from my pov, while I'm currently stuck with being able to update the `egui/eframe` glow backend to destroy and recreate surface state on suspend + resume on Android, considering how Glutin currently tightly couples the Context with the Window.

Intuitively this is more the kind of design I had been expecting to find with a separate Surface type and then having `make_current()` accept a surface.

glutin/src/api/egl/display.rs
```diff
@@ -0,0 +1,286 @@
...
+    pub(crate) raw_display: RawDisplayHandle,
+}
+
+impl Display {
+    /// Create EGL display.
+    ///
+    /// # Safety
+    ///
+    /// `RawDisplay` must point to a valid system display.
+    pub unsafe fn from_raw(raw_display: RawDisplayHandle) -> Result<Self> {
```
i509VCB, line 67: Would be nice to have a `from_raw` function that lives as a guest on the display pointer. This would probably need to be renamed

@technobaboo you might have some comments on this since I am evaluating using the reworked glutin in smithay in the future and you use egl as a guest from stereokit.

i509VCB, line 67: Possibly a function as well to create a display from an EGLPlatform enum and EGLNativeDisplay.

glutin/src/api/egl/display.rs
```diff
@@ -0,0 +1,286 @@
...
+impl Sealed for Display {}
+
+impl AsRawDisplay for Display {
+    fn raw_display(&self) -> RawDisplay {
+        RawDisplay::Egl(self.inner.raw)
+    }
+}
+
+/// Collect egl extensions for the given `display`.
+fn get_extensions(egl: &Egl, display: EGLDisplay) -> HashSet<&'static str> {
```
i509VCB, line 273: Would be nice to have the extension list be public for egl.

kchibisov: Yeah' want to expose that as well, not sure how though in a more reasonable way.

glutin/src/api/egl/display.rs
technobaboo: I can't see any issues with it so far!

kchibisov: I want to add generic methods to build from raw, just not sure how we should approach it. I know that we could keep a pointer, but how about `Config` given that I provide a way to retrieve it or display?

That would mean that all the Api should return an `Option` for those?

I'm not quite sure how to evaluate it.

glutin/src/context.rs
```diff
@@ -1,168 +1,405 @@
...
 }
 
-impl Context<PossiblyCurrent> {
-    /// See [`ContextWrapper::get_proc_address()`].
-    pub fn get_proc_address(&self, addr: &str) -> *const core::ffi::c_void {
-        self.context.get_proc_address(addr)
+pub trait NotCurrentGlContext: Sealed {
+    type PossiblyCurrentContext: PossiblyCurrentGlContext;
+
+    fn treat_as_current(self) -> Self::PossiblyCurrentContext;
```
i509VCB, line 186: Should this be unsafe?

glutin/src/platform/x11.rs
```diff
@@ -0,0 +1,81 @@
...
+pub trait X11GlConfigExt {
+    // The `X11VisualInfo` that must be used to inititalize the Xlib window.
+    fn x11_visual(&self) -> Option<X11VisualInfo>;
+}
+
+/// The X11 visual info.
+///
+/// This must be used when building X11 window, so it'll be compatible with the underlying Api.
+pub struct X11VisualInfo {
+    display: *mut Display,
```
i509VCB, line 23: How do we assert the display is valid for the lifetime of the visual info?

glutin/src/surface.rs
```diff
@@ -0,0 +1,421 @@
...
+pub enum NativePixmap {
+    /// XID of X11 pixmap.  
+    X11Pixmap(u64),
+
+    /// HBITMAP handle for windows bitmap.
+    WindowsPixmap(isize),
+}
+
+impl NativePixmap {
+    pub fn raw(&self) -> *mut ffi::c_void {
```
i509VCB, line 385: I don't think this is very useful.

glutin/src/surface.rs
```diff
@@ -0,0 +1,421 @@
...
+impl ResizeableSurface for WindowSurface {}
+
+// A platform native pixmap.
+#[derive(Debug, Clone, Copy)]
+pub enum NativePixmap {
+    /// XID of X11 pixmap.  
+    X11Pixmap(u64),
+
+    /// HBITMAP handle for windows bitmap.
+    WindowsPixmap(isize),
```
i509VCB, line 381: windows-sys defines the type as that?

glutin/src/surface.rs
```diff
@@ -0,0 +1,421 @@
...
+    }
+}
+
+impl ResizeableSurface for WindowSurface {}
+
+// A platform native pixmap.
+#[derive(Debug, Clone, Copy)]
+pub enum NativePixmap {
+    /// XID of X11 pixmap.  
+    X11Pixmap(u64),
```
i509VCB, line 378: Aren't xids a u32?

glutin/src/context.rs
kchibisov: No. This is safe, we cast to more restrictive type. `NotCurrentContext` is the context that is known to be not current, so it can be `Send`. `PossiblyCurrentContext` could still be not current.

glutin/src/surface.rs
kchibisov: I think it's defined as `c_ulong` in header, so that's why I have it that way. But yeah, `XID` is `u32`.

glutin/src/surface.rs
kchibisov: Yes, just code was written before `windows-sys` added as dependency.

glutin/src/surface.rs
kchibisov: Probably.

glutin/src/platform/x11.rs
kchibisov: That's the problem in general with RawDisplayHandle, since it could in theory be dropped. We could open inner connection and has our own XDisplay. I'm actually not entirely sure how to handle that other than open another connection?

glutin/src/api/egl/display.rs
i509VCB: I think we can provide a config during construction if needed.

glutin/src/api/cgl/context.rs
```diff
@@ -0,0 +1,243 @@
...
+                let is_equal: BOOL = msg_send![current, isEqual: self.inner.raw];
+                is_equal != NO
+            } else {
+                false
+            }
+        })
+    }
+
+    fn get_proc_address(&self, addr: &CStr) -> *const ffi::c_void {
+        let symbol_name = CFString::new(addr.to_str().unwrap());
```
i509VCB, line 148: Considering https://github.com/openglonmetal/MGL exists, did we want to allow specifying the library to load from?

kchibisov: No. It uses different context, so it's more like a different platform here. So it'll be `mgl` for example. We`re using `NSOpenGLContext` which is not compatible with mgl context.

glutin/src/api/egl/display.rs
technobaboo: I personally use it because OpenXR has fairly specific configs it wants, so passing it in definitely needs to be an option.

kchibisov:
I've added WGL support. So at this point glutin should be able to show window on every platform. One issue left here is that error reporting is lacking, since I'm not quite sure what to expect from it.

The next step is to add missing context features like api picking, es context, and vsync handling.

Also need to reiterate config picking and options we pass to it. Maybe need enum instead of `bool` to have more control on what is happening.

glutin/src/surface.rs
kchibisov: Yeah it's `c_ulong` in `raw-window-handle` as well and that's because it's `typedef unsigned long XID;` in `X.h`. But yeah, on the wire it's `CARD32`.

glutin/src/surface.rs
kchibisov: I think it's that way to reduce the amount of `cfg` and to akin to `raw-window-handle` here os it's available on all platforms.

glutin/src/surface.rs
kchibisov: I've added it for EGL, moved to the EGL directly.

.github/workflows/ci.yml
```diff
@@ -21,10 +21,10 @@ jobs:
     - uses: actions/checkout@v1
     - uses: hecrj/setup-rust-action@v1
       with:
-        rust-version: stable
+        rust-version: nightly
         components: rustfmt
     - name: Check Formatting
-      run: cargo +stable fmt --all -- --check
+      run: cargo +nightly fmt --all -- --check
```
i509VCB, line 27: I'm a bit concerned with nightly formatting differing from stable formatting of some options are enabled.

kchibisov: I'm using nightly formatter and I don't want to use stable formatting. I explicitly specify all the options I care about. Stable can't help me here and kind of useless.

glutin_wgl_sys/src/lib.rs
kchibisov: Made it use only necessary ones.

kchibisov:
This should be in a good shape and has coverage for everything old `glutin` has and more.

The public Api could be not optimal in some ways and got change a bit due to `WGL` being absolutely horrible(had to add methods to pass `RawWindowHandle` around, because Windows can't live without HWND/HDC), but all in all it's sort of **Ok** from my point of view.

## Public Api refactoring

The Api is based around `Display`, `Config`, `Surface`, `NotCurrentContext` and `PossiblyCurrentContext`.

All of them are created from the `Display` and `unsafe` since they can accept raw pointers, the methods on the created objects safe though.

The thread safety was reworked and now is more strict compared to old glutin, which had unsound Api here. So now

`Display` -> `Send` + `Sync`.
`Config` -> `Send` + `Sync`
`Surface` -> `!Send` + `!Sync`, and that's where old glutin was broken. You can't move EGLSurface accross thread boundaries for example.
`NotCurrentContext` -> `Send` + `!Sync`, you can send only not current context.
`PossiblyCurrentContext` -> `!Send` + `!Sync`, type safe wrapper to prevent moving current context across the threads.

Every Api/Windowing platform is optional and could be excluded. The Api(e.g. EGL) could also be used directly without the top level `Display` type, however given that every top level type is a `enum` the call to the Api specific extension could be made directly trivially by matching on it.

One more note, is that ios support is currently dropped, since it wasn't really working before from what I can see, and there's no maintainer for it.

## Internal structure 

The internal structure was designed with maintainability and extensibility in mind. For example each specific extension could be added right on the Api implementation without the requirement of adding it to other Api's (see `swap_buffers_with_damege`).

The parity between the Api implementations is being kept with the help of traits: `GlDisplay`, `GlSurface`, `GlConfig`, `NotCurrentGlContext` and others, so each Api implements those traits. So if the functionality that was specific to e.g. EGL is present on other platforms, it could be moved to one of those traits. The traits are sealed, since they are purely for maintainability reasons, and there're no plans to unseal them due to limited not growing set of apis that exist and do work.

The thread safety is ensured in doc tests, so `Surface` won't go `Send` for example and if something should be ensured that not `Send` or `Sync` it should be done in tests as well.

The directory structure is trivial, the Api impls are in `src/api`, the platform specific code(e.g. X11 visual) is inside `src/platform`, and the top level structures are in `src/*.rs`.

## Side notes

I'm not that much familiar with macOS and windows development. The implementations were tested (**EXCEPT Windows + EGL, someone should test it at least**), but the code could be missing some extra checks or not safe. Also I have no clue how to check for errors on macOS (old glutin also skipped that part).

The code review is very welcome, but I don't insist, since we can always fix things in retrospective during beta releases.

cc @MarijnS95 , @rib  - android
cc @madsmtm - macOS
cc @maroider - GLX/EGL
cc @msiglreith - windows WGL + EGL.
cc @i509VCB - EGL

glutin/src/platform/x11.rs
```diff
@@ -0,0 +1,95 @@
...
+use x11_dl::xlib::{VisualIDMask, XID};
+use x11_dl::xrender::Xrender;
+
+/// The XLIB handle.
+pub(crate) static XLIB: Lazy<Option<Xlib>> = Lazy::new(|| Xlib::open().ok());
+
+/// The XRENDER handle.
+static XRENDER: Lazy<Option<Xrender>> = Lazy::new(|| Xrender::open().ok());
+
+/// The GlConfig extension trait to get X11 options out of the config.
```
i509VCB, line 17:
```suggestion
/// The GlConfig extension trait to get X11 specific properties from a config.
```

glutin/src/prelude.rs
```diff
@@ -0,0 +1,17 @@
+//! The glutin prelude.
+//!
+//! The purpose of this module is to bring common imports, given that all
+//! graphics api are on the traits for the documetation sharing purposes.
```
i509VCB, line 4: Attempted rewording using the unix prelude from Rust for reference.

```suggestion
//! The purpose of this module is to make accessing common imports more convenient.
//! The prelude also imports traits shared by the implementations of graphics apis.
```

glutin/src/surface.rs
```diff
@@ -0,0 +1,501 @@
...
+    /// Check whether the surface is the current read surface to the current
+    /// thread.
+    fn is_current_read(&self, context: &Self::Context) -> bool;
+
+    /// Set swap interval for the surface.
+    ///
+    /// See the docs for [`crate::surface::SwapInterval`] on the details.
+    fn set_swap_interval(&self, context: &Self::Context, interval: SwapInterval) -> Result<()>;
+
+    /// Resize the surface to the new size.
```
i509VCB, line 64:
```suggestion
    /// Resize the surface to a new size.
```

glutin/src/surface.rs
```diff
@@ -0,0 +1,501 @@
...
+    /// thread.
+    fn is_current_draw(&self, context: &Self::Context) -> bool;
+
+    /// Check whether the surface is the current read surface to the current
+    /// thread.
+    fn is_current_read(&self, context: &Self::Context) -> bool;
+
+    /// Set swap interval for the surface.
+    ///
+    /// See the docs for [`crate::surface::SwapInterval`] on the details.
```
i509VCB, line 61:
```suggestion
    /// See [`crate::surface::SwapInterval`] for details.
```

glutin/src/surface.rs
```diff
@@ -0,0 +1,501 @@
...
+        Default::default()
+    }
+
+    /// Specify whether the surface should support srgb or not. Passing `None`
+    /// means you don't care.
+    ///
+    /// # Api-specific.
+    ///
+    /// This only controls EGL surfaces, since the rest are using context for
+    /// that.
```
i509VCB, line 111:
```suggestion
    /// This only controls EGL surfaces, other platforms use the context for that.
```

glutin/src/surface.rs
```diff
@@ -0,0 +1,501 @@
...
+    /// Specify whether the single buffer should be used instead of double
+    /// buffering. This doesn't guarantee that the resulted buffer will have
+    /// only single buffer, to know that the single buffer is actually used
+    /// query the created surface with [`Surface::is_single_buffered`].
+    ///
+    /// The surface is requested as double buffered by default.
+    ///
+    /// # Api-specific.
+    ///
+    /// This is EGL specific, since the rest are using it on the context.
```
i509VCB, line 128:
```suggestion
    /// This is EGL specific other platforms use the context for that.
```

glutin/src/surface.rs
```diff
@@ -0,0 +1,501 @@
...
+    pub fn with_single_buffer(mut self, single_buffer: bool) -> Self {
+        self.attributes.single_buffer = single_buffer;
+        self
+    }
+
+    /// Build the surface attributes suitable to create a window surface.
+    pub fn build(
```
i509VCB, line 135: Does this need to be unsafe?

glutin_egl_sys/src/lib.rs
```diff
@@ -28,22 +25,26 @@ pub mod egl {
     pub type NativeWindowType = super::EGLNativeWindowType;
 
     include!(concat!(env!("OUT_DIR"), "/egl_bindings.rs"));
+
+    // TODO should upstream this.
```
i509VCB, line 29: This would probably require fixing up gl_generator, which is abandoned sadly.

I tried to see what would happen if I updated egl.xml and gl_generator failed to generate code.

glutin_examples/examples/window.rs
```diff
@@ -1,37 +1,97 @@
...
+    let raw_window_handle = window.raw_window_handle();
 
-    println!("Pixel format of the window's GL context: {:?}", windowed_context.get_pixel_format());
+    // Create the GL display. This will create display automatically for the
+    // underlying GL platform. See support module on how it's being done.
+    let gl_display = create_display(raw_display, raw_window_handle);
 
-    let gl = support::load(windowed_context.context());
+    // Create the config we'll be used for window. We'll use the native window
+    // raw-window-handle for it to get the rigth visual and use proper hdc. Note
```
i509VCB, line 34:
```suggestion
    // raw-window-handle for it to get the right visual and use proper hdc. Note
```

glutin_examples/examples/window.rs
```diff
@@ -1,37 +1,97 @@
...
+        *control_flow = ControlFlow::Wait;
         match event {
-            Event::LoopDestroyed => (),
             Event::WindowEvent { event, .. } => match event {
-                WindowEvent::Resized(physical_size) => windowed_context.resize(physical_size),
-                WindowEvent::CloseRequested => *control_flow = ControlFlow::Exit,
+                WindowEvent::Resized(size) => {
+                    if size.width != 0 && size.height != 0 {
+                        // Some platforms like EGL require resizing GL surface to update the size
+                        // Notable platforms here are Wayland and macOS, other don't require it
```
i509VCB, line 71:
```suggestion
                        // Notable platforms here are Wayland and macOS, others don't require it
```

glutin/src/lib.rs
```diff
@@ -1,664 +1,91 @@
-//! The purpose of this library is to provide an OpenGL [`Context`] on as many
-//! platforms as possible.
+//! The purpose of this library is to provide OpenGL `[context]` on as many
```
i509VCB, line 1: Would it be worthwhile to add `#![deny(unsafe_op_in_unsafe_fn)]`?
Also I think we have a extern "C" function typedefs in the code, so `#![deny(improper_ctypes, improper_ctypes_definitions)]`

glutin/src/surface.rs
kchibisov: No, building attributes is safe, creating a surface from them is unsafe, and that's how it is.

glutin/src/context.rs
```diff
@@ -1,168 +1,584 @@
...
+            },
+            #[cfg(wgl_backend)]
+            (Self::Wgl(context), Surface::Wgl(surface)) => {
+                Ok(PossiblyCurrentContext::Wgl(context.make_current(surface)?))
+            },
+            #[cfg(cgl_backend)]
+            (Self::Cgl(context), Surface::Cgl(surface)) => {
+                Ok(PossiblyCurrentContext::Cgl(context.make_current(surface)?))
+            },
+            _ => unreachable!(),
```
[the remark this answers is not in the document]
kchibisov: I know, we just not support this case. I'm not even sure that you can mix any of that.

glutin/src/display.rs
```diff
@@ -0,0 +1,502 @@
...
+/// Get the raw handle to the [`Display`].
+pub trait AsRawDisplay {
+    /// A raw handle to the underlying Api display.
+    fn raw_display(&self) -> RawDisplay;
+}
+
+/// The graphics display to handle underlying graphics platform in a
+/// cross-platform way.
+///
+/// The display could be accessed from any thread.
```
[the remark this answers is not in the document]
kchibisov: It does support moving context around. There's also no dedicated display on macOS. Do you mean that the context must be created on the main thread and then moved to other thread on macOS?

glutin/src/lib.rs
```diff
@@ -1,664 +1,91 @@
...
 //!
-//! # Building a [`WindowedContext<T>`]
+//! However the library doesn't force into using a cross platform abstraction,
+//! for example, when only `[EGL]` is desired, it can be used directly.
 //!
-//! A [`WindowedContext<T>`] is composed of a [`Window`] and an OpenGL
-//! [`Context`].
+//! The initialization starts by loading and connecting to the underlying
+//! graphics platform Api when creating a `[display]`. This object is used to
+//! create all the OpenGL objects, such as `[config]`, `[context]`, and
```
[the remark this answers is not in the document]
kchibisov: Yeah, I'm just not sure how to call them.

glutin/src/lib_loading.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+#[derive(Clone)]
+pub struct SymWrapper<T> {
+    sym: T,
+    _lib: Arc<Library>,
+}
+
+impl<T: SymLoading> SymWrapper<T> {
+    pub fn new(lib_paths: &[&str]) -> Result<Self, ()> {
```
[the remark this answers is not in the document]
kchibisov: I mean you load what you've passed? I don't see this being unsafe? This is a helper to load a random library with a given name. If you pass `libvulkan.dll' you'll load it, but you've passed it?

glutin_egl_sys/src/lib.rs
kchibisov: Yeah, it's sad that it's dead. There's some other generator, but I forgot its name. It's used by wgpu or glow, don't remember.

glutin/src/display.rs
```diff
@@ -0,0 +1,502 @@
...
+        self.window_handle = Some(window_handle);
+        self
+    }
+
+    /// The hook to register glutin error handler in X11 error handling
+    /// function.
+    ///
+    /// The hook registrar must be provided in case GLX will be used.
+    #[cfg(glx_backend)]
+    pub fn with_glx_error_registrar(mut self, error_registrar: XlibErrorHookRegistrar) -> Self {
```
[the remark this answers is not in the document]
kchibisov: Setting it is safe, creating display with it is `unsafe` and display creation api's are all unsafe.

glutin/src/lib_loading.rs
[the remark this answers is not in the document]
i509VCB: Hmm, so symbols will be fine, I guess the next thing is to look at what libloading says: https://docs.rs/libloading/0.7.3/libloading/struct.Library.html

glutin/src/display.rs
```diff
@@ -0,0 +1,502 @@
...
+    /// Create a graphics platform display from the given raw display handle.
+    ///
+    /// # Safety
+    ///
+    /// The `display` must point to the valid platform display.
+    pub unsafe fn from_raw(display: RawDisplayHandle, picker: DisplayPicker) -> Result<Self> {
+        #[cfg(glx_backend)]
+        let registrar = picker
+            .glx_error_registrar
+            .expect("glx was requested, but error hook registrar wasn't provided.");
```
i509VCB, line 162: I've tried using glutin locally with only EGL and this will be panic if only EGL is requested but the platform supports GLX (which will occur on Linux by default).

glutin/src/api/cgl/config.rs
```diff
@@ -0,0 +1,230 @@
...
+use crate::private::Sealed;
+
+use super::display::Display;
+
+impl Display {
+    pub(crate) unsafe fn find_configs(
+        &self,
+        template: ConfigTemplate,
+    ) -> Result<Box<dyn Iterator<Item = Config> + '_>> {
+        let mut attrs = Vec::<u32>::with_capacity(32);
```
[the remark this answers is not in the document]
kchibisov: Hm, we can add that, old glutin had it for example and I wasn't sure whether it was really needed. Seems like we should.

glutin/src/api/egl/config.rs
```diff
@@ -0,0 +1,377 @@
...
+        if raw_ty & egl::PIXMAP_BIT as u32 != 0 {
+            ty.insert(ConfigSurfaceTypes::PIXMAP);
+        }
+
+        ty
+    }
+
+    fn api(&self) -> Api {
+        let mut api = Api::empty();
+        let raw_api = self.raw_attribute(egl::RENDERABLE_TYPE as EGLint) as u32;
```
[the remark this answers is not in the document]
kchibisov: I'm not familiar with OpenCL at all, so we should have it addressed later, maybe have it as a separate platform, don't know how it works, sorry.

glutin/src/api/egl/display.rs
```diff
@@ -0,0 +1,349 @@
...
+
+impl Display {
+    /// Create EGL display with the native display.
+    ///
+    /// # Safety
+    ///
+    /// `raw_display` must point to a valid system display. Using zero or
+    /// `[std::ptr::null]` for the display will result in using
+    /// `EGL_DEFAULT_DISPLAY`, which is not recommended or will
+    /// work on a platfrom with a concept of native display, like Wayland.
```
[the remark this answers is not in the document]
kchibisov: I wonder how we can avoid that, maybe we should force to pass `Arc` at least? it's kind of sad that we should say that you should keep that everything alive until all the objects are gone.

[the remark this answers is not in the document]
i509VCB: Well for things like the display handle and window handle this requirement already applies.

glutin_egl_sys/src/lib.rs
i509VCB: I think it's phosphorus, but it doesn't support egl.xml yet. It's used to generate the gl46 and gles32 crates along with glow's functions.

glutin/src/lib.rs
kchibisov: Added them.

glutin_egl_sys/src/lib.rs
kchibisov: That's sad. Well, We can leave for now with manually defined platform.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Unreleased
 
+- Replace `winit` dependency with `raw-window-handle`.
+- The Api is now built around `Display`, `Surface`, `Config`, and `Surface` for more info see crate documentation and examples.
+- The ios support was removed for the lack of maintainance for now. In case there's a need for it, contributions are welcome.
+- The context creation is no longer limited to winit's supported platforms.
+- The underlying Api providers are publically exposed now, so glutin could be used with just e.g. `EGL`.
+- Fixed soundness issues with `Surface` MT safety, since before `EGLSurface` could be sent to a different thread, wich is not safe.
```
i509VCB, line 8:
```suggestion
- Fixed soundness issues with `Surface` MT safety, since before `EGLSurface` could be sent to a different thread, which is not safe.
```

glutin/src/display.rs
kchibisov: Should be fixed now.

glutin/src/api/glx/config.rs
```diff
@@ -0,0 +1,352 @@
...
+        if raw_ty & glx::PIXMAP_BIT as u32 != 0 {
+            ty.insert(ConfigSurfaceTypes::PIXMAP);
+        }
+
+        ty
+    }
+
+    fn api(&self) -> Api {
+        let mut api = Api::OPENGL;
+        if self.inner.display.inner.client_extensions.contains("GLX_EXT_create_context_es2_profile")
```
i509VCB, line 272: GLX defines the following constants:
```
	GLX_CONTEXT_ES_PROFILE_BIT_EXT		0x00000004
	GLX_CONTEXT_ES2_PROFILE_BIT_EXT		0x00000004
```

In theory this means that the extension being present also implies `Api::GLES`.

Long story short, Khronos did a mistake and we have the GLX_EXT_create_context_es_profile name as well.

WGL has a similar case as well.

glutin/src/api/glx/context.rs
```diff
@@ -0,0 +1,420 @@
...
+                // Direct context
+                1,
+                attrs.as_ptr(),
+            ))
+        }
+    }
+
+    fn create_context_legacy(&self, config: &Config, shared_context: GLXContext) -> GLXContext {
+        let render_type =
+            if config.float_pixels() { glx_extra::RGBA_FLOAT_TYPE_ARB } else { glx::RGBA_TYPE };
```
i509VCB, line 202: `RGBA_FLOAT_TYPE_ARB` is not guarded by a test for the extension?

glutin/src/api/glx/display.rs
```diff
@@ -0,0 +1,215 @@
...
+            _ => {
+                return Err(
+                    ErrorKind::NotSupported("provided native display isn't supported").into()
+                )
+            },
+        };
+
+        let glx = match GLX.as_ref() {
+            Some(glx) => glx,
+            None => return Err(ErrorKind::NotFound.into()),
```
i509VCB, line 59: Would a special error variant for failing to load the library/required symbols make sense?

glutin/src/api/glx/surface.rs
```diff
@@ -0,0 +1,302 @@
...
+        }
+    }
+}
+
+impl<T: SurfaceTypeTrait> GlSurface<T> for Surface<T> {
+    type Context = PossiblyCurrentContext;
+    type SurfaceType = T;
+
+    fn buffer_age(&self) -> u32 {
+        self.raw_attribute(glx_extra::BACK_BUFFER_AGE_EXT as c_int) as u32
```
i509VCB, line 196: The implementation of `raw_attribute` says that it should never fail. But what happens if the extension providing BACK_BUFFER_AGE_EXT is not present?

glutin/src/api/glx/config.rs
```diff
@@ -0,0 +1,352 @@
...
+                self.inner.display.inner.raw.cast(),
+                *self.inner.raw,
+                attr,
+                &mut val,
+            );
+            val as c_int
+        }
+    }
+
+    pub(crate) fn is_singe_buffered(&self) -> bool {
```
i509VCB, line 196:
```suggestion
    pub(crate) fn is_single_buffered(&self) -> bool {
```

glutin/src/api/wgl/config.rs
```diff
@@ -0,0 +1,512 @@
...
+            } else {
+                Ok(())
+            }
+        }
+    }
+
+    pub(crate) fn is_single_buffered(&self) -> bool {
+        match self.inner.descriptor.as_ref() {
+            Some(descriptor) => (descriptor.dwFlags & gl::PFD_DOUBLEBUFFER) == 0,
+            None => self.raw_attribute(wgl_extra::DOUBLE_BUFFER_ARB as c_int) == 0,
```
i509VCB, line 296: Does the implementation return 0 if DOUBLE_BUFFER_ARB is not available due to missing extensions.

glutin/src/api/egl/display.rs
```diff
@@ -0,0 +1,351 @@
...
+                (egl::PLATFORM_WAYLAND_EXT, handle.display)
+            },
+            #[cfg(x11_platform)]
+            RawDisplayHandle::Xlib(handle) if extensions.contains("EGL_EXT_platform_x11") => {
+                attrs.push(egl::PLATFORM_X11_SCREEN_EXT as EGLint);
+                attrs.push(handle.screen as EGLint);
+                (egl::PLATFORM_X11_EXT, handle.display)
+            },
+            #[cfg(x11_platform)]
+            RawDisplayHandle::Xcb(handle) if extensions.contains("EGL_MESA_platform_xcb") => {
```
i509VCB, line 167: Already mentioned on matrix, also need to test for `EGL_EXT_platform_xcb`. That's the actual extension because Mesa did an oops. But we should continue to match the mesa name since people are running systems with the old name.

glutin/src/api/glx/context.rs
kchibisov: It does. It's inside config.

glutin/src/api/wgl/config.rs
kchibisov: There's no extension for it besides the `WGL_ARB_pixel_format`, which is what we use when we don't have `descriptor`.

glutin_examples/examples/support/mod.rs
```diff
@@ -1,365 +1,91 @@
...
+pub fn surface_attributes(window: &Window) -> SurfaceAttributes<WindowSurface> {
+    let (width, height): (u32, u32) = window.inner_size().into();
+    let raw_window_handle = window.raw_window_handle();
+    SurfaceAttributesBuilder::<WindowSurface>::new().build(
+        raw_window_handle,
+        NonZeroU32::new(width).unwrap(),
+        NonZeroU32::new(height).unwrap(),
+    )
 }
-\0";
```
madsmtm, line 131: Could we re-add an example with shaders?

madsmtm (approved):
I have quite limited experience with OpenGL, but seems to work on my machine: MacBook Pro (Retina, 13-inch, Late 2013), macOS Mojave 10.14.6.

kchibisov:
>I have quite limited experience with OpenGL, but seems to work on my machine: MacBook Pro (Retina, 13-inch, Late 2013), macOS Mojave 10.14.6.

I mean I know that it works on it. I was asking more for semantics of obj-c handling, etc.

glutin_examples/examples/support/mod.rs
kchibisov: I don't think there's a major value in it? it's just more code to have around, but if anyone wants it they can.

PR #1437, opened 2022-08-10
Update glutin_egl_sys to 0.1.6

paulotten:


kchibisov:
Thanks, but given that the patch is simple, but doesn't address all the issues that could happen I've opened #1438.

PR #1438, opened 2022-08-10
Explicitly specify minimum crate versions

kchibisov:
This should fix building from crates.io

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fix build failures when building from crates.io
+
 # Version 0.29.0 (2022-07-30)
 
 - Fix crash when creating OpenGLES context without explicit version.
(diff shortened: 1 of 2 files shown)


MarijnS95 (approved):
> This should fix building from crates.io

I don't think it fixes all that much given that `cargo` automatically picks the latest semver-compatible version, but it definitely helps when a certain (older) version is already "locked" in place in `Cargo.lock` and users don't run `cargo update` all that often.

Despite that I agree we should make sure the minimum required version is set in `Cargo.toml` if we know what that is - but at the same time shouldn't unnecessarily fully-qualify versions when that hasn't been done.

Perhaps `cargo +nightly -Zminimal-versions` in the CI?

kchibisov:
>Perhaps cargo +nightly -Zminimal-versions in the CI?

This is very tricky to get working actually, but I have plans to do that after decoupling, since we'll have very few deps.

MarijnS95:
The main problem is typically transitive dependencies that themselves don't set an appropriate minimum version... And not everyone wants to spend lots of CI time to (build)test many configurations.

PR #1440, opened 2022-08-12
Move to `objc2`

madsmtm:
Corresponding `winit` PR: https://github.com/rust-windowing/winit/pull/2427

- [ ] Tested on all platforms changed
  - [x] macOS 10.14
  - [ ] iOS simulator
  - [ ] iOS 9.3.6 (iPad mini 1st gen)

PR #1441, opened 2022-08-12
iOS: Fix UB in `Context::create_context`

madsmtm:
We were calling an `init` method on the same object multiple times, which is not allowed in Objective-C.

Found while making #1440. With this change I can now run `glutin` on my old iPad.


kchibisov:
There's no ios anymore.

PR #1439, opened 2022-08-10
Glutin version v0.29.1

kchibisov:

PR #1446, opened 2022-08-23
Use get_instance_handle() over GetModuleHandle(NULL)

### LINKED ISSUE #1445: ERROR_CLASS_DOES_NOT_EXIST opening window from DLL
Building a glutin context compiled as a dynamically loaded DLL causes:

```plaintext
OsError("GetClassInfoExW function failed: <localized message> (os error 1411)")
```

This error is caused by the `GetModuleHandleW(std::ptr::null())` usage in `glutin/src/api/wgl/mod.rs`, the function returns the handle for the executable, not the DLL.

A similar bug was fixed in winit, see [winit/2301](https://github.com/rust-windowing/winit/pull/2301)

SamRodri:
See the [winit](https://github.com/rust-windowing/winit/pull/2301) discussion about this fix.

fixes #1445

### DIFF
--- a/glutin/src/api/wgl/mod.rs
+++ b/glutin/src/api/wgl/mod.rs
@@ -16,6 +16,7 @@ use winapi::shared::ntdef::LPCWSTR;
 use winapi::shared::windef::{HDC, HGLRC, HWND};
 use winapi::um::libloaderapi::*;
 use winapi::um::wingdi::*;
+use winapi::um::winnt::IMAGE_DOS_HEADER;
 use winapi::um::winuser::*;
 
 use std::ffi::{CStr, CString, OsStr};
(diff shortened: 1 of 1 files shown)


SamRodri:
I see  that there is a major rewrite underway, maybe @kchibisov can copy this patch into his pull request?

kchibisov:
Actually, I don't think I need this patch at all on my branch. I just need to use `hinstance` from the handle instead of that.

I'm not sure how I feel about adding this patch right now given that I'm unlikely to make a glutin release.

Is there a strong need for this patch?

SamRodri:
No rush on my part, I can depend on my fork until your update is released.

kchibisov:
Handled in #1439.

PR #1451, opened 2022-09-03
Glutin version v0.30.0-beta.1

kchibisov:


glutin/Cargo.toml
```diff
@@ -1,11 +1,11 @@
 [package]
 name = "glutin"
-version = "0.29.1"
-authors = ["The glutin contributors", "Pierre Krieger <pierre.krieger1708@gmail.com>"]
+version = "0.30.0-beta.1"
+authors = ["Kirill Chibisov <contact@kchibisov.com>"]
```
kchibisov, line 4: Given that this is purely for maintenance and the crate got an entire rewrite I'll update this to the current active maintainers (which end up being me).

The `The glutin contributors` doesn't provide any value and there's no such entry any where as `glutin contributors` (We don't maintain CONTRIBUTORS.md with mail map or something).

madsmtm: The field is optional, so you could also just remove it: https://rust-lang.github.io/rfcs/3052-optional-authors-field.html

PR #1453, opened 2022-09-03
Fix macOS NSView release call

madsmtm:
Looked through message sends and stuff after #1435, looks good otherwise.

### DIFF
--- a/glutin/src/api/cgl/surface.rs
+++ b/glutin/src/api/cgl/surface.rs
@@ -71,7 +71,7 @@ impl<T: SurfaceTypeTrait> Drop for Surface<T> {
     fn drop(&mut self) {
         unsafe {
             if self.ns_view != nil {
-                let _: () = msg_send![self.ns_view, retain];
+                let _: () = msg_send![self.ns_view, release];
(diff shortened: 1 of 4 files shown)


glutin/src/api/cgl/surface.rs
```diff
@@ -71,7 +71,7 @@ impl<T: SurfaceTypeTrait> Drop for Surface<T> {
     fn drop(&mut self) {
         unsafe {
             if self.ns_view != nil {
-                let _: () = msg_send![self.ns_view, retain];
+                let _: () = msg_send![self.ns_view, release];
```
kchibisov, line 74: I think it should just decrease the count? We borrow the object from the winit, we don't need to release? Winit should drop it.

kchibisov (changes requested):

glutin/src/api/cgl/surface.rs
madsmtm: We call `retain` inside `create_window_surface`

kchibisov (approved):

PR #1459, opened 2022-09-06
context/config: don't force OpenGL api

kchibisov:
This should allow creating window on android without tweaking the default
for the Api.

--

@MarijnS95 This should solve your issue with `Gles` requests by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- Config doesn't force OpenGL Api by default.
+- `Config` doesn't force OpenGL `Api` by default.
 - `Display::create_context` now uses the most recent available `Api` from the `Config` when `ContextApi` is not specified in `ContextAttributes`.
 
 # Version 0.30.0-beta.2 (2022-09-03)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+- Config doesn't force OpenGL Api by default.
+- `Display::create_context` now uses the most recent available `Api` from the `Config` when `ContextApi` is not specified in `ContextAttributes`.
+
 # Version 0.30.0-beta.2 (2022-09-03)
 
 - macOS: Fix memory leak in `Surface`.
--- a/glutin/src/api/egl/config.rs
+++ b/glutin/src/api/egl/config.rs
@@ -122,7 +122,7 @@ impl Display {
         if let Some(requested_api) = template.api {
             let mut api = 0;
             if requested_api.contains(Api::GLES1) {
-                api |= egl::OPENGL_ES_API;
+                api |= egl::OPENGL_ES_BIT;
             }
             if requested_api.contains(Api::GLES2) {
                 api |= egl::OPENGL_ES2_BIT;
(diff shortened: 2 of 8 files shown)


glutin/src/api/egl/context.rs
```diff
@@ -33,8 +33,10 @@ impl Display {
         let supports_opengl = self.inner.version > Version::new(1, 3);
 
         let (api, version) = match context_attributes.api {
-            ContextApi::OpenGl(version) if supports_opengl => (egl::OPENGL_API, version),
-            ContextApi::Gles(version) => (egl::OPENGL_ES_API, version),
+            Some(ContextApi::OpenGl(version)) if supports_opengl => (egl::OPENGL_API, version),
+            Some(ContextApi::Gles(version)) => (egl::OPENGL_ES_API, version),
+            None if config.api().contains(Api::OPENGL) => (egl::OPENGL_API, None),
+            None => (egl::OPENGL_ES_BIT, None),
```
MarijnS95, line 39:
```suggestion
            None => (egl::OPENGL_ES_API
            , None),
```

glutin/src/api/egl/config.rs
```diff
@@ -119,21 +119,23 @@ impl Display {
...
-        }
-        if template.api.contains(Api::GLES3) {
-            api |= egl::OPENGL_ES3_BIT;
-        }
-        if template.api.contains(Api::OPENGL) {
-            api |= egl::OPENGL_BIT;
+        if let Some(requested_api) = template.api {
+            let mut api = 0;
+            if requested_api.contains(Api::GLES1) {
+                api |= egl::OPENGL_ES_API;
```
MarijnS95, line 125: BIT?

MarijnS95:
Seems okay, besides some BIT/API typos. If I understand correctly, this:

1. Doesn't set any API bit (when not explicitly requested) in the config attributes' `RENDERABLE_TYPE` when creating an EGL display connection, which allows all APIs to be used;
2. Subsequently the EGL display connection can be queried for the exact same attribute, now containing exactly the API(s) it supports (this doesn't seem to be a subset if `RENDERABLE_TYPE` was set explicitly in `ChooseConfig`);
3. Creating a GL context now looks at that list of _actually supported_ APIs to automatically figure out what kind of context to create, defaulting to `OPENGL` if both were supported.

kchibisov:
>creating an EGL display connection, which allows all APIs to be used

When querying the config, display can use multiple apis without any issue.

>Subsequently the EGL display connection can be queried for the exact same attribute, now containing exactly the API(s) it supports (this doesn't seem to be a subset if RENDERABLE_TYPE was set explicitly in ChooseConfig);

You can get the api config supports from it.

>Creating a GL context now looks at that list of actually supported APIs to automatically figure out what kind of context to create, defaulting to OPENGL if both were supported

Yes.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+- Config doesn't force OpenGL Api by default.
```
MarijnS95, line 3:
```suggestion
- `Config` doesn't force OpenGL `Api` by default.
```

MarijnS95 (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+- `Config`` doesn't force OpenGL `Api` by default.
```
MarijnS95, line 3:
```suggestion
- `Config` doesn't force OpenGL `Api` by default.
```

PR #1460, opened 2022-09-08
api: move `get_proc_address` to `GlDisplay`

kchibisov:
This should made it clear that you should call this only once and not on each context creation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - `Config` doesn't force OpenGL `Api` by default.
 - `Display::create_context` now uses the most recent available `Api` from the `Config` when `ContextApi` is not specified in `ContextAttributes`.
+- `PossiblyCurrentGlContext::get_proc_address` method was moved to `GlDisplay::get_proc_address`.
 
 # Version 0.30.0-beta.2 (2022-09-03)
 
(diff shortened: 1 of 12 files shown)


MarijnS95:
Thanks! That was a blocker for #1417 to function in a clean way :)

kchibisov:
I've added a comment wrt that, since having function on the context is missleading in the first place, because you'll try to load it each time you call the function.

MarijnS95:
Yeah you mentioned in #1417 but it's probably worth mentioning it explicitly in the PR here too to investigate.

Can you add an optional current context argument to `get_proc_addr()` and assert it's not-None on `WGL`?

kchibisov:
>Can you add an optional current context argument to get_proc_addr() and assert it's not-None on WGL?

The thing is that it'll still load, it will be just a limited set of functions. You won't have shaders, but you'll have `glClear` for example. It's not like it won't work at all, that's the point. I don't want to have a custom argument here, since you can't crash due to that, you'll just load not everything and that's sort of clear that way anyway.

I'll see whether it'll raise issues and adjust if folks will struggle with that.

MarijnS95 (approved):

MarijnS95:
> The thing is that it'll still load, it will be just a limited set of functions.

I did not know that.

Then indeed, I agree that we shouldn't crash, documenting this properly is all we can do. Let's see if any issues arise.

PR #1463, opened 2022-09-09
fix(api): multisampling requests

### LINKED ISSUE #1462: How do i use mutisampiling on glutin >= 0.30?

kchibisov:
We were asking for sample buffers, but not samples per buffer, which is what required for AA and used.

Fixes #1462.


MarijnS95 (approved):

ar37-rs:
I think https://github.com/brendanzab/gl-rs/blob/master/gl/examples/triangle.rs is a good candidate for the example, if the triangle looking good smooth, it means it's working.

kchibisov:
I mean that you can clearly know that it picked or not by getting `num_samples` from the config.

kchibisov:
@ar37-rs made example more complex and now it picks the config with the maximum sample buffers here #1465.

PR #1465, opened 2022-09-09
example: make example more complex

kchibisov:
This should make use of shaders and pick config
with the maximum sample buffers.

Besides that, this commit also fixes the core
profile picking, since you must specify the
version higher than 3.3.

### DIFF
--- a/glutin/src/context.rs
+++ b/glutin/src/context.rs
@@ -176,7 +176,7 @@ impl ContextAttributesBuilder {
     ///
     /// # Api-specific
     ///
-    /// **macOS:** - not supported.
+    /// **macOS:** - not supported, the latest is picked automatically.
     ///
     /// [`GlProfile`]: crate::context::GlProfile
     pub fn with_profile(mut self, profile: GlProfile) -> Self {
@@ -186,7 +186,7 @@ impl ContextAttributesBuilder {
 
     /// Set the desired OpenGL context api. See the docs of [`ContextApi`].
     ///
-    /// By default the lastest supported by the config is picked.
+    /// By default the supported api will be picked.
     ///
     /// [`ContextApi`]: crate::context::ContextApi
     pub fn with_context_api(mut self, api: ContextApi) -> Self {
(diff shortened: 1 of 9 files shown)


ar37-rs (approved):

glutin_examples/examples/support/mod.rs
```diff
@@ -44,12 +45,15 @@ impl GlWindow {
...
-    ConfigTemplateBuilder::new()
+    let builder = ConfigTemplateBuilder::new()
         .with_alpha_size(8)
-        .with_transparency(true)
         .compatible_with_native_window(raw_window_handle)
-        .with_surface_type(ConfigSurfaceTypes::WINDOW)
-        .build()
+        .with_surface_type(ConfigSurfaceTypes::WINDOW);
+
+    #[cfg(cgl_backend)]
```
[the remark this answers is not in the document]
kchibisov: I don't think that WGL needs that. It should work fine without that?

madsmtm (commented):
Cool, this example helps an OpenGL newbie like me a lot in understanding `glutin`'s role

glutin/src/context.rs
```diff
@@ -280,7 +282,9 @@ pub enum GlProfile {
 pub enum ContextApi {
     /// OpenGL Api version that should be used by the context.
     ///
-    /// When using `None` as a `ApiVersion` any OpenGl context will be picked.
+    /// When using `None` as a `ApiVersion` any OpenGl context will be picked,
```
MarijnS95, line 285:
```suggestion
    /// When using `None` as `ApiVersion` any OpenGl context will be picked,
```

Same below on the `Gles` item.

MarijnS95: Also, `OpenGl` or `OpenGL`?

MarijnS95 (approved):
Runs on Linux desktop with EGL, haven't really dug further than that.

Will submit a change to generalize this and subsequently make it work on Android (#1417) after this is merge.

kchibisov:
I've decided to use `StructGenerator` and also added more `null` checking.

PR #1472, opened 2022-09-21
Remove intradoc "link label" indirection for items that are already in scope

MarijnS95:
- [ ] Ran `cargo doc --workspace --all-features --no-deps --document-private-items` and observed no broken link errors

### DIFF
--- a/glutin/src/context.rs
+++ b/glutin/src/context.rs
@@ -146,13 +146,9 @@ impl ContextAttributesBuilder {
     /// Sets the robustness of the OpenGL context. See the docs of
     /// [`Robustness`].
     ///
-    /// The default is [`NotRobust`], because this is what typically expected
-    /// when you create an OpenGL context.  However for safety you should
-    /// consider [`RobustLoseContextOnReset`].
-    ///
-    /// [`Robustness`]: crate::context::Robustness
-    /// [`NotRobust`]: crate::context::Robustness::NotRobust
-    /// [`RobustLoseContextOnReset`]: crate::context::Robustness::RobustLoseContextOnReset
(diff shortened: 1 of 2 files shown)


glutin/src/context.rs
```diff
@@ -466,7 +453,7 @@ impl Sealed for NotCurrentContext {}
 ///
 /// The context that could be current on the current thread can neither be
 /// [`Send`] nor [`Sync`]. In case you need to use it on a different thread
-/// [`make it not current`].
+/// [make it not current].
```
MarijnS95, line 456: This is just regular text, doesn't need to be an inline code span.

glutin/src/context.rs
```diff
@@ -146,13 +146,9 @@ impl ContextAttributesBuilder {
...
     /// [`Robustness`].
     ///
-    /// The default is [`NotRobust`], because this is what typically expected
-    /// when you create an OpenGL context.  However for safety you should
-    /// consider [`RobustLoseContextOnReset`].
-    ///
-    /// [`Robustness`]: crate::context::Robustness
-    /// [`NotRobust`]: crate::context::Robustness::NotRobust
-    /// [`RobustLoseContextOnReset`]: crate::context::Robustness::RobustLoseContextOnReset
+    /// The default is [`Robustness::NotRobust`], because this is what typically
```
MarijnS95, line 149: I can revert these again since `NotRobust` isn't an item in scope but an _enum variant_. We can still reference to it using `` [`NotRobust`]: Robustness::NotRobust `` etc, let me know!

kchibisov (approved):

PR #1461, opened 2022-09-09
Move to `objc2`

madsmtm:
Helps with following Cocoa's memory management rules (for example, https://github.com/rust-windowing/glutin/pull/1453 wouldn't have happened with this) - and while I was at it, I fixed a few C `enum` definitions (`objc2`'s `verify_message` feature is a great help with that).

See commit messages for details.

- [x] Tested on all platforms changed
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [x] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [x] Created or updated an example program if it would help users understand this functionality

PR #1474, opened 2022-09-21
refactor: use NO constant on macos

keiya01:
I fixed error that `is_equal != 0` is not matched type on macos.


madsmtm:
Fixed also in https://github.com/rust-windowing/glutin/pull/1461

PR #1475, opened 2022-09-22
Fix example crashing on linux with LLVMpipe Driver

ar37-rs:
* to fix crashing on linux with LLVMpipe Driver (and possibly other ancient drivers) turn off window transparency by default.

RUST_LOG:
```rust
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { raw_code: Some(8), raw_os_message: Some("BadMatch (invalid parameter attributes)"), kind: BadMatch }', glutin_examples/src/lib.rs:135:61
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### DIFF
--- a/glutin_examples/src/lib.rs
+++ b/glutin_examples/src/lib.rs
@@ -31,7 +31,7 @@ pub fn main() {
     // We create a window before the display to accomodate for WGL, since it
     // requires creating HDC for properly loading the WGL and it should be taken
     // from the window you'll be rendering into.
-    let window = WindowBuilder::new().with_transparent(true).build(&event_loop).unwrap();
+    let window = WindowBuilder::new().build(&event_loop).unwrap();
     let raw_window_handle = window.raw_window_handle();
 
     // Create the GL display. This will create display automatically for the
@@ -121,6 +121,7 @@ pub fn main() {
                             NonZeroU32::new(size.height).unwrap(),
                         );
                         renderer.resize(size.width as i32, size.height as i32);
+                        gl_window.window.request_redraw();
                     }
                 },
                 WindowEvent::CloseRequested => {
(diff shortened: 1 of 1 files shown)


ar37-rs:
Hi @MarijnS95, i have trouble running example on my Ubuntu VM.

ar37-rs:
i meant https://sourceforge.net/projects/vcxsrv + WSL2, i'm having trouble running the example on that.

kchibisov:
I'm aware of that issue and I'm not sure how to properly handle it in a forward compatible way.

Basically the situation is that we pick some visual for the window with alpha mask and try to use it, however EGL on X11 isn't capable of doing so. The way forward here is to pick visual first and then build a window, but on Windows you have similar situation, but you need window first, and with GLX you can use visual from the winit to perform the match.

I'd thought about that issue before and yet I don't have a good solution for that. If we only we didn't have Windows WGL in this equation it would be much simpler, but we're out of luck...

The most correct way is to pass the visual right to winit when building it, so it can judge wrt transparency based on that. Would probably have ugly startup code for windows, but who cares at this point...

If someone has an idea how to solve the issue:

1. On windows we need HDC from the window to query pixel formats and to load ICD.
2. On X11 we need to pass visual from the config to `WindowBuilder`.

So the building pipeline would be on non windows:

1. Create Display
2. Query configs
3. Pass visual to window and build it.

And on windows the order is 3, 1, 2, which is a problem, but I doubt there's a solution unless we glue all types together, which is terribly broken.

I guess we'd have to leave with this fragmentation, since you can use windows without WGL and it'll work just fine.

ar37-rs:
@kchibisov agree on that. thing can be tricky, but if someone has an idea, a problem can be solved.

kchibisov:
I mean it's not really a problem given that it can't be solved on the API level, it's just requires different order in example on WGL, nothing stops from using the order that fits your case the best.

kchibisov:
The recent master should have a proper fix for that.

PR #1481, opened 2022-10-07
cgl: fix `Eq` implementation

kchibisov:
We should compare `BOOL` with `NO` and not with `0`.


madsmtm:
Looks good.

For reference, originally in https://github.com/rust-windowing/glutin/pull/1474

kchibisov:
Yeah, I'd use that PR instead but it was deleted.

PR #1482, opened 2022-10-11
README: switch link to new Matrix room

Be-ing:
- [N/A] Tested on all platforms changed
- [N/A] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [N/A] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [N/A] Created or updated an example program if it would help users understand this functionality

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ glutin = "0.30.0-beta.2"
 
 Join us in any of these:
 
-[![Matrix](https://img.shields.io/badge/Matrix-%23rust--windowing%3Amatrix.org-blueviolet.svg)](https://matrix.to/#/#rust-windowing:matrix.org)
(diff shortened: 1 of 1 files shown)


kchibisov:
I'm not entirely sure wrt that. The thing is that the current link is also correct, it's a link to the room where you can go to `#winit`. Before we also had `#glutin`. Though maybe it's fine to point to `#winit` directly.

i509VCB:
Maybe linking to the rust-windowing matrix space would be more apt?

kchibisov:
I'm not familiar with any of that tbf, my link is the one with `irc` in it :p . But whatever works better for matrix folks here.

Be-ing:
The rust-windowing Matrix room's message says it is abandoned and redirects to the winit room.

kchibisov:
I mean that I've clicked on the link in element and it showed glutin room(archived) and winit room.

i509VCB:
I believe the space link is https://matrix.to/#/#rust-windowing:matrix.org

kchibisov (approved):

PR #1502, opened 2022-10-21
chore: enable doc_auto_cfg

### LINKED ISSUE #1498: Set doc_auto_cfg when docs.rs generates docs
This should attach feature labels to every type and function for which those exist.

kchibisov:
Fixes #1498.

### DIFF
--- a/glutin/Cargo.toml
+++ b/glutin/Cargo.toml
@@ -57,3 +57,14 @@ objc = "0.2.7"
 
 [build-dependencies]
 cfg_aliases = "0.1.1"
+
+[package.metadata.docs.rs]
+rustdoc-args = ["--cfg", "docsrs"]
+targets = [
+    "aarch64-linux-android",
+    "x86_64-unknown-linux-gnu",
+    "x86_64-pc-windows-msvc",
+    "x86_64-apple-darwin",
+    "i686-unknown-linux-gnu",
+    "i686-pc-windows-msvc",
+]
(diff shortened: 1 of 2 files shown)


kchibisov:
Do you happen to know if that's all we need to make it work?

MarijnS95:
> Do you happen to know if that's all we need to make it work?

@kchibisov You also need to enable the `docsrs` feature when building on/for https://docs.rs, see for example: https://github.com/rust-windowing/android-ndk-rs/pull/293/files

(That's the way to detect `docs.rs` - and a simple `cfg(doc)` would break anyone running `cargo doc` on the stable toolchain)

kchibisov:
Could I leave metadata wrt targets blank?

MarijnS95:
@kchibisov The targets instruct what target(s) to build the docs for, see the "Platform" selector on the top at https://docs.rs/glutin/latest/glutin/.

kchibisov:
@MarijnS95 seems like it should be fine now?

MarijnS95 (approved):
Yup, lovely! Up to you to decide what `targets` to build, I don't think we have much if any architecture-dependent docs (like we would have for the NDK and its related crates, obviously).

glutin/Cargo.toml
```diff
@@ -57,3 +57,14 @@ objc = "0.2.7"
...
+[package.metadata.docs.rs]
+rustdoc-args = ["--cfg", "docsrs"]
+targets = [
+    "aarch64-linux-android",
+    "x86_64-unknown-linux-gnu",
+    "x86_64-pc-windows-msvc",
+    "x86_64-apple-darwin",
+    "i686-unknown-linux-gnu",
+    "i686-pc-windows-msvc",
+]
```
MarijnS95, line 70: GitHub complains about a missing trailing newline :stuck_out_tongue:

PR #1506, opened 2022-10-25
Glutin version 0.30.0

kchibisov:
--
cc @i509VCB 
cc @chrisduerr 

If anyone has any concerns let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,42 +1,15 @@
 # Unreleased
 
-# Version 0.30.0-beta.3
+# Version 0.30.0
 
-- `Config` doesn't force OpenGL `Api` by default.
-- `Display::create_context` now uses the most recent available `Api` from the `Config` when `ContextApi` is not specified in `ContextAttributes`.
-- **Breaking:** `PossiblyCurrentGlContext::get_proc_address` method was moved to `GlDisplay::get_proc_address`.
-- **Breaking:** `ConfigTemplateBuilder::with_sample_buffers` now called `ConfigTemplateBuilder::with_multisampling`.
-- **Breaking:** `GlConfig::sample_buffers` now called `GlConfig::num_samples` and returns the amount of samples in multisample buffer.
+- **This version of `glutin` has been rewritten from the ground and no longer depends on `winit`, the `raw-window-handle` is now used instead of it.**
+- The Api is now built around `Display`, `Surface`, `Config`, and `Surface`. For more info see crate documentation and examples.
 - **Breaking:** Bump MSRV from `1.57` to `1.60`.
-- Fix `GlProfile::Core` requesting without explicit version.
-- Pick the latest available profile on macOS.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,42 +1,15 @@
...
-- Fix crash in `glx::surface::Surface::set_swap_interval`.
-
-# Version 0.30.0-beta.2 (2022-09-03)
-
-- macOS: Fix memory leak in `Surface`.
-
-# Version 0.30.0-beta.1 (2022-09-03)
-
 - Replace `winit` dependency with `raw-window-handle`.
 - The Api is now built around `Display`, `Surface`, `Config`, and `Surface` for more info see crate documentation and examples.
```
MarijnS95, line 7:
```suggestion
- The Api is now built around `Display`, `Surface`, `Config`, and `Surface`. For more info see crate documentation and examples.
```

MarijnS95 (approved):
Lovely!

Would the `winit` removal warrant a `1.0.0` release, finally? Or do you want to let it boil for a little while longer? Does this **entire codebase refactor** need a bit more front-and-center attention in the changelog?

kchibisov:
>Would the winit removal warrant a 1.0.0 release, finally? Or do you want to let it boil for a little while longer? Does this entire codebase and API refactor need a bit more front-and-center attention in the changelog?

I think this needs more battle testing, but yeah, `glutin` is not far from potential `v1.0.0` release. Though I'd wait for `raw-window-handle v1.0.0` which should be soon as well.

It would also be nice to see major consumers moving to `glutin 0.30`, like `egui`, `iced`, etc.

There're some api changes that we still need to make, but most of them are non-breaking. For example it would be nice to expose loaded libraries is some form and ability to create display from `raw` pointer. Like from raw `EGLDisplay`. I also need `GAT`s for some stuff internally, which should be soon, so I can remove redundant internal traits with generics...

It's also not clear what to do with `wasm` and `ios` support right now. It seems like both are possible to do, just need more attention, maybe with a leaner codebase there will be some attention to that problem and we have folks that care about it. Only time will tell. But having support for most platforms is already a good start.

>Does this entire codebase and API refactor need a bit more front-and-center attention in the changelog

I'm not sure how to put it in the CHANGLELOG, but I write something for `Release page` for sure. The thing is that we have entirely new crate in the end that has nothing to do with the old glutin other than name. So the CHANGELOG won't help here.

I do think that it would be nice to have more examples wrt complex handling of things, like multiwindow/multitheared stuff, context reset, but I don't have much time to write such examples myself (I ported alacritty with all of that stuff except multithreading, but **winit** on macOS is hard here, since I'm not sure how safe it is).

MarijnS95:
I think you can just put it as the top-most, maybe even bold changelog point (`- This version of Glutin has been rewritten from the ground up blah blah and the new API revolves around [the four new types you mention later]`).

kchibisov:
I guess it's not that a bad idea.

PR #1512, opened 2022-10-28
set default target for docs.rs

i509VCB:
If the default target is not set and a list of targets is specifed, the first target is selected. Likely not too useful to have android be the default target on docs.rs

- [X] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [X] Created or updated an example program if it would help users understand this functionality

### DIFF
--- a/glutin/Cargo.toml
+++ b/glutin/Cargo.toml
@@ -68,3 +68,4 @@ targets = [
     "i686-unknown-linux-gnu",
     "i686-pc-windows-msvc",
 ]
+default-target = "x86_64-unknown-linux-gnu"


MarijnS95:
Just swap the target list?

In any case, I asked previously if compiling docs for all these individual targets is necessary?

i509VCB:
Probably best to be explicit here.

kchibisov:
How does that work when I access `docs.rs` from macOS? Shouldn't it figure out that I should use `darwin` target? I've expected that `docs.rs` resolves to a platform you're on by default?

MarijnS95:
Seems like it doesn't - the default docs don't indicate it but only contain `egl` docs.

Shouldn't we select a platform as default that supports the most backends? Or if we wish to spend more complexity, use `cfg(docsrs)` to enable _all_ API just so that the top-level docs include everything? Some other projects to it, but it might be a hassle in certain places where things may not compile for the given target.

kchibisov:
Hm, the platform with the most features is the linux then. I'm fine with explicitly setting the target like it was proposed.

kchibisov (approved):

PR #1515, opened 2022-10-29
fix: call to GetDC in swap_buffers

baehny:
- fix the error that frequent calls to GetDC slows down swap_buffers over time
- moved the call of GetDC to the surface creation and cached the HDC in WGL's GlSurface

Frequent calls to GetDC slows down swap_buffers over time. The issue is also described in https://stackoverflow.com/questions/37347939/swapbuffers-getting-slow-over-time.

The change fixes the issue, but I'm not quite sure if it is problematic that the aquired handle is not release with a following
call to ReleaseDC (see https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getdc).

If required I can provide an example to reproduce the issue.

Frametimes before the change:

grafik

Frametimes after the change:

grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- On Windows (WGL), fixed that `Surface::swap_buffers` takes longer with every call caused by frequent calls of the win32 function `HDC GetDC(HWND hWnd)`.
+- On WGL, fixed that `Surface::swap_buffers` takes longer with every call caused by frequent calls of the win32 function `HDC GetDC(HWND hWnd)`.
 
 # Version 0.30.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- On Windows (WGL), fixed that `Surface::swap_buffers` takes longer with every call caused by frequent calls of the win32 function `HDC GetDC(HWND hWnd)`.
+
 # Version 0.30.0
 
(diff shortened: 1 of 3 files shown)


kchibisov:
Thanks for the patch. I'm not using Windows for a long time so I'd expect missing something like that. 

>The change fixes the issue, but I'm not quite sure if it is problematic that the aquired handle is not release with a following call to ReleaseDC

We can release it when dropping the surface, since they will reside on the same thread. I'm not sure if it could cause issues with multiple windows.

kchibisov (changes requested):
Would need a CHANGELOG entry about that.

baehny:
I have added the CHANGELOG entry, added the missing call to ReleaseDC and also removed another call to GetDC in make_current.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- On Windows (WGL), fixed that `Surface::swap_buffers` takes longer with every call caused by frequent calls of the win32 function `HDC GetDC(HWND hWnd)`.
```
kchibisov, line 3:
```suggestion
- On WGL, fixed that `Surface::swap_buffers` takes longer with every call caused by frequent calls of the win32 function `HDC GetDC(HWND hWnd)`.
```

kchibisov (changes requested):

kchibisov (approved):

kchibisov:
Thanks!

PR #1517, opened 2022-10-30
api: add glutin-winit crate

### LINKED ISSUE #1516: How to migrate from `0.29`?
I'm taking a look a migrating my various gl examples, crates & my game to the latest version. However, looking at the example seems to be the only way to understand the changes and I'm struggling to see how I should migrate. Previously I would use [a relatively small amount of declarative code](https://github.com/alexheretic/glyph-brush/blob/27b9fef02b8805f225d6fd34fadfb272d95e07f7/glyph-brush/examples/opengl.rs#L62..L72) to provide me with what I needed, a gl window I can draw into. The only other non-winit bit then was the `swap_buffers` call.

The new API now requires me to know about display-apis, tbh I don't particularly want to. Isn't there a "best" choice for each platform that the previous versions were using? The example also uses explicit [non-trivial conditional compilation to choose the display api](https://github.com/rust-windowing/glutin/blob/2a2a97209c49929027beced68e1989b8486bdec9/glutin_examples/src/lib.rs#L273). Do you expect all _glutin_ users to re-implement this conditional compilation? Doing that would be quite verbose, I'd also have to unpick the features used in the example since display-api features don't really make sense downstream. It seems like it would end up a lot more code, which would be undesirable particularly in glyph-brush examples etc.

I'm also unsure exactly how the old code made the decisions I now need to satisfy (without going through the `0.29` code). It seems quite tricky for me to migrate my game without introducing changes in behaviour on one or more of the platforms I support (linux, windows, mac). It would be very useful to have "before and after" code examples, e.g. old `ContextBuilder` usage is now...

Overall I'm not sure I agree that _glutin_ should have changed into a lower level crate. It doesn't seem to be able to do the job it could do for me before. Perhaps it could have been a new _glutin-core_ crate or something and _glutin_ could keep working a bit more like it used to? But I guess the ship has sailed there. On the other hand perhaps I've read the changes incorrectly, or my usage is more niche than I thought. I'd appreciate advice on how I should (or if I shouldn't?) migrate.

kchibisov:
This crate should provide a cross-platform bootstrapping of glutin with the winit crate to reduce boilderplate when using with winit.

Fixes #1516.

cc the ones who might be interested in such a thing.

@chrisduerr 
@alexheretic 
@ids1024

### DIFF
--- a/glutin-winit/src/lib.rs
+++ b/glutin-winit/src/lib.rs
@@ -1,0 +1,220 @@
+//! This library provides helpers for cross-platform [`glutin`] bootstrapping
+//! with [`winit`].
+
+#![deny(rust_2018_idioms)]
+#![deny(rustdoc::broken_intra_doc_links)]
+#![deny(clippy::all)]
+#![deny(missing_debug_implementations)]
+#![deny(missing_docs)]
+#![cfg_attr(feature = "cargo-clippy", deny(warnings))]
+
+use std::error::Error;
+
+use glutin::config::{Config, ConfigTemplateBuilder};
+use glutin::display::{Display, DisplayApiPreference};
+#[cfg(x11_platform)]
+use glutin::platform::x11::X11GlConfigExt;
+use glutin::prelude::*;
+
+use raw_window_handle::{HasRawDisplayHandle, RawWindowHandle};
+
+#[cfg(wgl_backend)]
+use raw_window_handle::HasRawWindowHandle;
+
+use winit::error::OsError;
+use winit::event_loop::EventLoopWindowTarget;
+use winit::window::{Window, WindowBuilder};
+
+#[cfg(glx_backend)]
+use winit::platform::unix;
+#[cfg(x11_platform)]
+use winit::platform::unix::WindowBuilderExtUnix;
+
+#[cfg(all(not(egl_backend), not(glx_backend), not(wgl_backend), not(cgl_backend)))]
+compile_error!("Please select at least one api backend");
+
+/// The helper to perform [`Display`] creation and OpenGL platform
+/// bootstrapping with the help of [`winit`] with little to no platform specific
+/// code.
+///
+/// This is only required for the initial setup. If you want to create
+/// additional windows just use the [`finalize_window`] function and the
+/// configuration you've used either for the original window or picked with the
+/// existing [`Display`].
+///
+/// [`winit`]: winit
+/// [`Display`]: glutin::display::Display
+#[derive(Default, Debug, Clone)]
+pub struct DisplayBuilder {
+    preference: ApiPrefence,
+    window_builder: Option<WindowBuilder>,
+}
+
+impl DisplayBuilder {
+    /// Create new display builder.
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// The preference in picking the configuration.
+    pub fn with_prefence(mut self, preference: ApiPrefence) -> Self {
+        self.preference = preference;
+        self
+    }
+
+    /// The window builder to use when building a window.
+    ///
+    /// By default no window is created.
+    pub fn with_window_builder(mut self, window_builder: Option<WindowBuilder>) -> Self {
+        self.window_builder = window_builder;
+        self
+    }
+
+    /// Initialize the OpenGL platform and create a compatible window to use
+    /// with it when the [`WindowBuilder`] was passed with
+    /// [`Self::with_window_builder`]. It's optional, since on some
+    /// platforms like `Android` it is not available early on, so you want to
+    /// find configuration and later use it with the [`finalize_window`].
+    /// But if you don't care about such platform you can always pass
+    /// [`WindowBuilder`].
+    ///
+    /// # Api-specific
+    ///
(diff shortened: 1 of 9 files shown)


glutin-winit/src/lib.rs
```diff
@@ -0,0 +1,219 @@
+//! This library provides helpers for cross-platform glutin bootstrapping with
+//! winit.
+
+#![deny(rust_2018_idioms)]
```
notgull, line 4: Should `forbid` be used with these lints to prevent them from being overridden by modules later on?

Also, I think that you can group all of these `deny`s into one block to reduce noise.

glutin-winit/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+        mut self,
+        window_target: &EventLoopWindowTarget<T>,
+        template_builder: ConfigTemplateBuilder,
+        config_picker: Picker,
+    ) -> Result<(Option<Window>, Config), Box<dyn Error>>
+    where
+        Picker: FnOnce(Box<dyn Iterator<Item = Config> + '_>) -> Config,
+    {
+        // XXX with WGL backend window should be created first.
+        #[cfg(wgl_backend)]
```
notgull, line 96: I think that the `wgl_backend` parts and the `not(wgl_backend)` parts could be split into separate functions here.

glutin-winit/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+        Ok((window, gl_config))
+    }
+}
+
+fn create_display<T>(
+    window_target: &EventLoopWindowTarget<T>,
+    _api_preference: ApiPrefence,
+    _raw_window_handle: Option<RawWindowHandle>,
+) -> Result<Display, Box<dyn Error>> {
+    #[cfg(egl_backend)]
```
notgull, line 141: I think that `cfg_if` would be more idiomatic here.

kchibisov: won't it be way more code? Those elements have no cost of being crated I'm not sure I can do that with `cfg_if`, since it'll really becomes a madness pretty fast. Be aware that most of these enums are not available. If you have an idea how to write it with `cfg_if` let me know.

glutin-winit/src/lib.rs
kchibisov: I'm not sure about what noise are you talking about here. I don't want to have 100 chars long `deny` strings. The way they here is the way I like to write them and it's easier for me to see them like that.

glutin-winit/src/lib.rs
kchibisov: Be aware that it'll be more messy, since I'd have to:

1. Add 2 more functions since I want to have shared docs without duplicating them.
2. They will be mostly the same? 

I just don't see a benefit in doing so, the function is tiny on its own.

kchibisov:
Do you feel like we should have a special treatment for macOS here since it can get only one config at a time? Like multiple template builders to try or another `Optional` config to try and re-pick the configuration? What do you think?

alexheretic:
I gave the code a try and could get it to work for one of my examples: https://github.com/alexheretic/glyph-brush/pull/159

`glutin_winit::DisplayBuilder` definitely helps! I think _glutin-winit_ could go further and help with wrapping the surface & context too a bit like old glutin did. Ideally the code could be as declarative as before, but with the option to dive deeper for the customisation I assume people want.

kchibisov:
>I think glutin-winit could go further and help with wrapping the surface & context too a bit like old glutin did.

That's not desired for android. The surface gets destroyed from time to time, but the context could live for the entire application.  The context could die as well on some platform (ContextLost), but the surface will continue to live leading to recreating stuff.

The design you want just doesn't work. It's not like the code isn't declarative, you just have 2 more types to keep around, context and surface and both are here for a good reason.

alexheretic:
> That's not desired for android. The surface gets destroyed from time to time, but the context could live for the entire application. The context could die as well on some platform (ContextLost), but the surface will continue to live, so you can rebuild the context and process like nothing happened.

While this may have been a flaw in the previous api, none of these concerns are app specific. It was winit/glutin's job to solve these kinds of things. I feel like there could probably still be an API that wraps these concerns, with a `rebuild_context` method or whatever (which by the sounds of it you could ignore outside android?).

Presumably the objective of _glutin-winit_ is to provide the best common code for everyone using both of those things, which is what _glutin_ tried to do previously. Right now old _glutin_ is still doing a better job for me. The work here helps, but imo doesn't go far enough.

kchibisov:
No, you can lose context on desktop GL as well. That's a common thing, like when you unplug your GPU or you switch GPU in-between.

>The work here helps, but I wonder if it could go further.

The only thing you have right now is that you have separate Surface and separate context. And how would you handle with such wrapping multiple surfaces being used by the same context? What if you have different read and draw surfaces? Old glutin wasn't thinking of all of that things and added unnecessary limitations here.

Be aware that Context and Surface are separate concepts, they don't tied together, there's no surface associated with context directly. If you feel like such wrappers are required I'm not against having them, I just feel like they are mostly useless? I know that old glutin works better for you since you have to write 10 lines less code, but you're mostly talking about toy examples where you have context and surface tied together.

PR #1527, opened 2022-11-16
Fix build on windows (examples)

cdecompilador:

### Why?
On the sub-crate `glutin-winit` it was used a symlink to share its build script with glutin, but symlinks on windows git are problematic, not enabled by default and only easily available on window 10+, so to avoid platform specific behaviour I just suggest using the cargo manifest attribute to specify the path of the build script outside its base folder.

### DIFF
--- a/glutin-winit/Cargo.toml
+++ b/glutin-winit/Cargo.toml
@@ -9,6 +9,9 @@ readme = "README.md"
 repository = "https://github.com/rust-windowing/glutin"
 edition = "2021"
 
+# Custom path for the build script
+build = "../glutin/build.rs"
+
 [features]
 default = ["egl", "glx", "x11", "wayland", "wgl"]
 egl = ["glutin/egl"]
--- a/glutin-winit/build.rs
+++ b/glutin-winit/build.rs
@@ -1,1 +1,0 @@
-../glutin/build.rs


cdecompilador:
I've tried doing a mock app, consuming as git repo and it worked, since crates works under the hood similar, and glutin/glutin-winit are sent together on the same workspace it should work the same as before.

Symlinks are common on linux specific git repositories, windows now supports git symlinks, but with some previous requirements:
1) Windows has to had developer mode enabled (only on Windows 10+) or launch every cargo command on an admin priv console
2) You have to enable the symlinks manually with a config and you can only do so if your git is updated

But you are right in that is clearer with symlinks, In case of merge the build.rs should be mentioned so users learning from the examples don't get confused.

Another intermediate option which is clear and allows unix/windows users to compile out of the box is maintaining the symlink but also the cargo manifest entry pointing to where the original file is, since symlinks on "not setup"/old windows are just files with a path inside

madsmtm:
> Though, I'd guess specifying `build.rs` path like that will work with crates.io?

Nope, if you try to do `cargo package --package glutin-winit`, it won't copy the build script to the package, and inspecting the generated package folder you will see that the `build` entry in the generated `Cargo.toml` is not changed.

cdecompilador:
that's odd, and should be a `cargo package` bug in my opinion, I'll research if its really it and for the moment the best change that should be made is a entry on the `glutin_examples/README.md` to tell windows users what to do to to run them depending if they are on new or older versions

cdecompilador:
If the `build = "../glutin"` would need to be accepted, we depend on this: https://github.com/rust-lang/cargo/pull/11389 so we should draft

If just an entry on `glutin_examples/README.md` advising windows users on how to make it work on their pc is selected, we can continue.

kchibisov:
Nice that we've discovered that cargo bug. For now I'm not quite sure how to handle that, maybe it's wise to split glutin and glutin-winit build scripts just for the sake of splitting since glutin might be ahead of the glutin-winit? Though those crates should be in-sync.

kchibisov:
I've removed the symlink for now.

MarijnS95:
@kchibisov you pduplicated the file](https://github.com/rust-windowing/glutin/commit/65d5e33ec12c6775d48690d4c18e65cd31725450) with an `// XXX keep in sync with glutin's build.rs.`? I find this PR much cleaner by pointing `cargo` to the right file... unless it somehow doesn't get picked up when packaging (test with `cargo package -l`)?

kchibisov:
This PR doesn't work unless the cargo fixes the bug. So for now I'd just duplicate. See the rust bug mentioned here.

MarijnS95:
@kchibisov Ack, quickly scrolling the PR lead me to believe that the bug was instead around `cargo` not being able to checkout / create the symlink somehow; yeah that bug is unfortunate but could be circumvented using `include = []`? Afaik there are more filename fields in `[package]` that don't get copied to the package by default...

kchibisov:
I don't think that include will work? Given that I'd need a correct `build.rs` path, which will be pointing in a wrong location? I'd keep it that way for now. it's really unlikely to change anyway unless we have more clear solution...

MarijnS95:
@kchibisov I was thinking of using `build = "build.rs"` plus `include = ["../glutin/build.rs", ..]` plus the symlink.

kchibisov:
But the issue is that symlink breaks on windows? Not sure how it'll help.

MarijnS95:
`include` would copy the original file, not the symlink... But I don't think it supports files outside of the package root directory (since the relative path that is provided is also where the file ends up in the final package).

PR #1531, opened 2022-11-26
ci: fix android job

kchibisov:


MarijnS95:
Fwiw I'm trying to back out of this change:

https://github.com/dvc94ch/cargo-subcommand/pull/25
https://github.com/rust-windowing/android-ndk-rs/issues/371

The change to `cargo doc` was just cleanup; a docs build doesn't need access to the Android compiler nor linker.

kchibisov:
@MarijnS95 could you look into the recent android failure and update CI?

MarijnS95:
@kchibisov The problem is that `-p glutin` is passed twice. We don't deduplicate that and `cargo-apk` doesn't want to build two packages at once.

Is that something you can fix in yml?

MarijnS95:
Should be fixed in https://github.com/rust-windowing/glutin/pull/1538.

kchibisov:
@MarijnS95 yeah, it just recently broke after the fix for the CI. So it's kind of strange if you haven't changed anything.

MarijnS95:
I did do a release recently, to fix some argument parsing breakage where were actually loosing the value of the first `-p` argument for complicated reasons...

Really, trying to reinvent cargo, its argument parsing and workspace detection is such a stupid and unreasonable hassle.

PR #1537, opened 2022-12-12
Fix initialization of EGL display from xcb display handle

tronical:
* https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_platform_xcb.txt suggests that the value for EGL_PLATFORM_XCB_SCREEN_EXT is 0x31DE
* Pass EGL_PLATFORM_XCB_SCREEN_EXT as screen selection attribute instead of the platform parameter PLATFORM_XCB_EXT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fixed EGL display initialization with XcbDisplayHandle.
+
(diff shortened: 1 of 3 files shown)


kchibisov:
Could you rebase? Should be good to go after that.

kchibisov (approved):

PR #1545, opened 2022-12-20
Add glutin-winit GlWindow trait helper

alexheretic:
I thought I'd experiment with reducing the noise using glutin-winit vs glutin `0.29`. This pr adds `glutin_winit::GlWindow` trait with extensions for winit `Window`s. (It could alternatively be called `glutin_winit::WindowExt` if you prefer).

I updated the example replacing the example `GlWindow` struct as it's simpler to use the trait.

I think this is quite a light touch, but still useful. Wdyt?

- [x] ~Tested on all platforms changed~
- [x] ~Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users~

### DIFF
--- a/glutin-winit/CHANGELOG.md
+++ b/glutin-winit/CHANGELOG.md
@@ -1,0 +1,15 @@
+# Unreleased
+
+- Add traits `GlWindow` with helper methods for building and resizing surfaces using a winit `Window`.
+
+# Version 0.2.1
+
+- Fix WGL window initialization.
+
+# Version 0.2.0
+
+- Fix API typo.
+
+# Version 0.1.0
+
+- Implement _glutin-winit_ helpers.
--- a/glutin-winit/src/window.rs
+++ b/glutin-winit/src/window.rs
@@ -1,0 +1,80 @@
+use glutin::context::PossiblyCurrentContext;
+use glutin::surface::{
+    GlSurface, ResizeableSurface, Surface, SurfaceAttributes, SurfaceAttributesBuilder,
+    SurfaceTypeTrait, WindowSurface,
+};
+use raw_window_handle::HasRawWindowHandle;
+use std::num::NonZeroU32;
+use winit::window::Window;
+
+/// [`Window`] extensions for working with [`glutin`] surfaces.
+pub trait GlWindow {
+    /// Build the surface attributes suitable to create a window surface.
+    ///
+    /// # Panics
+    /// Panics if either window inner dimension is zero.
+    ///
+    /// # Example
+    /// ```no_run
+    /// use glutin_winit::GlWindow;
+    /// # let winit_window: winit::window::Window = unimplemented!();
+    ///
+    /// let attrs = winit_window.build_surface_attributes(<_>::default());
+    /// ```
+    fn build_surface_attributes(
+        &self,
+        builder: SurfaceAttributesBuilder<WindowSurface>,
+    ) -> SurfaceAttributes<WindowSurface>;
+
+    /// Resize the surface to the window inner size.
+    ///
+    /// No-op if either window size is zero.
+    ///
+    /// # Example
+    /// ```no_run
+    /// use glutin_winit::GlWindow;
+    /// # use glutin::surface::{Surface, WindowSurface};
+    /// # let winit_window: winit::window::Window = unimplemented!();
+    /// # let (gl_surface, gl_context): (Surface<WindowSurface>, _) = unimplemented!();
+    ///
+    /// winit_window.resize_surface(&gl_surface, &gl_context);
+    /// ```
+    fn resize_surface(
+        &self,
+        surface: &Surface<impl SurfaceTypeTrait + ResizeableSurface>,
+        context: &PossiblyCurrentContext,
+    );
+}
+
+impl GlWindow for Window {
+    fn build_surface_attributes(
+        &self,
+        builder: SurfaceAttributesBuilder<WindowSurface>,
+    ) -> SurfaceAttributes<WindowSurface> {
+        let (w, h) = self.inner_size().non_zero().expect("invalid zero inner size");
+        builder.build(self.raw_window_handle(), w, h)
+    }
+
+    fn resize_surface(
+        &self,
+        surface: &Surface<impl SurfaceTypeTrait + ResizeableSurface>,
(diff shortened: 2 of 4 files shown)


glutin_examples/src/lib.rs
```diff
@@ -143,8 +142,8 @@ pub fn main() {
                         // Notable platforms here are Wayland and macOS, other don't require it
                         // and the function is no-op, but it's wise to resize it for portability
                         // reasons.
-                        if let Some((gl_context, gl_window)) = &state {
-                            gl_window.surface.resize(
+                        if let Some((gl_context, gl_surface, _)) = &state {
+                            gl_surface.resize(
                                 gl_context,
                                 NonZeroU32::new(size.width).unwrap(),
                                 NonZeroU32::new(size.height).unwrap(),
```
alexheretic, line 149: I left the resize usage plain here as it was before, though `GlWindow::resize_surface` could be used.

Note: In my game & examples I don't process resizes (which also involves more than just surface resizing) like this for each `WindowEvent::Resized` as (iirc) the events can outpace the resize operations. I found it better to process resizes after all events so multiple queue resizes would result in a single resize. For this case `GlWindow::resize_surface` de-clutters the most.

kchibisov:
Yes could you create a separate CHANGELOG in the glutin-winit crate?

glutin-winit/src/window.rs
```diff
@@ -0,0 +1,81 @@
...
+    ///
+    /// # Example
+    /// ```no_run
+    /// use glutin_winit::GlWindow;
+    /// # let winit_window: winit::window::Window = unimplemented!();
+    /// # let (gl_surface, gl_context) = unimplemented!();
+    ///
+    /// winit_window.resize_surface(&gl_surface, &gl_context);
+    /// ```
+    fn resize_surface(&self, surface: &Surface<WindowSurface>, context: &PossiblyCurrentContext);
```
kchibisov, line 40: Could we accept `&impl Traits` here instead? So it'll be more flexible.

glutin-winit/src/window.rs
```diff
@@ -0,0 +1,81 @@
...
+    fn resize_surface(&self, surface: &Surface<WindowSurface>, context: &PossiblyCurrentContext) {
+        if let Some((w, h)) = self.inner_size().non_zero() {
+            surface.resize(context, w, h)
+        }
+    }
+}
+
+/// [`winit::dpi::PhysicalSize<u32>`] non-zero extensions.
+pub trait NonZeroU32PhysicalSize {
+    /// Converts to non-zero `(width, height)`.
```
kchibisov, line 61: The trait is unused. Speaking aside it's fine to add such helper to winit types directly.

kchibisov (changes requested):

glutin-winit/src/window.rs
alexheretic: It is used, but could be make non-public. It just seemed useful to have working with winit physical sizes and glutin API.

kchibisov: You may want to submit that into winit instead? Like as I said, I don't see any issue with such thing, I just think that it may benefit winit directly, since you want to check that both of your sizes are not zero. So I'd suggest to move it there instead.

glutin-winit/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
+# Unreleased (0.2.2)
+- Add traits `GlWindow`, `NonZeroU32PhysicalSize` with helper methods for building and resizing surfaces
+  using a winit `Window`.
+
+# Version 0.2.1
+- Fix WGL window initialization.
```
kchibisov, line 6: Add a new line.

glutin-winit/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
...
+
+# Version 0.2.0
```
kchibisov, line 8: ditto.

glutin-winit/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
...
+- Fix API typo.
+
+# Version 0.1.0
```
kchibisov, line 11: ditto.

glutin-winit/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
+# Unreleased (0.2.2)
+- Add traits `GlWindow`, `NonZeroU32PhysicalSize` with helper methods for building and resizing surfaces
```
kchibisov, line 2:
```suggestion
# Unreleased

- Add traits `GlWindow`, `NonZeroU32PhysicalSize` with helper methods for building and resizing surfaces
```

kchibisov: Without the part about `NonZeroU32PhysicalSize`.

kchibisov (changes requested):

glutin-winit/src/window.rs
alexheretic: I'd rather make this private than block the pr on winit releases tbh

glutin-winit/CHANGELOG.md
alexheretic: I think it makes sense to use a explicit suggestion for this kind of edit, easy for you to write and me to apply. You can also just do such edits directly.

kchibisov:
I mean that making it private doesn't make any sense, since it's not used internally anyway? I'd rather remove for now and bring back if necessary.

alexheretic:
> I mean that making it private doesn't make any sense, since it's not used internally anyway? I'd rather remove for now and bring back if necessary.

But it is used, in both `GlWindow::build_surface_attributes` & `GlWindow::resize_surface`.

kchibisov (approved):

alexheretic:
Thanks for reviewing @kchibisov !

alexheretic:
@kchibisov any chance of a new release of _glutin-winit_? I'd like to make use of these features.

kchibisov:
I was going to bump along the winit bump, but I guess I can patch bump glutin-winit. Will do that later today.

kchibisov:
Released `v0.2.2`.

Next time you could open an issue if you need either glutin or glutin-winit released, since most bumps to them will likely be non-breaking.

PR #1546, opened 2022-12-20
Fix/address lints

alexheretic:
Some minor lint fixes I did to make rust-analyzer happy when I was doing #1545.

- [x] ~Tested on all platforms changed~
- [x] ~Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users~
- [x] ~Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior~
- [x] ~Created or updated an example program if it would help users understand this functionality~

### DIFF
--- a/glutin/src/api/egl/config.rs
+++ b/glutin/src/api/egl/config.rs
@@ -1,4 +1,5 @@
 //! Everything related to finding and manipulating the `EGLConfig`.
+#![allow(clippy::unnecessary_cast)] // needed for 32bit & 64bit support
 
 use std::ops::Deref;
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


kchibisov:
Be aware that those casts are here for a reason, since they handle 32bit vs 64bit differences.

And as you can clearly see your 32bit CI failed due to missmatched types. So we should likely silent that lint.

alexheretic:
> Be aware that those casts are here for a reason, since they handle 32bit vs 64bit differences.
> 
> And as you can clearly see your 32bit CI failed due to missmatched types. So we should likely silent that lint.

Ah that makes sense, thanks. I'll instead silence these then.

alexheretic:
I've added `allow(clippy::unnecessary_cast)` and left in the one other minor fix.

kchibisov (approved):

PR #1548, opened 2022-12-29
egl/config: fix wrong configs not being filtered

kchibisov:
This fixes an issue when the user passes the raw window handle to which the config must comply, however it wasn't taken into the account with EGL on xlib/xcb.

### DIFF
--- a/glutin/src/api/egl/config.rs
+++ b/glutin/src/api/egl/config.rs
@@ -5,6 +5,8 @@ use std::ops::Deref;
 use std::sync::Arc;
 use std::{fmt, mem};
 
+use raw_window_handle::RawWindowHandle;
+
 use glutin_egl_sys::egl;
 use glutin_egl_sys::egl::types::{EGLConfig, EGLint};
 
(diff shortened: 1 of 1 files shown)


glutin/src/api/egl/config.rs
```diff
@@ -184,7 +186,22 @@ impl Display {
...
+                            xcb.visual_id as u32 == config.native_visual()
+                        },
+                        Some(RawWindowHandle::Xlib(xlib)) => {
+                            xlib.visual_id as u32 == config.native_visual()
+                        },
+                        _ => true,
+                    };
+                }
+                keep &= !template.transparency || config.supports_transparency().unwrap_or(true);
+                keep
```
MarijnS95, line 204: Would it be clearer / cleaner to chain multiple `.filter()`s?

kchibisov: You can't put `cfg` right on it, so you'll have an empty filter...

MarijnS95: Not to put it on the `.filter()`, but to at least get rid of the `mut keep` and multiple `&=`.

`RawWindowHandle` doesn't put these enum variants and types behind a `cfg` anyway (but behind `c_ulong`) so I don't think you need the CFG anyway, or maybe on the match arms themselves if you want to consider those compatible if the backend was not enabled? `native_visual()` isn't behind a cfg either but might just return an error code when the attribute is not set?

glutin/src/api/egl/config.rs
```diff
@@ -184,7 +184,23 @@ impl Display {
...
+                let mut keep = true;
+                // Filter configs not compatible with the native window.
+                #[cfg(x11_platform)]
+                {
+                    use raw_window_handle::RawWindowHandle;
+                    keep &= match template.native_window {
+                        Some(RawWindowHandle::Xcb(xcb)) => {
+                            xcb.visual_id as u32 == config.native_visual()
+                        },
+                        Some(RawWindowHandle::Xlib(xlib)) => {
```
MarijnS95, line 196: `glx` pushes the `visual_id` into config attributes instead. Should the same happen here, or should filtering happen there, or both, or are they different mechanisms altogether?

kchibisov: You can't do what we're doing with GLX, since the visual is ignored in the EGL query, so you need to do that afterwards.

>EDIT: And should glx also handle xcb's visual id?

We force Xlib for glx, since we ask for Xlib error handler. And xcb doesn't have support for glx, so you use xlib with xcb anyway.

glutin/src/api/egl/config.rs
kchibisov: Hm, I guess that's fair.

MarijnS95 (approved):

glutin/src/api/egl/config.rs
MarijnS95: > You can't do what we're doing with GLX, since the visual is ignored in the EGL query, so you need to do that afterwards.

Ah okay, thanks for explaining that in the comment as well!

> > EDIT: And should glx also handle xcb's visual id?
> 
> We force Xlib for glx, since we ask for Xlib error handler. And xcb doesn't have support for glx, so you use xlib with xcb anyway.

Ah okay, I was reading conflicting information then since, as far as I understood, xcb is "just a wrapper library" on top of the x11 protocol, but I don't have any experience with these.

kchibisov: >Ah okay, I was reading conflicting information then since, as far as I understood, xcb is "just a wrapper library" on top of the x11 protocol, but I don't have any experience with these.

That's true, but GLX is also a separate protocol, so you must implement it as well in your wrapper and xcb doesn't do that. I'd like to not use xlib at all, but it's what it is.

PR #1549, opened 2023-01-02
Fixed EGL/GLX `Surface::width` returning height instead of width

Melchizedek6809:

Tested it by adding a temporary `println!()` outputting the `gl_window.surface.width()` and `gl_window.surface.height()`, which returned the height before, and now returns the expected value.


kchibisov (approved):

kchibisov:
Thanks!

PR #1551, opened 2023-01-02
Added `Surface::{width,height}` implementations for CGL/WGL

Melchizedek6809:

Hello again,

I've looked into implementing these missing methods on CGL/WGL. For the CGL implementation I'm not sure whether the nil check is necessary, if not then the maybe the return type could be changed to `u32` since all Platforms have an implementation now?


madsmtm (commented):
Ideally, the `!= nil` check in CGL should be moved to `create_window_surface`. Perhaps one could even change the type from `ns_view: id` to `ns_view: NonNull<Object>`.

Melchizedek6809:
@madsmtm Alright, I've moved the check into `create_window_surface`. Changing the type to `ns_view: NonNull<Object>` seems like a bigger change, I might tackle that in a separate PR.

Apart from that this newest change means that `Surface::{width,height}` will always return `Some` on all APIs, would changing the return type from `Option<u32>` to `u32` be appropriate?

kchibisov:
@MarijnS95 could you look into the android failure? It looks like the ndk just doesn't build, given that traces don't go into glutin?

glutin/src/api/cgl/surface.rs
```diff
@@ -86,11 +88,17 @@ impl<T: SurfaceTypeTrait> GlSurface<T> for Surface<T> {
     }
 
     fn width(&self) -> Option<u32> {
-        None
+        unsafe {
+            let frame: NSRect = msg_send![self.ns_view, frame];
```
kchibisov, line 92: What will happen when you call to that not on the main thread?

MarijnS95:
@kchibisov this was [caused by `num_enum 0.5.8`](https://github.com/illicitonion/num_enum/issues/94) which has aptly been yanked. I've restarted the CI, should become green soon again.

glutin/src/api/cgl/surface.rs
madsmtm: Good catch, I just sorta assumed that `Surface` was main-thread-only.

In practice, this will very likely work, since most Objective-C properties uses atomics internally. But in theory here be nasal demons.

A solution could be to wrap it in something like this that we also have in `winit`: https://github.com/rust-windowing/winit/blob/08ce3af3e18928cae2c943447d1fb88de3187a0b/src/platform_impl/macos/util/async.rs#L31-L37

Melchizedek6809: I've added an example that moves the Window off the main thread, which seems to work on Windows and Linux/X11 but crashes on MacOS with an `Illegal Instruction: 4` error even without my changes. I'll try and see whats the cause of that.

kchibisov: The macOS isn't really thread safe to be honest, but we should know whether it crashed in glutin or winit. Since both could do that.

That error means that we've tried to dispatch something on a thread we shouldn't use.

Melchizedek6809: @kchibisov Ahalright, it seems to be caused within glutin, the call to `setView_` within `ContextInner::make_current` seems to be the culprit. Do you know whether this can be fixed?

The full backtrace I get via LLDB looks as follows:

```
* thread #5, stop reason = EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP, subcode=0x0)
  * frame #0: 0x00007ff80b0104b4 AppKit`-[NSOpenGLContext setView:] + 189
    frame #1: 0x00000001000a3635 off_main_thread`_$LT$$LP$A$C$$RP$$u20$as$u20$objc..message..MessageArguments$GT$::invoke::h050a965616ecd9eb(imp=(libobjc.A.dylib`objc_msgSend), obj=0x000060000020ba00, sel=Sel @ 0x0000700001ac3110, (null)=(*mut objc::runtime::Object) @ 0x0000700001ac3140) at mod.rs:128:17
    frame #2: 0x00000001000a5e51 off_main_thread`objc::message::platform::send_unverified::ha544a2f8ddea49c3(obj=0x000060000020ba00, sel=Sel @ 0x0000700001ac31b0, args=(*mut objc::runtime::Object) @ 0x0000700001ac31d8) at mod.rs:27:9
    frame #3: 0x000000010009c5d2 off_main_thread`_$LT$$BP$mut$u20$objc..runtime..Object$u20$as$u20$cocoa..appkit..NSOpenGLContext$GT$::setView_::h829f6bdc297eb8d9 [inlined] objc::message::send_message::hc6eb294c5c5c7f1b(obj=0x000060000020ba00, sel=Sel @ 0x0000700001ac32b0, args=(*mut objc::runtime::Object) @ 0x0000700001ac32d0) at mod.rs:178:5
    frame #4: 0x000000010009c5ba off_main_thread`_$LT$$BP$mut$u20$objc..runtime..Object$u20$as$u20$cocoa..appkit..NSOpenGLContext$GT$::setView_::h829f6bdc297eb8d9(self=0x000060000020ba00, view=0x000000010450b450) at appkit.rs:2211:9
    frame #5: 0x00000001000072f7 off_main_thread`glutin::api::cgl::context::ContextInner::make_current::_$u7b$$u7b$closure$u7d$$u7d$::hea0311cbc8bcf548 at context.rs:219:13
    frame #6: 0x000000010001c2f6 off_main_thread`objc::rc::autorelease::autoreleasepool::h6973074709271cda(f={closure_env#0}<glutin::surface::WindowSurface> @ 0x0000700001ac33c0) at autorelease.rs:29:5
    frame #7: 0x000000010000726c off_main_thread`glutin::api::cgl::context::ContextInner::make_current::ha0b0e226658c6541(self=0x0000700001ac3450, surface=0x0000700001ac37a8) at context.rs:216:9
    frame #8: 0x000000010000709f off_main_thread`_$LT$glutin..api..cgl..context..NotCurrentContext$u20$as$u20$glutin..context..NotCurrentGlContextSurfaceAccessor$LT$T$GT$$GT$::make_current::he542782203ec0f72(self=NotCurrentContext @ 0x0000700001ac3450, surface=0x0000700001ac37a8) at context.rs:93:9
    frame #9: 0x0000000100011917 off_main_thread`_$LT$glutin..context..NotCurrentContext$u20$as$u20$glutin..context..NotCurrentGlContextSurfaceAccessor$LT$T$GT$$GT$::make_current::he5b45ac66cc96fe0(self=NotCurrentContext @ 0x0000700001ac36e0, surface=0x0000700001ac37a8) at context.rs:404:48
    frame #10: 0x0000000100012bed off_main_thread`off_main_thread::main::_$u7b$$u7b$closure$u7d$$u7d$::_$u7b$$u7b$closure$u7d$$u7d$::h05a7ee092c6f388e at off_main_thread.rs:121:38
    frame #11: 0x0000000100009d42 off_main_thread`std::sys_common::backtrace::__rust_begin_short_backtrace::h699e07ea3162f75a(f=<unavailable>) at backtrace.rs:121:18
    frame #12: 0x000000010001c0b2 off_main_thread`std::thread::Builder::spawn_unchecked_::_$u7b$$u7b$closure$u7d$$u7d$::_$u7b$$u7b$closure$u7d$$u7d$::h55d6008904560451 at mod.rs:551:17
    frame #13: 0x000000010000fbb2 off_main_thread`_$LT$core..panic..unwind_safe..AssertUnwindSafe$LT$F$GT$$u20$as$u20$core..ops..function..FnOnce$LT$$LP$$RP$$GT$$GT$::call_once::h5994e9127706bfa8(self=<unavailable>, _args=<unavailable>) at unwind_safe.rs:271:9
    frame #14: 0x00000001000171e4 off_main_thread`std::panicking::try::do_call::h74211c14ae498b45(data="") at panicking.rs:483:40
    frame #15: 0x00000001000172ed off_main_thread`__rust_try + 29
    frame #16: 0x0000000100017086 off_main_thread`std::panicking::try::h63010fe35faccfce(f=<unavailable>) at panicking.rs:447:19
    frame #17: 0x000000010002eb42 off_main_thread`std::panic::catch_unwind::h4db136f8d16dec6c(f=<unavailable>) at panic.rs:137:14
    frame #18: 0x000000010001bf1d off_main_thread`std::thread::Builder::spawn_unchecked_::_$u7b$$u7b$closure$u7d$$u7d$::h80add62b5c1ad3d8 at mod.rs:550:30
    frame #19: 0x0000000100027441 off_main_thread`core::ops::function::FnOnce::call_once$u7b$$u7b$vtable.shim$u7d$$u7d$::hd410c5011bc272b3((null)=0x00006000021038e0, (null)=<unavailable>) at function.rs:251:5
    frame #20: 0x00000001000d9057 off_main_thread`std::sys::unix::thread::Thread::new::thread_start::h7b576c3bd89f934a [inlined] _$LT$alloc..boxed..Box$LT$F$C$A$GT$$u20$as$u20$core..ops..function..FnOnce$LT$Args$GT$$GT$::call_once::h2611f89e824929e3 at boxed.rs:1987:9 [opt]
    frame #21: 0x00000001000d9051 off_main_thread`std::sys::unix::thread::Thread::new::thread_start::h7b576c3bd89f934a [inlined] _$LT$alloc..boxed..Box$LT$F$C$A$GT$$u20$as$u20$core..ops..function..FnOnce$LT$Args$GT$$GT$::call_once::hb87e0c9c6cb0305b at boxed.rs:1987:9 [opt]
    frame #22: 0x00000001000d904a off_main_thread`std::sys::unix::thread::Thread::new::thread_start::h7b576c3bd89f934a at thread.rs:108:17 [opt]
    frame #23: 0x00007ff8081cd4e1 libsystem_pthread.dylib`_pthread_start + 125
    frame #24: 0x00007ff8081c8f6b libsystem_pthread.dylib`thread_start + 15
```

kchibisov: Could you show what you've done wrt multithreading? I have a feeling that you're doing something wrong.

kchibisov: Hm, We might need some extra calls around that on the main thread. Could you bring `libdispatch` and issue the calls on the main thread like `setView` and `update`? And `block` for them until completion?

That way it should probably work.

kchibisov: @madsmtm I'm pretty sure you have helpers for stuff like that, right? What's your recommendation would be?

Melchizedek6809: @kchibisov Sure, it's a new example in this PR which can be found under `glutin_examples/examples/off_main_thread.rs`. Very likely that I'm doing something wrong, though a better error message would be much appreciated then :)

Melchizedek6809: | Hm, We might need some extra calls around that on the main thread. Could you bring libdispatch and issue the calls on the main thread like setView and update? And block for them until completion?

Alright, I'll try that.

kchibisov: The main issue is that apply requires some operations to always be on the `main` thread. And given that things like `setView` are not documented at all, you have to guess. Last time we've checked it worked for example.

I've just found an SDL2 thread about those operations requiring a main thread, so libdispatch should be used to issue them on the main thread, otherwise you'll end with SIGILL.

Melchizedek6809: @kchibisov Could you take a look at this issue? I can't seem to figure out a good way to run these commands on the `main` thread without having to rework a lot of the backend.

madsmtm: Ooops, I had this comment left in my "pending review":

> Honestly, I'm still unsure of what "the best way" to do it is (though I think https://github.com/madsmtm/objc2/pull/344 is a step in the correct direction).
>
> But as I alluded to previously, just wrapping everything that touches `Surface::ns_view` or `ContextInner::raw` in a `run_on_main` helper function would definitely work.

kchibisov: I'll try to look into it once I have time, but you could look into https://github.com/rust-windowing/winit/blob/2e4d79f57a9cdc3124e78d3fabc4da3daea7748c/src/platform_impl/macos/util/async.rs#L76 on how to run on main, it shouldn't require much changes, since
apple uses libdispatch and has a concept of dispatching on the main thread. You'd likely need to block on the result as wait...

kchibisov:
@madsmtm could you take a look, I think I've fixed all thread safety issues with macOS and were able to run stuff off the main thread, though, I've removed the example due to its complexity...

I've also clarified the docs and ensured that we return physical sizes from the `width`/height` functions.

kchibisov:
Also, @madsmtm do you know a way to somehow wakeup macOS's main thread from `set_wait` in winit? Without side channels and such, maybe there's a global method to magically wake-it-up?

madsmtm (approved):
> @madsmtm could you take a look

Looks good to me now 👍

> Also, @madsmtm do you know a way to somehow wakeup macOS's main thread from `set_wait` in winit? Without side channels and such, maybe there's a global method to magically wake-it-up?

I assume you mean after having called `ControlFlow::set_wait`? But yes, what we do in `winit` to do so is to send an empty `NSEvent`:

https://github.com/rust-windowing/winit/blob/5ba6bdef4931e66eb4c4be87c90a51fad7675ee5/src/platform_impl/macos/event_loop.rs#L260

madsmtm:
You definitely could - though I don't think I understand why you need to?

madsmtm:
No, you needn't worry about that: `Queue::main().exec_sync(|| ...)` wakes up the main event loop for you

kchibisov:
Huh, ok.

Issue #1550, opened 2023-01-02
Can't create context on old Win 10 Laptop


Melchizedek6809:
When trying to run the example it panics because it can't create an ES context (I suppose it tried creating a GL one before). 
The Device in question is an X220t with an i5-2520M / HD Graphics 3000 running up-to date Windows 10.
The same machine can create/run an OpenGL 3.1 context with old glutin just fine, so the hardware/driver should be fine.

```
$ cargo run --example window
    Blocking waiting for file lock on build directory
   Compiling glutin_examples v0.1.3 (C:\Users\benny\src\glutin\glutin_examples)
    Finished dev [unoptimized + debuginfo] target(s) in 3.16s
     Running `target\debug\examples\window.exe`
Picked a config with 0 samples
thread 'main' panicked at 'failed to create context: Error { raw_code: None, raw_os_message: None, kind: NotSupported("extension to create ES context with wgl is not present") }', glutin_examples\src\lib.rs:74:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: process didn't exit successfully: `target\debug\examples\window.exe` (exit code: 101)
```

kchibisov:
I think this was resolved.

PR #1553, opened 2023-01-10
Lowered fallback GL Version from 3.3 -> 2.1

Melchizedek6809:

This fixes an issue I've opened: https://github.com/rust-windowing/glutin/issues/1550

### DIFF
--- a/glutin_examples/src/lib.rs
+++ b/glutin_examples/src/lib.rs
@@ -9,7 +9,7 @@ use winit::window::{Window, WindowBuilder};
 use raw_window_handle::HasRawWindowHandle;
 
 use glutin::config::{Config, ConfigTemplateBuilder};
-use glutin::context::{ContextApi, ContextAttributesBuilder};
+use glutin::context::{ContextApi, ContextAttributesBuilder, Version};
 use glutin::display::GetGlDisplay;
 use glutin::prelude::*;
 use glutin::surface::{Surface, SurfaceAttributesBuilder, SwapInterval, WindowSurface};
@@ -77,11 +77,22 @@ pub fn main() {
     let fallback_context_attributes = ContextAttributesBuilder::new()
         .with_context_api(ContextApi::Gles(None))
         .build(raw_window_handle);
+
+    // There are also some old devices that support neither modern OpenGL nor GLES.
+    // To support these we can try and create a 2.1 context.
+    let legacy_context_attributes = ContextAttributesBuilder::new()
+        .with_context_api(ContextApi::OpenGl(Some(Version::new(2, 1))))
+        .build(raw_window_handle);
+
     let mut not_current_gl_context = Some(unsafe {
         gl_display.create_context(&gl_config, &context_attributes).unwrap_or_else(|_| {
-            gl_display
-                .create_context(&gl_config, &fallback_context_attributes)
-                .expect("failed to create context")
+            gl_display.create_context(&gl_config, &fallback_context_attributes).unwrap_or_else(
+                |_| {
+                    gl_display
+                        .create_context(&gl_config, &legacy_context_attributes)
+                        .expect("failed to create context")
+                },
+            )
         })
     });
 


kchibisov:
I'm not sure that this behavior is what should be in the modern lib. Be aware that OpenGL 3.3 code your likely to write is not compatible with the `Compatible' profile. So if you support compatible profile you should explicitly ask for it. Fallback to it won't work most of the time if you don't know that you're going to use compat profile.

What you could do is to ask for the compat profile in the examples.

Melchizedek6809:
But shouldn't programs that requires a 3.3 core context be explicit about it? Keep in mind that if asked for an unspecified core context gluting still defaults to 3.3 as before, this only changes things when nothing specific was requested.

I at least was quite suprised to see that glutin would then default to a modern OpenGL version like 3.3, since I'd assume a library to default/fallback to some highly compatible version when not asked for something specific, or at least to check whether that particular more modern version is supported by the system.

GLFW for example seems to open a context with the highest supported version if not asked for a specific version, and
SDL2 also creates a 2.1 context on my machine when not asked for something specific.

Since this seems to be somewhat common behaviour I think glutin would be improved by not diverging here, with a default to 2.1 probably being the simpler solution.

kchibisov:
>But shouldn't programs that requires a 3.3 core context be explicit about it? Keep in mind that if asked for an unspecified core context gluting still defaults to 3.3 as before, this only changes things when nothing specific was requested.

Every modern program will require either Core or Gles, compatible is something that may not be even present with some hardware that is new, but has low version.

>GLFW for example seems to open a context with the highest supported version

You can't get highest supported version if you're not asking for it. Moreover be aware that some systems create a context with maximum supported version if you don't pass anything to them automatically.

>Since this seems to be somewhat common behaviour I think glutin would be improved by not diverging here, with a default to 2.1 probably being the simpler solution.

Well I guess, but you should clearly ask for profile. Maybe we should enforce the profile, since you'll stick with some version of it anyway?

Melchizedek6809:
> Every modern program will require either Core or Gles, compatible is something that may not be even present with some hardware that is new, but has low version.

Yeah, it's just that old pre 3.2 devices don't support a Core context and asking for it results in an Error.

> You can't get highest supported version if you're not asking for it. Moreover be aware that some systems create a context with maximum supported version if you don't pass anything to them automatically.

I suppose this is being done by GLFW to provide that functionality, it just seemed like a more complicated (although more convenient) solution.

> Well I guess, but you should clearly ask for profile. Maybe we should enforce the profile, since you'll stick with some version of it anyway?

Maybe even enforce that a Version/Profile pair be provided? That should get rid of all the guesswork.

Apart from that, another thing that I've been wondering about, there seems to be no clear way to create a vanilla 2.1 Context. From what I've been reading `ARB_Compatibility` got introduced with 3.1 and the Core/Compatibility split came in 3.2. So there should be a way to request a non-Compatibility, non-Core context. It seems that changing the return type of `context::pick_profile` from `(GlProfile, Version)` to `(Option<GlProfile>, Version)` should help here, since from what I can tell it's stored as an Option during context creation anyways, it's just forced to `Some` after the call to `context::pick_profile`.

kchibisov:
>Yeah, it's just that old pre 3.2 devices don't support a Core context and asking for it results in an Error.

I mean that's true, but you should specify the version. For example you may have the same error on modern devices when compatible profile is not present. Not specifying the version
is also not really an option, since Android for example will pick GLES 1.0 while mesa will
pick OpenGL Core 4.6.(As the maximum supported one). And this behavior is no near stable.

>I suppose this is being done by GLFW to provide that functionality, it just seemed like a more complicated (although more convenient) solution.

It was done by glutin before, and the only way to do that is basically iterate over all versions in decreasing order and try. But tbh I don't find this behavior any good, because you always have only subset of versions supported, so if you can do the probing yourself.

>Apart from that, another thing that I've been wondering about, there seems to be no clear way to create a vanilla 2.1 Context. From what I've been reading ARB_Compatibility got introduced with 3.1 and the Core/Compatibility split came in 3.2. So there should be a way to request a non-Compatibility, non-Core context. It seems that changing the return type of context::pick_profile from (GlProfile, Version) to (Option<GlProfile>, Version) should help here, since from what I can tell it's stored as an Option during context creation anyways, it's just forced to Some after the call to context::pick_profile.

I mean that it's not really a thing anymore? If you don't specify the profile for OpenGL for example mesa will always assign you one. If you read for example EGL specification wrt version policy you'll see that they always map to compat/core when it comes to OpenGL. If on Windows it behaves differently we may adjust that, sure.

With GLES it's a bit different, since profiles are not really a thing for it.

Melchizedek6809:
I've changed the example to now try a 2.1 context when creating the 3.3 core and 3 ES fails. This fixes the example, at least on my old laptop.

kchibisov:
Could you rebase?

kchibisov (changes requested):
Actually, we might at use compat for (`None`,`None`). We'd also need to update the docs wrt that. Since right now it says that when you don't provide the `ContextApi` it'll pick the supported api. However the issue is that we can't say for sure what is supported. Maybe we should request a context without a profile at all in this case and just say that version is 2.0?

I remember that there's an issue that android for example could pick GLES 1.0 if you don't tell it anything at all, which will fail on the example(not that I care that much about the example failure).

I guess we should try to ask for something that is more likely to work and state that. So requesting without a profile just version 2.0 should work?

What do you think?

Melchizedek6809:
> Could you rebase?

Sure, just rebased

Melchizedek6809:
> I guess we should try to ask for something that is more likely to work and state that. So requesting without a profile just version 2.0 should work?

Yeah, a plain 2.0 context when nothing in particular is requested sounds good.

Though I suppose in the long run it might be  better to encourage users to request particular versions/profiles, or maybe even require a version/profile combination on context creation? Since whatever version will be chosen as the default one will be wrong in many cases, and even probing for the highest supported version gets complicated by the presence of ES.

What do you think?

Melchizedek6809:
> I guess we should try to ask for something that is more likely to work and state that. So requesting without a profile just version 2.0 should work?

Yeah, a plain 2.0 context when nothing in particular is requested sounds good.

Though I suppose in the long run it might be  better to encourage users to request particular versions/profiles, or maybe even require a version/profile combination on context creation? Since whatever version will be chosen as the default one will be wrong in many cases, and even probing for the highest supported version gets complicated by the presence of ES.

What do you think?

kchibisov:
>Though I suppose in the long run it might be better to encourage users to request particular versions/profiles, or maybe even require a version/profile combination on context creation? Since whatever version will be chosen as the default one will be wrong in many cases, and even probing for the highest supported version gets complicated by the presence of ES.

The thing is that you don't have profiles for ES context at all. We can keep it that way for now...

Melchizedek6809:
> The thing is that you don't have profiles for ES context at all.

Yeah, though we could check in the `build` method and only require a profile for OpenGl context. 

> We can keep it that way for now...

Alright, what about this PR? It would be nice if the example worked on older machines out of the box, and now it only changes the example code itself.

glutin_examples/src/lib.rs
```diff
@@ -77,11 +77,22 @@ pub fn main() {
     let fallback_context_attributes = ContextAttributesBuilder::new()
         .with_context_api(ContextApi::Gles(None))
         .build(raw_window_handle);
+
+    // There are also some old devices that support neither modern OpenGL nor GLES.
+    // To support these we can try and create a 2.1 context.
+    let legacy_context_attributes = ContextAttributesBuilder::new()
+        .with_context_api(ContextApi::OpenGl(Some(Version { major: 2, minor: 1 })))
```
kchibisov, line 84: Use `new` method.

Melchizedek6809: Alright, I'm using `Version::new` now.

kchibisov (approved):

PR #1554, opened 2023-01-15
ci: fix android ci job

kchibisov:
This should install cargo-apk using stable and not the MSRV, since testing its MSRV is irrelevant for glutin.


MarijnS95:
Argh one of our dependencies had an MSRV bump :angry:

kchibisov:
I mean it doesn't really matter, since it's a tool to run stuff.

MarijnS95:
Not to glutin, but it's super annoying when MSRVs are bumped without breaking semver.

PR #1557, opened 2023-01-17
glx: harden error handling

kchibisov:
This also adjusts the use of the hook for the upcoming winit update which has a bug leaking errors. Now glutin hook will handle when it forces sync with X server.

### DIFF
--- a/glutin/src/api/glx/mod.rs
+++ b/glutin/src/api/glx/mod.rs
@@ -3,7 +3,7 @@
 
 use std::ffi::{self, CStr, CString};
 use std::ops::{Deref, DerefMut};
-use std::sync::atomic::{AtomicI32, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicI32, Ordering};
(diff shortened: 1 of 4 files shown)


glutin/src/api/glx/mod.rs
```diff
@@ -162,13 +174,26 @@ fn glx_error_hook(_display: *mut ffi::c_void, xerror_event: *mut ffi::c_void) ->
 
 /// Get the error from the X11.
 fn last_glx_error(display: display::GlxDisplay) -> Result<()> {
+    // Ensure that the access is safe by locking the display.
+    unsafe {
+        (XLIB.as_ref().unwrap().XLockDisplay)(*display as *mut _);
```
kchibisov, line 179: Note, that operations accessing the display are safe, since XSync has implicit Lock(locks on Xlib are recursive). But to ensure that we're not clearing `last_error` I've added extra lock on the display just in case...

PR #1560, opened 2023-02-02
glutin-winit: Update to winit 0.28

alexheretic:
Also fix the rust 1.67 clippy lints (since these are deny).

@kchibisov can this be released as _glutin-winit_ `0.3`?

- [x] Tested on all platforms changed _(assuming winit itself was tested)_
- ~[x] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior~
- ~[x] Created or updated an example program if it would help users understand this functionality~

### DIFF
--- a/glutin_examples/Cargo.toml
+++ b/glutin_examples/Cargo.toml
@@ -16,17 +16,17 @@ egl = ["glutin-winit/egl", "png"]
 glx = ["glutin-winit/glx"]
 wgl = ["glutin-winit/wgl"]
 x11 = ["glutin-winit/x11"]
-wayland = ["glutin-winit/wayland", "winit/wayland-dlopen", "winit/wayland-csd-adwaita-notitle"]
+wayland = ["glutin-winit/wayland", "winit/wayland-dlopen", "winit/wayland-csd-adwaita"]
 
 [dependencies]
 glutin = { path = "../glutin", default-features = false }
-winit = { version = "0.28", default-features = false }
+winit = { version = "0.28.1", default-features = false }
 glutin-winit = { path = "../glutin-winit", default-features = false }
 raw-window-handle = "0.5.0"
 png = { version = "0.17.6", optional = true }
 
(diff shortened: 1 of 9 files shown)


kchibisov:
Seems like you should add the responsible winit features for android, like game/native activity as well. Could you also point to winit 0.28.1 explicitly.

>@kchibisov can this be released as glutin-winit 0.3?

Sure.

glutin_examples/Cargo.toml
```diff
@@ -20,7 +20,7 @@ wayland = ["glutin-winit/wayland", "winit/wayland-dlopen", "winit/wayland-csd-ad
 
 [dependencies]
 glutin = { path = "../glutin", default-features = false }
```
kchibisov, line 22: I think you can also disable `winit/wayland-csd-adwaita-notitle` since winit is using `ab_glyph` now by default.

kchibisov (changes requested):

glutin_examples/examples/android.rs
```diff
@@ -1,6 +1,10 @@
 #![cfg(target_os = "android")]
 
-#[ndk_glue::main(backtrace = "on")]
-fn main() {
-    glutin_examples::main()
+use winit::event_loop::EventLoopBuilder;
+use winit::platform::android::EventLoopBuilderExtAndroid;
+
+#[no_mangle]
+fn android_main(app: winit::platform::android::activity::AndroidApp) {
```
MarijnS95, line 7: Nit: I'd import `AndroidApp` just like the trait is imported, instead of fully qualifying it.

MarijnS95 (approved):
Not tested but this looks sensible.

alexheretic:
> Not tested but this looks sensible.

Tbf android isn't really a platform I can help support. I just followed the migration recommendations from the winit repo to make it look correct & compile.

kchibisov (approved):

PR #1567, opened 2023-02-09
Fix initializing glutin display with winit provided window handle on X11

tronical:
Commit f3ea5be2d6194d84ea370e6d59f4ee898325dd5c introduced a match on the window handle providex X11 visual with the EGL config. Unfortunately winit's implementation of raw_window_handle() sets the visual_id to 0, which is documented as the unknown visual. Handle this setup by filtering only if the visual id is known.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fix EGL/GLX display initialization when the provided raw-window-handle has an unknown visual_id.
+
 # Version 0.30.4
 
 - Fixed EGL display initialization with XcbDisplayHandle.
--- a/glutin/src/api/egl/config.rs
+++ b/glutin/src/api/egl/config.rs
@@ -191,10 +191,10 @@ impl Display {
                 // XXX This can't be done by passing visual in the EGL attributes
                 // when calling `eglChooseConfig` since the visual is ignored.
                 match template.native_window {
-                    Some(RawWindowHandle::Xcb(xcb)) => {
+                    Some(RawWindowHandle::Xcb(xcb)) if xcb.visual_id > 0 => {
                         xcb.visual_id as u32 == config.native_visual()
                     },
-                    Some(RawWindowHandle::Xlib(xlib)) => {
+                    Some(RawWindowHandle::Xlib(xlib)) if xlib.visual_id > 0 => {
                         xlib.visual_id as u32 == config.native_visual()
                     },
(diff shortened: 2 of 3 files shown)


kchibisov:
I might suggest to fix winit along the way warn when you pass `0`? Though maybe it's fine to ignore on `0`?

The visual stuff on EGL should be broken right now, because we don't keep native display around, which is unfortunate copy-paste error at some point.

tronical:
> I might suggest to fix winit along the way warn when you pass `0`? Though maybe it's fine to ignore on `0`?

In raw-window-handle at least 0 is documented as "unknown". That lead me to thinking that it's fine to ignore on 0 (in this patch).

> I guess we should convert it to `Err` instead, because the thing is that to actually render visual on window and config visual should be the same, and asking for config that can render to that visual we should assume non-zero, otherwise you shouldn't pass it at all for X11.

True true. But then as a "user" I could work around it by passing None as raw-window-handle and I'd still get my display, won't I?

> Patching winit should be trivial though.

That's true. And I agree that should lead to best results, then we can actually match properly. Meanwhile my preference is glutin handling the "unknown". If you prefer that one shouldn't use glutin with a r-w-h that has a zero visual, then I can work around it on the call site until a winit release is available, no problem.

kchibisov (changes requested):
For now we can ignore `zero` adding a note in docs, that 

>True true. But then as a "user" I could work around it by passing None as raw-window-handle and I'd still get my display, won't I?

The `None` is fine, it means that you pass the visual to winit yourself(which is preferred way to do that anyway) from the config you've matched. Because winit doesn't provide you with visual picking at all(nor that it should).

Let's just ignore zero and open a bug in winit to set the `visual_id`. The same thing you should do for GLX.

Be aware that the thing is that if your config's visual doesn't match window visual the stuff will fail, (that's why you can see in examples/glutin-winit reverse order on X11, where you match config and pass visual to winit, since it's more reliable to actually create anything).

Only on windows you'd need to actually pass this value.

tronical:
>  The same thing you should do for GLX.

Good point, done :)

tronical:
> Only on windows you'd need to actually pass this value.

Thank you for the explanation.

kchibisov (approved):

tronical:
Filed https://github.com/rust-windowing/winit/issues/2681

kchibisov:
Thanks.

PR #1572, opened 2023-02-11
chore: derive `Default` for enums when possible

kchibisov:


kchibisov:
This required rust 1.62.0.

PR #1575, opened 2023-02-15
 fix: remove unused error-prone submodule setup

proletarius101:
According to https://git-scm.com/docs/gitsubmodules, there must be a working directory at `path/to/bar/` if there's a `submodule.foo.path = path/to/bar` entry in `.gitmodules`. Failure to have this will result in errors in certain build environments (e.g. Flatpak: https://github.com/standardnotes/app/pull/516)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "deps/apk-builder"]
-	path = deps/apk-builder
-	url = https://github.com/rust-windowing/android-rs-glue


MarijnS95:
Ah, @kchibisov also assigned me as reviewer, likely to check. No, I am not aware of any use of `apk-builder`, the CI uses `cargo-apk` which it installs ad-hoc (and the readme documents this for users. The submodule should be removed instead.

kchibisov (approved):

kchibisov:
>Ah, @kchibisov also assigned me as reviewer, likely to check. No, I am not aware of any use of apk-builder

Thanks, for confirming, that it's redundant. I also had a feeling that this file was  just laying around without anyone noticing it.

@proletarius101 thanks for removing it :p

PR #1578, opened 2023-02-26
Improved examples by replacing state with enum

zatrit:

Might need to test this on different platforms as I only tested on Windows and Android


kchibisov:
I'm not quite sure that I like such changes, one approach in simplifying is using `treat_as_possibly_current` and remove all enum stuff all together keeping the option.

kchibisov:
If you want to update to use `treat_as_possibly_current` you could continue in this PR.

PR #1582, opened 2023-03-24
Make display, surface and context creation safe

notgull:

Uses rust-windowing/raw-window-handle#116 to make display, surface and context creation safe.

I want to get a good idea of what this kind of API would look like. I want to test out the new safe window handles before they're released onto the world.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+- **Breaking:** Bump `raw-window-handle` to v0.6.
+- **Breaking:** Make `glutin` types generic over the window/display they are wrapping.
+
 # Version 0.31.2
 
 - Fixed EGL not setting context version with EGL versions before 1.5 and missing context ext.
--- a/glutin-winit/Cargo.toml
+++ b/glutin-winit/Cargo.toml
@@ -19,8 +19,8 @@ wayland = ["glutin/wayland", "winit/wayland"]
 
 [dependencies]
 glutin = { version = "0.31.0", path = "../glutin", default-features = false }
-raw-window-handle = "0.5.2"
-winit = { version = "0.29.2", default-features = false, features = ["rwh_05"] }
+raw-window-handle = { version = "0.6.0", features = ["std"] }
+winit = { version = "0.29.2", default-features = false, features = ["rwh_06"] }
 
 [build-dependencies]
 cfg_aliases = "0.1.1"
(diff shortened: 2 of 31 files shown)


notgull:
Hmm, looks like this model kind of breaks `glutin-winit`. Right now, an `&EventLoopWindowTarget` is used to provide the display handle for creating the `Display`. However, since the reference to the ELWT will disappear for event handling, this can't be used as a persistent handle for the display.

A solution would be to use an `&Window` or an `Rc<Window>` instead, but the `Display` initialization generally happens before the `Window` is created. Maybe we could add a type to `winit` similar to `EventLoopWindowTarget` that doesn't disappear during event handling?

notgull:
Rebased around rust-windowing/raw-window-handle#125, aka `raw-window-handle` v0.6

notgull:
Rebased on the newest master

notgull:
The main caveat is that the `OwnedDisplayHandle` change wasn't accepted into `winit`, so the `DisplayBuilder` function is now `unsafe`.

emilk:
Any chance for a glutin release with `raw-window-handle` 0.6? 🙏

kchibisov:
I'm not sure how soon, since I don't have time to review the glutin stuff now. it's not like you can't use it at all if you have only raw-window-handle 0.6, since you can convert 0.6 to 0.5 and e.g. winit provides all 3 of them at the same time, so you can pick 0.5 for glutin and 0.6 for other winit stuff.

But it'll be review/merge within a month...

emilk:
Ah, winit supports all three of them at the same time, I didn't realize that (thought I had to pick one). Then I can probably figure something out. Thanks for the pointer!

MarijnS95:
@emilk what makes you juggle multiple versions, instead of locking winit to 0.5 and using that in glutin?

emilk:
[`eframe`](https://crates.io/crates/eframe) has to support both `glutin` (locked at 0.5) and `wgpu` (locked at 0.6).

MarijnS95:
Right, that is indeed a painful situation.

notgull:
Don't really feel like rebasing this and it's unlikely to be reviewed, so I'll abandon this in favor of #1670

kchibisov:
It'll be though, but there's a big chance that I write parts myself. Like I'd like to use safe handles myself, it's just winit is lacking and I said it to you before. Forcing everyone to just start `Arc` everything sounds a bit wasteful.

Issue #1566, opened 2023-02-08
Unable to create surface on old i915 system with EGL on X11


Tobi29:
Whilst trying to make some glutin based code work on an old machine with Intel 945GME for graphics, I noticed a couple of issues with the EGL backend of glutin when running on Linux with X11:

- [`NativeWindow::raw`](https://github.com/rust-windowing/glutin/blob/master/glutin/src/api/egl/surface.rs#L502) returns a pointer to the window id, instead of the actual window id. This causes `eglCreateWindowSurface` to fail with `EGL_BAD_NATIVE_WINDOW` inside of `Display::create_window_surface`. (There is more code like that around for XCB and pixmaps, which do the same dodgy cast.)
- [`DisplayInner::_native_display`](https://github.com/rust-windowing/glutin/blob/master/glutin/src/api/egl/display.rs#L464) is never set to anything but `None` by both `Display::new` and `Display::with_device`. This results in `X11GlConfigExt::x11_visual` always returning `None`.

Both of these can be avoided by using GLX instead, which is fine, but technically this *should* work with EGL as well (and it does with at least the first one fixed).
Not a deal breaker, but they both look like oversights to me and they do not cause any issues on newer systems it seems, making them annoying to troubleshoot.

The old system is running Linux 6.1.8 and Mesa 21.3.9 Amber for reference.

kchibisov:
>[NativeWindow::raw](https://github.com/rust-windowing/glutin/blob/master/glutin/src/api/egl/surface.rs#L502) returns a pointer to the window id, instead of the actual window id. This causes eglCreateWindowSurface to fail with EGL_BAD_NATIVE_WINDOW inside of Display::create_window_surface. (There is more code like that around for XCB and pixmaps, which do the same dodgy cast.)

The API accepts a pointer though. Have you tried passing `XID` and not `*XID`? For me it was straight segfaulting without a cast (which it should for you as well). Like as you can see we take a pointer to a XID not cast XID to pointer.

>[DisplayInner::_native_display](https://github.com/rust-windowing/glutin/blob/master/glutin/src/api/egl/display.rs#L464) is never set to anything but None by both Display::new and Display::with_device. This results in X11GlConfigExt::x11_visual always returning None.


Now that's a bug, I'm sure it wasn't like that before, and it got somehow changed...

Tobi29:
[...]

PR #1585, opened 2023-04-01
egl/surface: Don't dereference native window pointer before passing to legacy `CreateWindowSurface`

### LINKED ISSUE #1584: Android example crash
When using the latest release version on Android 8.1:
```
cargo apk r -p glutin_examples --example android
```
I get:
```
03-27 08:10:45.476  4018  4018 I zygote64: Late-enabling -Xcheck:jni
03-27 08:10:45.619  4018  4018 W System  : ClassLoader referenced unknown path:
03-27 08:10:45.704  4018  4023 I zygote64: Do partial code cache collection, code=30KB, data=22KB
03-27 08:10:45.705  4018  4023 I zygote64: After code cache collection, code=30KB, data=22KB
03-27 08:10:45.705  4018  4023 I zygote64: Increasing code cache capacity to 128KB
03-27 08:10:45.812  4018  4018 W Thread-2: type=1400 audit(0.0:83): avc: denied { search } for name="clients" dev="debugfs" ino=6282 scontext=u:r:untrusted_app:s0:c512,c768 tcontext=u:object_r:debugfs_ion:s0 tclass=dir permissive=0
03-27 08:10:45.822  4018  4035 I zygote64: android::hardware::configstore::V1_0::ISurfaceFlingerConfigs::hasWideColorDisplay retrieved: 0      
03-27 08:10:45.824  4018  4034 I RustStdoutStderr: Picked a config with 4 samples
03-27 08:10:45.842  4018  4018 I SurfaceFactory: [static] sSurfaceFactory = com.mediatek.view.impl.SurfaceFactoryImpl@ddb8066
03-27 08:10:45.849  4018  4018 D WindowClient: Add to mViews: DecorView@74ed7fd[NativeActivity], this = android.view.WindowManagerGlobal@d0f3cf2
03-27 08:10:45.857  4018  4018 V PhoneWindow: DecorView setVisiblity: visibility = 0, Parent = ViewRoot{93adf43 rust.example.android/android.app.NativeActivity,ident = 0}, this = DecorView@74ed7fd[NativeActivity]
03-27 08:10:45.896  4018  4034 I RustStdoutStderr: Android window available
--------- beginning of crash
03-27 08:10:45.897  4018  4035 F libc    : Fatal signal 11 (SIGSEGV), code 1, fault addr 0xc05f776ef4 in tid 4035 (Thread-2), pid 4018 (example.android)
03-27 08:10:45.976  4018  4035 I libc    : orig_dumpable: 1
```
What could this be? WebGL works on this device.

(When running eframe there is an error with string "OpenGL ES-CM 1.1", I guess that's the ES version, but ES 3.1 should be supported on Android 8.1)

MarijnS95:
Fixes #1584

Just like for KHR and EXT the native window _is_ `egl::NativeWindowType` (which is a `*const c_void` under the hood).  By casting it to _a pointer of_ `egl::NativeWindowType` instead, and derferencing that, we'd be reading the first few bytes of whatever this opaque handle was pointing to (if it even is a valid pointer) and segfaulting on at least Android inside `libEGL`.  In other words, never dereference opaque handles which may not even be a valid memory address in the first place.

    - [x] Android
    - [x] Wayland
    - [x] X11 (@kchibisov)
    - [x] Windows
- [x] Created or updated an example program if it would help users understand this functionality (N/A)

### Legend

❌ = not supported
✔️ = tested and working
Nothing = untested

||`eglCreateWindowSurface`|`eglCreatePlatformWindowSurface`|`eglCreatePlatformWindowSurfaceEXT`|
|-|-|-|-|
|Wayland|❌|`void *`✔️|`void *`✔️|
|Windows|`HWND`|`HWND`|`HWND`|
|X|`XID`✔️|`XID *`✔️|`XID *`✔️|
|Android|`void *`✔️|`void *`✔️|❌|

||`eglCreatePixmapSurface`|`eglCreatePlatformPixmapSurface`|`eglCreatePlatformPixmapSurfaceEXT`|
|-|-|-|-|
|Windows|`HBITMAP`|`HBITMAP`|`HBITMAP`|
|X|`XID`|`XID *`|`XID *`|

### DIFF
--- a/glutin-winit/build.rs
+++ b/glutin-winit/build.rs
@@ -6,7 +6,6 @@ fn main() {
     // Setup alias to reduce `cfg` boilerplate.
     cfg_aliases! {
         // Systems.
-        android: { target_os = "android" },
         wasm: { target_arch = "wasm32" },
         macos: { target_os = "macos" },
         ios: { target_os = "ios" },
--- a/glutin-winit/build.rs
+++ b/glutin-winit/build.rs
@@ -6,20 +6,21 @@ fn main() {
     // Setup alias to reduce `cfg` boilerplate.
     cfg_aliases! {
         // Systems.
-        wasm: { target_arch = "wasm32" },
-        macos: { target_os = "macos" },
-        ios: { target_os = "ios" },
-        apple: { any(target_os = "ios", target_os = "macos") },
-        free_unix: { all(unix, not(apple), not(android)) },
+        android_platform: { target_os = "android" },
+        wasm_platform: { target_family = "wasm" },
+        macos_platform: { target_os = "macos" },
+        ios_platform: { target_os = "ios" },
+        apple: { any(ios_platform, macos_platform) },
+        free_unix: { all(unix, not(apple), not(android_platform)) },
 
         // Native displays.
-        x11_platform: { all(feature = "x11", free_unix, not(wasm)) },
-        wayland_platform: { all(feature = "wayland", free_unix, not(wasm)) },
+        x11_platform: { all(feature = "x11", free_unix, not(wasm_platform)) },
+        wayland_platform: { all(feature = "wayland", free_unix, not(wasm_platform)) },
 
         // Backends.
-        egl_backend: { all(feature = "egl", any(windows, unix), not(apple), not(wasm)) },
-        glx_backend: { all(feature = "glx", x11_platform, not(wasm)) },
-        wgl_backend: { all(feature = "wgl", windows, not(wasm)) },
-        cgl_backend: { all(macos, not(wasm)) },
+        egl_backend: { all(feature = "egl", any(windows, unix), not(apple), not(wasm_platform)) },
+        glx_backend: { all(feature = "glx", x11_platform, not(wasm_platform)) },
+        wgl_backend: { all(feature = "wgl", windows, not(wasm_platform)) },
+        cgl_backend: { all(macos_platform, not(wasm_platform)) },
     }
 }
--- a/glutin/src/api/egl/surface.rs
+++ b/glutin/src/api/egl/surface.rs
@@ -431,13 +431,13 @@ enum NativeWindow {
     #[cfg(x11_platform)]
     Xcb(u32),
 
-    #[cfg(target_os = "android")]
+    #[cfg(android_platform)]
     Android(*mut ffi::c_void),
 
     #[cfg(windows)]
-    Win32(*mut ffi::c_void),
+    Win32(isize),
 
-    #[cfg(unix)]
+    #[cfg(free_unix)]
     Gbm(*mut ffi::c_void),
 }
 
@@ -466,13 +466,13 @@ impl NativeWindow {
             RawWindowHandle::Xlib(window_handle) => Self::Xlib(window_handle.window as _),
             #[cfg(x11_platform)]
             RawWindowHandle::Xcb(window_handle) => Self::Xcb(window_handle.window as _),
-            #[cfg(target_os = "android")]
+            #[cfg(android_platform)]
             RawWindowHandle::AndroidNdk(window_handle) => {
                 Self::Android(window_handle.a_native_window)
             },
             #[cfg(windows)]
-            RawWindowHandle::Win32(window_hanlde) => Self::Win32(window_hanlde.hwnd),
-            #[cfg(unix)]
+            RawWindowHandle::Win32(window_handle) => Self::Win32(window_handle.hwnd as _),
+            #[cfg(free_unix)]
             RawWindowHandle::Gbm(window_handle) => Self::Gbm(window_handle.gbm_surface),
             _ => {
                 return Err(
@@ -511,10 +511,10 @@ impl NativeWindow {
             #[cfg(x11_platform)]
             Self::Xcb(window_id) => window_id as egl::NativeWindowType,
             #[cfg(windows)]
-            Self::Win32(hwnd) => hwnd as egl::NativeWindowType,
-            #[cfg(target_os = "android")]
+            Self::Win32(hwnd) => hwnd,
+            #[cfg(android_platform)]
             Self::Android(a_native_window) => a_native_window,
-            #[cfg(unix)]
+            #[cfg(free_unix)]
             Self::Gbm(gbm_surface) => gbm_surface,
         }
     }
@@ -541,10 +541,10 @@ impl NativeWindow {
             #[cfg(x11_platform)]
             Self::Xcb(window_id) => window_id as *const _ as *mut ffi::c_void,
             #[cfg(windows)]
-            Self::Win32(hwnd) => *hwnd,
-            #[cfg(target_os = "android")]
+            Self::Win32(hwnd) => *hwnd as *const ffi::c_void as *mut _,
+            #[cfg(android_platform)]
             Self::Android(a_native_window) => *a_native_window,
-            #[cfg(unix)]
+            #[cfg(free_unix)]
             Self::Gbm(gbm_surface) => *gbm_surface,
         }
     }
--- a/glutin/src/api/egl/surface.rs
+++ b/glutin/src/api/egl/surface.rs
@@ -195,11 +195,10 @@ impl Display {
             },
             EglDisplay::Legacy(display) => unsafe {
                 let attrs: Vec<EGLint> = attrs.into_iter().map(|attr| attr as EGLint).collect();
-                // This call accepts raw value, instead of pointer.
                 self.inner.egl.CreateWindowSurface(
                     display,
                     *config.inner.raw,
-                    *(native_window.as_ptr() as *const egl::NativeWindowType),
+                    native_window.as_ptr().cast(),
                     attrs.as_ptr(),
                 )
             },
--- a/glutin/src/api/egl/surface.rs
+++ b/glutin/src/api/egl/surface.rs
@@ -431,10 +431,13 @@ enum NativeWindow {
     #[cfg(x11_platform)]
     Xcb(u32),
 
+    #[cfg(target_os = "android")]
     Android(*mut ffi::c_void),
 
+    #[cfg(windows)]
     Win32(*mut ffi::c_void),
 
+    #[cfg(unix)]
     Gbm(*mut ffi::c_void),
 }
 
@@ -463,10 +466,13 @@ impl NativeWindow {
             RawWindowHandle::Xlib(window_handle) => Self::Xlib(window_handle.window as _),
             #[cfg(x11_platform)]
             RawWindowHandle::Xcb(window_handle) => Self::Xcb(window_handle.window as _),
+            #[cfg(target_os = "android")]
             RawWindowHandle::AndroidNdk(window_handle) => {
                 Self::Android(window_handle.a_native_window)
             },
+            #[cfg(windows)]
             RawWindowHandle::Win32(window_hanlde) => Self::Win32(window_hanlde.hwnd),
+            #[cfg(unix)]
             RawWindowHandle::Gbm(window_handle) => Self::Gbm(window_handle.gbm_surface),
             _ => {
                 return Err(
@@ -503,8 +509,11 @@ impl NativeWindow {
             Self::Xlib(window_id) => window_id as egl::NativeWindowType,
             #[cfg(x11_platform)]
             Self::Xcb(window_id) => window_id as egl::NativeWindowType,
+            #[cfg(windows)]
             Self::Win32(hwnd) => hwnd as egl::NativeWindowType,
+            #[cfg(target_os = "android")]
             Self::Android(a_native_window) => a_native_window,
+            #[cfg(unix)]
             Self::Gbm(gbm_surface) => gbm_surface,
         }
     }
@@ -523,8 +532,11 @@ impl NativeWindow {
             Self::Xlib(window_id) => window_id as *const _ as *mut ffi::c_void,
             #[cfg(x11_platform)]
             Self::Xcb(window_id) => window_id as *const _ as *mut ffi::c_void,
+            #[cfg(windows)]
             Self::Win32(hwnd) => *hwnd,
+            #[cfg(target_os = "android")]
             Self::Android(a_native_window) => *a_native_window,
+            #[cfg(unix)]
             Self::Gbm(gbm_surface) => *gbm_surface,
         }
     }
--- a/glutin/src/api/egl/surface.rs
+++ b/glutin/src/api/egl/surface.rs
@@ -121,7 +121,7 @@ impl Display {
                 self.inner.egl.CreatePixmapSurface(
                     display,
                     *config.inner.raw,
-                    *(native_pixmap.as_ptr() as *const egl::NativePixmapType),
+                    native_pixmap.as_ptr(),
                     attrs.as_ptr(),
                 )
             },
@@ -526,7 +526,7 @@ impl NativePixmap {
         match self {
             Self::XlibPixmap(xid) => xid as *const _ as *mut _,
             Self::XcbPixmap(xid) => xid as *const _ as *mut _,
-            Self::WindowsPixmap(hbitmap) => hbitmap as *const _ as *mut _,
+            Self::WindowsPixmap(hbitmap) => *hbitmap as *const ffi::c_void as *mut _,
         }
     }
 }
--- a/glutin/src/api/egl/surface.rs
+++ b/glutin/src/api/egl/surface.rs
@@ -102,7 +102,7 @@ impl Display {
                 self.inner.egl.CreatePlatformPixmapSurface(
                     display,
                     *config.inner.raw,
-                    native_pixmap.as_ptr(),
+                    native_pixmap.as_platform_pixmap(),
                     attrs.as_ptr(),
                 )
             },
@@ -111,7 +111,7 @@ impl Display {
                 self.inner.egl.CreatePlatformPixmapSurfaceEXT(
                     display,
                     *config.inner.raw,
-                    native_pixmap.as_ptr(),
+                    native_pixmap.as_platform_pixmap(),
                     attrs.as_ptr(),
                 )
             },
@@ -121,7 +121,7 @@ impl Display {
                 self.inner.egl.CreatePixmapSurface(
                     display,
                     *config.inner.raw,
-                    native_pixmap.as_ptr(),
+                    native_pixmap.as_native_pixmap(),
                     attrs.as_ptr(),
                 )
             },
@@ -180,7 +180,7 @@ impl Display {
                 self.inner.egl.CreatePlatformWindowSurface(
                     display,
                     *config.inner.raw,
-                    native_window.as_ptr(),
+                    native_window.as_platform_window(),
                     attrs.as_ptr(),
                 )
             },
@@ -189,7 +189,7 @@ impl Display {
                 self.inner.egl.CreatePlatformWindowSurfaceEXT(
                     display,
                     *config.inner.raw,
-                    native_window.as_ptr(),
+                    native_window.as_platform_window(),
                     attrs.as_ptr(),
                 )
             },
@@ -198,7 +198,7 @@ impl Display {
                 self.inner.egl.CreateWindowSurface(
                     display,
                     *config.inner.raw,
-                    native_window.as_ptr(),
+                    native_window.as_native_window(),
                     attrs.as_ptr(),
                 )
             },
@@ -495,7 +495,27 @@ impl NativeWindow {
         }
     }
 
-    fn as_ptr(&self) -> *mut ffi::c_void {
+    fn as_native_window(&self) -> egl::NativeWindowType {
+        match *self {
+            #[cfg(wayland_platform)]
+            Self::Wayland(wl_egl_surface) => wl_egl_surface,
+            #[cfg(x11_platform)]
+            Self::Xlib(window_id) => window_id as egl::NativeWindowType,
+            #[cfg(x11_platform)]
+            Self::Xcb(window_id) => window_id as egl::NativeWindowType,
+            Self::Win32(hwnd) => hwnd as egl::NativeWindowType,
+            Self::Android(a_native_window) => a_native_window,
+            Self::Gbm(gbm_surface) => gbm_surface,
+        }
+    }
+
+    // NOTE: on X11 and Windows the native window are XID and HWND
+    // respectively, however the platform window should be behind a
+    // pointer. The function accepts the `&self`, so it automatically
+    // borrows the value and then we simply cast the pointer. On other
+    // platforms though, the default window handle is a pointer already,
+    // so we simply dereference the just taken borrow.
+    fn as_platform_window(&self) -> *mut ffi::c_void {
         match self {
             #[cfg(wayland_platform)]
             Self::Wayland(wl_egl_surface) => *wl_egl_surface,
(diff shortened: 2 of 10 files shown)


kchibisov:
This is actually wrong, because the function accepts `NativeWindowType`. If you look into the code we have the value is behind double pointer and then we deref it so we pass the `* void`.

For example on X11 the default type passed to the client is `ulong`, so we dereference it and pass the `ulong`.

See https://github.com/rust-windowing/glutin/issues/1566 . It was a bug before due to me blindly following the KHR extension, however the dereference should be present.

Maybe some logic is not really sane here... Would suggest to consult the EGL spec wrt android here...

MarijnS95:
> `egl::NativeWindowType` (which is a `*const c_void` under the hood).

Ouch, on Windows it's `isize`, probably because of being `HWND`:

https://github.com/rust-windowing/glutin/actions/runs/4583750441/jobs/8094795641#step:7:72

@kchibisov perhaps all backends should use `EGLNativeWindowType`, as it should not always be interpreted as a void pointer?

MarijnS95:
> If you look into the code we have the value is behind double pointer

`NativeWindow::as_ptr()` is not?

kchibisov:
On the other hand the KHR function accepts the `*NativeWindowTYpe`, so it must be a double pointer on Android/Wayland, and a single pointer on X11.

So if you look into both functions the signatures are different.

https://registry.khronos.org/EGL/sdk/docs/man/html/eglCreateWindowSurface.xhtml
https://registry.khronos.org/EGL/sdk/docs/man/html/eglCreatePlatformWindowSurface.xhtml

MarijnS95:
Oh I see what stupidity #1570 is doing. `as_ptr()` takes `&self` and you cast the auto-borrowed `window_id` to a raw pointer, whereas on wayland/windows/android/gbm this auto-borrow is first dereferenced :facepalm:

kchibisov:
>Oh I see what stupidity #1570 is doing. as_ptr() takes &self and you cast the auto-borrowed window_id to a raw pointer, whereas on wayland/windows/android/gbm this auto-borrow is first dereferenced facepalm

So it seems like I was also insane when I wrote this, because it's always a single pointer in the end. It just some non pointer types are moving to pointer types. I'd suggest to rewrite it in a more sane way...

MarijnS95:
> So it seems like I was also insane when I wrote this, because it's always a single pointer in the end. It just some non pointer types are moving to pointer types(which they clearly should). I'd suggest to rewrite it in a more sane way...

Yeah I think the fix shown there wouldn't have worked on Xlib/Xcb + the KHR/EXT extension (if that's even possible) because we'd be passing a pointer to `window_id` :)

MarijnS95:
@kchibisov so there are a few questions to answer, because we can write this "sanely" in a few different ways:

1. Always define `EGLNativeWindowType` to be `*mut c_void`;
2. Keep `EGLNativeWindowType` platform-specific, but more rigorously `cfg()` the entries in `NativeWindow` (so that we don't have to cast a `HWND` when not compiling for Windows, that match arm won't be reached anyway);

MarijnS95:
> So if you look into both functions the signatures are different.
> 
> https://registry.khronos.org/EGL/sdk/docs/man/html/eglCreateWindowSurface.xhtml https://registry.khronos.org/EGL/sdk/docs/man/html/eglCreatePlatformWindowSurface.xhtml

Not sure if you noticed already, but both functions are identical afaik. One takes `NativeWindowType` (typedef'd to `void *`) and the other takes `void *` directly without a typedef.

MarijnS95:
> No, way you did won't work on X11, because you need actually what it was doing before with pointers. Because the native window type was not the pointer in the first place actually.

Oh that's very unfortunate... Yikes. Then the original push was good afaik, after fixing casting issues on Windows?

kchibisov:
On X11 `NativeWindowType` is a XID though? I remember when I've passed it directly it wasn't working and if you look at what I pass now it's a `& NativeWindowType` on X11, because I don't deference.

kchibisov:
I can submit a patch to fix this issue if you don't mind.

MarijnS95:
> I can submit a patch to fix this issue if you don't mind.

No need for that, I've reverted the change and now made the output of `as_ptr()` consistent across the platforms.

That is, on Windows `EGLNativeWindowType = HWND = *const c_void`.

kchibisov:
Hm, but the thing is that the NativeWindow is different.

You could try something like that.

```diff
diff --git a/glutin/src/api/egl/surface.rs b/glutin/src/api/egl/surface.rs
index fc396c2..a585732 100644
--- a/glutin/src/api/egl/surface.rs
+++ b/glutin/src/api/egl/surface.rs
@@ -102,7 +102,7 @@ impl Display {
                 self.inner.egl.CreatePlatformPixmapSurface(
                     display,
                     *config.inner.raw,
-                    native_pixmap.as_ptr(),
+                    native_pixmap.as_platfrom_pixmap(),
                     attrs.as_ptr(),
                 )
             },
@@ -111,7 +111,7 @@ impl Display {
                 self.inner.egl.CreatePlatformPixmapSurfaceEXT(
                     display,
                     *config.inner.raw,
-                    native_pixmap.as_ptr(),
+                    native_pixmap.as_platfrom_pixmap(),
                     attrs.as_ptr(),
                 )
             },
@@ -121,7 +121,7 @@ impl Display {
                 self.inner.egl.CreatePixmapSurface(
                     display,
                     *config.inner.raw,
-                    *(native_pixmap.as_ptr() as *const egl::NativePixmapType),
+                    native_pixmap.as_native_pixmap(),
                     attrs.as_ptr(),
                 )
             },
@@ -180,7 +180,7 @@ impl Display {
                 self.inner.egl.CreatePlatformWindowSurface(
                     display,
                     *config.inner.raw,
-                    native_window.as_ptr().cast(),
+                    native_window.as_platform_window(),
                     attrs.as_ptr(),
                 )
             },
@@ -189,7 +189,7 @@ impl Display {
                 self.inner.egl.CreatePlatformWindowSurfaceEXT(
                     display,
                     *config.inner.raw,
-                    native_window.as_ptr().cast(),
+                    native_window.as_platform_window(),
                     attrs.as_ptr(),
                 )
             },
@@ -199,7 +199,7 @@ impl Display {
                 self.inner.egl.CreateWindowSurface(
                     display,
                     *config.inner.raw,
-                    *(native_window.as_ptr() as *const egl::NativeWindowType),
+                    native_window.as_native_window(),
                     attrs.as_ptr(),
                 )
             },
@@ -496,7 +496,27 @@ impl NativeWindow {
         }
     }
 
-    fn as_ptr(&self) -> *mut ffi::c_void {
+    fn as_native_window(&self) -> egl::NativeWindowType {
+        match *self {
+            #[cfg(wayland_platform)]
+            Self::Wayland(wl_egl_surface) => wl_egl_surface,
+            #[cfg(x11_platform)]
+            Self::Xlib(window_id) => window_id as egl::NativeWindowType,
+            #[cfg(x11_platform)]
+            Self::Xcb(window_id) => window_id as egl::NativeWindowType,
+            Self::Win32(hwnd) => hwnd as egl::NativeWindowType,
+            Self::Android(a_native_window) => a_native_window,
+            Self::Gbm(gbm_surface) => gbm_surface,
+        }
+    }
+
+    // NOTE: on X11 and Windows the native window are XID and HWND
+    // respectively, however the platform window should be behind a
+    // pointer. The function accepts the `&self`, so it automatically
+    // borrows the value and then we simply cast the pointer. On other
+    // platforms though, the default window handle is a pointer already,
+    // so we simply derefence the just taken borrow.
+    fn as_platform_window(&self) -> *mut ffi::c_void {
         match self {
             #[cfg(wayland_platform)]
             Self::Wayland(wl_egl_surface) => *wl_egl_surface,
@@ -523,7 +543,17 @@ impl Drop for NativeWindow {
 }
 
 impl NativePixmap {
-    fn as_ptr(&self) -> *mut ffi::c_void {
+    fn as_native_pixmap(&self) -> egl::NativePixmapType {
+        match *self {
+            Self::XlibPixmap(xid) => xid as egl::NativePixmapType,
+            Self::XcbPixmap(xid) => xid as egl::NativePixmapType,
+            Self::WindowsPixmap(hbitmap) => hbitmap as egl::NativePixmapType,
+        }
+    }
+
+    // NOTE: see the `NativeWindow::as_platform_window`. The same applies to this
+    // function.
+    fn as_platfrom_pixmap(&self) -> *mut ffi::c_void {
         match self {
             Self::XlibPixmap(xid) => xid as *const _ as *mut _,
             Self::XcbPixmap(xid) => xid as *const _ as *mut _,

```

MarijnS95:
@kchibisov Oh right, we should solve the same dereference for pixmaps then, since `as_ptr` now returns a pointer to `xid` or _a pointer to `HBITMAP`_, and derferences it for legacy display only.

In other words, that code would then only work for Xlib/Xcb on KHR/EXT, and only for legacy on Windows?

---

To make your patch work we'd have to also `cfg(windows)` the `Win32` arm, because it has to cast to `HWND = isize`, as pointed out by an earlier CI failure. Is that okay by you or do you rather keep the current approach?

MarijnS95:
Oh, reading back up on https://github.com/rust-windowing/glutin/issues/1566#issuecomment-1424652584, it seems that there _is_ a syntactical difference between the pointer and XID, specifically in the legacy `eglCreateWindowSurface`, but only on X for `window_id`/`XID`?

Did I grasp this correctly? Substitute for pixmap where appropriate:

||`eglCreateWindowSurface`|`eglCreatePlatformWindowSurfaceKHR`|`eglCreatePlatformWindowSurfaceEXT`|
|-|-|-|-|
|Windows|`HWND`|`HWND`|`HWND`|
|X|`XID`|`*XID`|`*XID`|
|Android|`void *`|`void *`|`void *`|

kchibisov:
Shouldn't it be `HWND *` for windows with `eglCreatePlatformWindowSurfaceKHR` and `HWND` with normal window? I don't have EGL on windows, so can't test for segfault :p

MarijnS95:
> Shouldn't it be `HWND *` for windows with `eglCreatePlatformWindowSurfaceKHR` and `HWND` with normal window? I don't have EGL on windows, so can't test for segfault :p

I'd have to reboot to Windows to test out this theory, but it's currently returning `HWND` (see the dereference in `master`) but for `HBITMAP` it returns a pointer to it :)

On that note, the public API takes `isize` for `HBITMAP` but `*mut c_void` for `HWND`...

kchibisov:
I think HWND was always an `*mut c_void` with the WinAPI crate, then when we've migrated to `windows-sys` it became `isize` instead, so in my understanding it was the same as with `XID`.

In general, both `HWND` and `HBITMAP` should be the same definition from what I understand, but I have no clue how to test what windows really wants from me here. For X11 I simply read the mesa code :/

MarijnS95:
@kchibisov Hmm right, no I think despite the `isize` in `windows-sys` it should still be passed as-is, but let's test.

For now this change is spiraling out of control into a rabbit-hole of changes. What should we tackle first, and the rest in what order? I can reboot to windows and test all three functions (assuming the extensions are available), and cook up a pixmap example to do the same.

Then, for Android as #1584 pointed out it must be the `ANativeWindow` pointer in all three cases, no changes. For wayland this should be the same.

Then if you can test that XID/window_id must only be a pointer on KHR/EXT I think we're good.

Finally, we have the type changes. I can use `egl::NativeWindow` but it must return a `*const c_void` as `&self` is not borrowed mutable - unless we want to do a const-cast. Then for windows, should we change the public `NativeWindow` to `isize` to match `NativePixmap` and `windows-sys`? We could even reuse the types from `windows-sys` which are currenltly only used in the WGL code... and `egl_sys` :)

kchibisov:
I've verified X11 with my patch(and everything works fine). So only the `EGL` part of windows was never tested, because simply my driver doesn't provide the `EGL` libraries. Be aware that you might end up with the same situation.

To test this on Windows you'll need to

1. Change the `glutin-winit` to prefer `EGL` and ensure that you're using EGL on Windows(my hardware can't do that, so I can't test).
2. Simply use the current patch semantics wrt what is pointer and what is not. From my understanding you should pass `&HWND` for `Platform` and `HWND` for native.

If any of that crashes adjust the semantics to not do so(maybe you'll need passing the thing as is). From my understanding though, the `HWND` is not a real pointer, given that it has arbitrary data encoded into it.

>We could even reuse the types from windows-sys which are currenltly only used in the WGL code... and egl_sys :)

We probably could, but it's a one more deps and a breaking change I'm not into for now...

MarijnS95:
> I've verified X11 with my patch(and everything works fine). So only the `EGL` part of windows was never tested, because simply my driver doesn't provide the `EGL` libraries. Be aware that you might end up with the same situation.

Okay, I'll apply your patch on top to make sure X11 starts working again. Did you test all of `Legacy`/`KHR`/`EXT` (by changing the logic in `Display::new`)?

> From my understanding though, the `HWND` is not a real pointer, given that it has arbitrary data encoded into it.

Heh, and so we're back at the original "don't dereference whatever comes out of `fn as_ptr()`".
 
> > We could even reuse the types from windows-sys which are currently only used in the WGL code... and egl_sys :)
> 
> We probably could, but it's a one more deps and a breaking change I'm not into for now...

Even without the dep, changing from `*mut c_void` to `isize` would also be breaking.

glutin_egl_sys/src/lib.rs
```diff
@@ -46,16 +46,4 @@ pub type EGLenum = raw::c_uint;
...
-pub type EGLNativeWindowType = *const raw::c_void;
-#[cfg(target_os = "android")]
-pub type EGLNativeWindowType = *const raw::c_void;
-#[cfg(any(
-    target_os = "dragonfly",
-    target_os = "freebsd",
-    target_os = "netbsd",
-    target_os = "openbsd"
-))]
 pub type EGLNativeWindowType = *const raw::c_void;
```
MarijnS95, line 49: We can probably revert this now, and simplify for `cfg(windows)` and `cfg(not(windows))`. Then, do the same for `EGLNativePixmapType` on `HBITMAP`?

kchibisov: I'd suggest to keep the crate types as they were before, at least unless we figure out the semantics.

kchibisov:
>Okay, I'll apply your patch on top to make sure X11 starts working again. Did you test all of Legacy/KHR/EXT (by changing the logic in Display::new)?

Yes, that's what I do. You can simply `Ok(::new.unwrap())` to ensure that you're using the right thing.

glutin/src/api/egl/surface.rs
```diff
@@ -431,10 +431,13 @@ enum NativeWindow {
     #[cfg(x11_platform)]
     Xcb(u32),
 
+    #[cfg(android)]
```
MarijnS95, line 434: > `operating system used in target family position`

Looks like the `android` CFG alias isn't working.

glutin_egl_sys/src/lib.rs
MarijnS95: Done, but as said that causes `as_native_window` to return an `isize` (typedef for `egl::NativeWindowType`) on Windows and `*mut c_void` everywhere else so I had to `cfg`-guard all other `NativeWindow` arms.

MarijnS95:
Unfortunately my Windows system doesn't have `libEGL.so` (AMD RX6800 XT) and I don't know of a sensible way to acquire it.

Regarding Mesa, perhaps we could peek through the EGL-over-WGL backend: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/12727?commit_id=3415bf0268884bf3c16d495a4e1afd30d67bc71d? Not sure if the native window type for EGL is still a `HWND` or something opaque Linux'y within WSL.

kchibisov:
>Regarding Mesa, perhaps we could peek through the EGL-over-WGL backend: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/12727?commit_id=3415bf0268884bf3c16d495a4e1afd30d67bc71d? Not sure if the native window type for EGL is still a HWND or something opaque Linux'y within WSL.

I've read through glfw code, and it's passed as is. Only on X11 you have the pointer trick thing I was doing.

glutin_egl_sys/src/lib.rs
kchibisov: Let's define it as pointer then, the crate is intended for glutin use only.

glutin/src/api/egl/surface.rs
kchibisov: While we're at it, I'd suggest to change the names, like `android_platform`.  Or similar to what we have in https://github.com/rust-windowing/winit/blob/fbea75d31f4025825e9bf88b99d7c92e4eb79273/build.rs#L35 given that winit has it right.

glutin-winit/build.rs
```diff
@@ -6,7 +6,6 @@ fn main() {
     // Setup alias to reduce `cfg` boilerplate.
     cfg_aliases! {
         // Systems.
-        android: { target_os = "android" },
         wasm: { target_arch = "wasm32" },
```
kchibisov, line 9: Use `android_platform` instead of just `android`, then it'll work. Same rename change should be done to the `macos`,`ios`, and `wasm`.

glutin/src/surface.rs
```diff
@@ -485,6 +485,8 @@ pub enum NativePixmap {
     XcbPixmap(u32),
 
     /// HBITMAP handle for windows bitmap.
+    // TODO: Consistency between HBITMAP and HWND should be solved in the
+    // next breaking release. Either both are `*mut c_void`, or `isize`.
```
kchibisov, line 489: I picked isize, because `hwnd`(inside the windows-sys) is `isize`. And given that we're using `isize` there's no inconsistency.

glutin/src/api/egl/surface.rs
```diff
@@ -432,10 +431,13 @@ enum NativeWindow {
     #[cfg(x11_platform)]
     Xcb(u32),
 
+    #[cfg(target_os = "android")]
     Android(*mut ffi::c_void),
 
+    #[cfg(windows)]
     Win32(*mut ffi::c_void),
 
+    #[cfg(unix)]
```
kchibisov, line 440: that's `free_unix`, macOS doesn't have gbm.

kchibisov (changes requested):

glutin/src/surface.rs
MarijnS95: Then I'll move this comment to `NativeWindow::Win32`, which uses `*mut c_void` while `windows-sys` says it should be an `isize`.

glutin-winit/build.rs
MarijnS95: Yeah, you said that yesterday in https://github.com/rust-windowing/glutin/pull/1585#discussion_r1155173734.

glutin/src/surface.rs
MarijnS95: In fact `NativeWindow` is private to `egl` so I've updated the type instead.

glutin/src/api/egl/surface.rs
MarijnS95: @kchibisov fwiw `winit` defines `wasm_platform` but the macros below use `not(wasm)` instead of `not(wasm_platform)`. Is there an undocumented `wasm` for this `target_family` [just like there's `unix` and `windows`](https://doc.rust-lang.org/reference/conditional-compilation.html#unix-and-windows)?

glutin/src/api/egl/surface.rs
```diff
@@ -523,11 +562,33 @@ impl Drop for NativeWindow {
 }
 
 impl NativePixmap {
-    fn as_ptr(&self) -> *mut ffi::c_void {
+    /// Returns the underlying handle value
+    fn as_native_pixmap(&self) -> egl::NativePixmapType {
+        match *self {
+            Self::XlibPixmap(xid) => xid as egl::NativePixmapType,
+            Self::XcbPixmap(xid) => xid as egl::NativePixmapType,
+            Self::WindowsPixmap(hbitmap) => hbitmap as egl::NativePixmapType,
```
MarijnS95, line 570: @kchibisov fwiw we can skip this cast if we similarly `#[cfg(windows)]` `WindowsPixmap`, because both `egl::NativePixmapType` and `HBITMAP` are `isize` there. Should I?

MarijnS95: Ah nevermind again, I keep mixing things up as `NativeWindow` (re-wrapper for `RawWindowHandle`) is private to `egl` but `NativePixmap` is global and public outside of `egl`.

glutin/src/api/egl/surface.rs
```diff
@@ -496,16 +501,50 @@ impl NativeWindow {
         }
     }
 
-    fn as_ptr(&self) -> *mut ffi::c_void {
+    /// Returns the underlying handle value
```
kchibisov, line 504:
```suggestion
    /// Returns the underlying handle value.
```

glutin/src/api/egl/surface.rs
```diff
@@ -496,16 +501,50 @@ impl NativeWindow {
...
+    /// This exists because of a discrepancy in the new
+    /// `eglCreatePlatformWindowSurface*` functions which take a pointer to the
+    /// `window_id` on X11 and Xlib, in contrast to the legacy
+    /// `eglCreateWindowSurface` which always takes the raw value.
+    ///
+    /// See also:
+    /// <https://gitlab.freedesktop.org/mesa/mesa/-/blob/4de9a4b2b8c41864aadae89be705ef125a745a0a/src/egl/main/eglapi.c#L1102-1127>
+    ///
+    /// # Safety:
+    /// The returned pointer is a cast of the `&self` borrow.
```
kchibisov, line 534:
```suggestion
    /// # Safety:
    ///
    /// On X11 the returned pointer is a cast of the `&self` borrow.
```

glutin/src/api/egl/surface.rs
```diff
@@ -496,16 +501,50 @@ impl NativeWindow {
...
+            Self::Win32(hwnd) => hwnd,
+            #[cfg(android_platform)]
+            Self::Android(a_native_window) => a_native_window,
+            #[cfg(free_unix)]
+            Self::Gbm(gbm_surface) => gbm_surface,
+        }
+    }
+
+    /// Returns a pointer to the underlying handle value on X11 and Xlib,
+    /// the raw underlying handle value on all other platforms.
```
kchibisov, line 523:
```suggestion
    /// Returns a pointer to the underlying handle value on X11,
    /// the raw underlying handle value on all other platforms.
```

glutin/src/api/egl/surface.rs
```diff
@@ -523,11 +562,33 @@ impl Drop for NativeWindow {
 }
 
 impl NativePixmap {
-    fn as_ptr(&self) -> *mut ffi::c_void {
+    /// Returns the underlying handle value
```
kchibisov, line 565:
```suggestion
    /// Returns the underlying handle value.
```

glutin/src/api/egl/surface.rs
```diff
@@ -523,11 +562,33 @@ impl Drop for NativeWindow {
...
+    fn as_native_pixmap(&self) -> egl::NativePixmapType {
+        match *self {
+            Self::XlibPixmap(xid) => xid as egl::NativePixmapType,
+            Self::XcbPixmap(xid) => xid as egl::NativePixmapType,
+            Self::WindowsPixmap(hbitmap) => hbitmap as egl::NativePixmapType,
+        }
+    }
+
+    /// Returns a pointer to the underlying handle value on X11 and Xlib,
+    /// the raw underlying handle value on all other platforms.
```
kchibisov, line 575:
```suggestion
    /// Returns a pointer to the underlying handle value on X11,
    /// the raw underlying handle value on all other platforms.
```

glutin/src/api/egl/surface.rs
```diff
@@ -523,11 +562,33 @@ impl Drop for NativeWindow {
...
+    /// This exists because of a discrepancy in the new
+    /// `eglCreatePlatformPixmapSurface*` functions which take a pointer to the
+    /// `xid` on X11 and Xlib, in contrast to the legacy
+    /// `eglCreatePixmapSurface` which always takes the raw value.
+    ///
+    /// See also:
+    /// <https://gitlab.freedesktop.org/mesa/mesa/-/blob/4de9a4b2b8c41864aadae89be705ef125a745a0a/src/egl/main/eglapi.c#L1166-1190>
+    ///
+    /// # Safety:
+    /// The returned pointer is a cast of the `&self` borrow.
```
kchibisov, line 586:
```suggestion
    /// # Safety:
    ///
    /// On X11 the returned pointer is a cast of the `&self` borrow.
```

kchibisov (changes requested):
Will also need a `CHANGELOG` entry for that.

kchibisov (approved):

PR #1590, opened 2023-04-10
Add dependabot support

notgull:
Adds support for dependabot in order to keep packages up to date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: weekly
+    commit-message:
+      prefix: ''
+    labels: []
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,7 @@ updates:
     directory: /
     schedule:
       interval: weekly
-    commit-message:
-      prefix: ''
-    labels: []
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
```
MarijnS95, line 3: Don't forget to also add github-actions.

.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
...
+    directory: /
+    schedule:
+      interval: weekly
+    commit-message:
+      prefix: ''
+    labels: []
```
MarijnS95, line 9: These seem to be the default and do not add value? Or is there special handling for `prefix: ''`? (I vaguely recall discussing this months ago)

notgull: Looks like it; I've removed them accordingly.

kchibisov:
I really don't like the dependabot, and given that  we have very limited amount of dependencies and we don't have any `lock` file we have nothing to update other than the `windows-sys`, and some macOS crates, both should be kept in sync with the winit anyway, so dependabot is sort of useless.

Given that we are a library the rest is bumped automatically.

notgull:
> Given that we are a library the rest is bumped automatically.

What if I configured Dependabot to ignore minor/patch version bumps and to only make PRs for major version bumps?

kchibisov:
Yeah, but it would be even better if it could list stuff in maintainers tab, and not even open more noise. Like they already have security alerts there from dependabot, could just add an extra to list the rest.

Though, issue is a way better, than stupid PR I'll close anyway.

MarijnS95:
@kchibisov what do you think about revisiting this? It could remind us about for example the `libloading 0.9` bump, and various stale GitHub actions bumps. I have quite a nice terse setup to PR.

kchibisov:
@MarijnS95 given that I don't have time to actively check anything for the time being, I think it's ok, given that glutin is low volume probably won't get lost.

PR #1593, opened 2023-05-07
docs: add notes about Wayland specific behavior

### LINKED ISSUE #1589: Document Wayland specifics on `make_current` and so on
Wayland with EGL have a notion of latching context, if you're not familiar with it, it
means that your resize, will apply only from the next frame. This is not intuitive
for anyone, and it could take folks time to understand what is going wrong with the
staff they are doing.

Such behavior should be explicitly documented. For details see the `EGL_KHR_platform_wayland`
on the khronos github page.

kchibisov:
The wayland platform has a back buffer latching behavior, some of it  is clear, like rendering to and querying `buffer_age`, such is specified in EGL_KHR_platform_wayland.

However some drivers(mesa) perform latching when creating or making context current, thus it's also documented for now.

In addition, the absence of traditional VSync was also documented.

Fixes #1589.

### DIFF
--- a/glutin/src/context.rs
+++ b/glutin/src/context.rs
@@ -56,6 +56,9 @@ pub trait NotCurrentGlContext: Sealed {
     /// # Platform specific
     ///
     /// **macOS:** - **This will block if your main thread is blocked.**
+    /// **Wayland:** - This call may latch the underlying back buffer (will do
+    /// with mesa drivers), meaning that all resize operations will apply after
+    /// the next [`GlSurface::swap_buffers`].
     fn make_current<T: SurfaceTypeTrait>(
         self,
         surface: &Self::Surface<T>,
--- a/glutin/src/display.rs
+++ b/glutin/src/display.rs
@@ -61,7 +61,16 @@ pub trait GlDisplay: Sealed {
     /// Some platforms use [`RawWindowHandle`] for context creation, so it must
     /// point to a valid object.
     ///
+    /// # Platform-specific
+    ///
+    /// **Wayland:** - This call may latch the underlying back buffer of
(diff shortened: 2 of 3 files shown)


kchibisov:
@chrisduerr given that you've faced this issue(with latching) in the wild, does these docs make it a bit clear?

glutin/src/context.rs
```diff
@@ -56,6 +56,9 @@ pub trait NotCurrentGlContext: Sealed {
     /// # Platform specific
     ///
     /// **macOS:** - **This will block if your main thread is blocked.**
+    /// **Wayland:** - This call may latch(will do with mesa drivers) the
+    /// underlying back buffer, meaning that all resize operations will
+    /// apply after the next [`GlSurface::swap_buffers`].
```
chrisduerr, line 61:
```suggestion
    /// **Wayland:** - This call may latch the underlying back buffer (will do with mesa drivers), meaning that all resize operations will apply after the next [`GlSurface::swap_buffers`].
```

I think switching it around like this is a little better. It doesn't interrupt the reading process as much because the parenthesis are right in front of a comma anyway. Same applies to the other doc comments of this structure.

chrisduerr (commented):
I think the documentation is fine, but I don't see many people reading it and immediately realizing the issue.

Chances are they run into this issue and with a bit of luck they might stumble across this comment to help them out. Which I think is better than nothing for sure.

kchibisov:
Part of that was to mention that swap buffers could block indefinitely on Wayland as well.

But at least it's stated in the docs, so it won't be a surprise for future readers...

chrisduerr (approved):

PR #1598, opened 2023-05-13
config: use platform specific default for `Api`

### LINKED ISSUE #1586: Android (egl) example picks wrong gl version
The crash of #1584 is resolved on master, but the template builder still only receives a config with GLES1 instead of the available 3.2 (as the official Java example does), causing no functionality to be available, just keeps repeating:
```
libEGL  : called unimplemented OpenGL ES API
```
It seems this limited config is picked because glutin-winit/src/lib.rs build() function does not initialize a raw window handle for egl.
Then in api/egl/display.rs new() the comment says it needs raw_display or EGL_DEFAULT_DISPLAY is used which is not recommended.

kchibisov:
Avoid pure ES1 configs with EGL by default. The old default behavior could be restored by using the `Api::empty`.

Fixes #1586.

### DIFF
--- a/glutin/src/config.rs
+++ b/glutin/src/config.rs
@@ -164,9 +164,17 @@ impl ConfigTemplateBuilder {
         self
     }
 
-    /// The set of apis that are supported by configuration.
+    /// The set of apis that are supported by this configuration.
     ///
-    /// By default api isn't specified when requesting the configuration.
+    /// The default [`Api`] depends on the used graphics platform interface. If
+    /// you want to do config filtering based on the [`Api`] yourself, use
+    /// [`Api::empty`].
+    ///
+    /// # Api-specific
+    ///
+    /// - **EGL:** [`Api::GLES2`] bit is set by default to avoid matching
+    ///   [`Api::GLES1`] configs;
+    /// - **GLX/WGL/CGL:** [`Api::OPENGL`] is always present in the result.
     #[inline]
     pub fn with_api(mut self, api: Api) -> Self {
         self.template.api = Some(api);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Fixed EGL dereferencing raw window handles on everything but X11 in legacy `Window` and `Pixmap` surface creation.
 - On GLX, fixed startup failure when passing default `Flush` with `KHR_context_flush_control`.
+- **Behavior change:** the `Config`'s `Api` now has platform specific default instead of being unspecified.
 
 # Version 0.30.7
 
--- a/glutin/src/api/egl/config.rs
+++ b/glutin/src/api/egl/config.rs
@@ -125,7 +125,8 @@ impl Display {
             config_attributes.push(num_samples as EGLint);
         }
 
-        if let Some(requested_api) = template.api {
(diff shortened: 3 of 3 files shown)


glutin/src/config.rs
```diff
@@ -166,7 +166,14 @@ impl ConfigTemplateBuilder {
 
     /// The set of apis that are supported by configuration.
     ///
-    /// By default api isn't specified when requesting the configuration.
+    /// The default is specific based on the used API. If you
+    /// want to do config filterig based in [`Api`] yourself use [`Api::empty`].
```
MarijnS95, line 170:
```suggestion
    /// want to do config filtering based on the [`Api`] yourself, use [`Api::empty`].
```

glutin/src/config.rs
```diff
@@ -166,7 +166,14 @@ impl ConfigTemplateBuilder {
 
     /// The set of apis that are supported by configuration.
     ///
-    /// By default api isn't specified when requesting the configuration.
+    /// The default is specific based on the used API. If you
```
MarijnS95, line 169: Not quite sure how to word this because we have the `Api` flag here which is different from the EGL/GLX/WGL/CGL "API", but they are named the same... Wikipedia calls EGL a "native graphics platform interface".

```suggestion
    /// The default [`Api`] depends on the used graphics platform interface. If you
```

But not really liking that yet.

glutin/src/config.rs
```diff
@@ -166,7 +166,14 @@ impl ConfigTemplateBuilder {
...
+    /// want to do config filtering based on the [`Api`] yourself, use [`Api::empty`].
+    ///
+    /// # Api-specific
```
MarijnS95, line 172:
```suggestion
    /// # Api-specific defaults
```

kchibisov: it's not only about defaults though, we just have a `Api-specific` section which is common with the rest.

As you can see the `GLX` line says that you always have OPENGL regardless of what you do.

glutin/src/config.rs
```diff
@@ -164,9 +164,16 @@ impl ConfigTemplateBuilder {
...
+    ///
+    /// **EGL:** - [`Api::GLES2`] bit is set by default to avoid matching
+    /// [`Api::GLES1`] configs.
+    /// **GLX/WGL/CGL:** - [`Api::OPENGL`] is always present in the result.
```
MarijnS95, line 176: Does rustdoc render this nicely btw? I have a hunch the `-` has to be at the beginning to get an _Unordered List_ otherwise this just becomes one line/paragraph?

glutin/src/config.rs
kchibisov: I think it's fine, at least it's better than what we had.

glutin/src/config.rs
kchibisov: You're right, there should be `-` to form a list.

MarijnS95: image
Yup, it doesn't like it:

```suggestion
    /// - **EGL:** [`Api::GLES2`] bit is set by default to avoid matching
    /// [`Api::GLES1`] configs.
    /// - **GLX/WGL/CGL:** [`Api::OPENGL`] is always present in the result.
```

MarijnS95 (approved):
Looks and sounds good, after fixing the unordered-list formatting 🎉

PR #1601, opened 2023-05-30
Fix lock on glxSwapBuffers with some drivers

kchibisov:
The lock handling is done internally, so no need to put extra synchronization which confuses mesa.

### DIFF
--- a/glutin/src/api/glx/mod.rs
+++ b/glutin/src/api/glx/mod.rs
@@ -21,8 +21,6 @@ pub mod context;
 pub mod display;
 pub mod surface;
 
-use display::GlxDisplay;
-
 /// When using Xlib we need to get errors from it somehow, however creating
 /// inner `XDisplay` to handle that or change the error hook is unsafe in
 /// multithreaded applications, given that error hook is per process and not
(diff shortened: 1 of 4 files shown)


kchibisov:
r @chrisduerr

chrisduerr (approved):
Works on my machine.

glutin/src/api/glx/mod.rs
```diff
@@ -175,16 +173,17 @@ fn glx_error_hook(_display: *mut ffi::c_void, xerror_event: *mut ffi::c_void) ->
     }
 }
 
+/// Prevent error being overwritten when accessing the handler from the multiple
+/// threads.
```
MarijnS95, line 177:
```suggestion
/// Prevent GL error from being overwritten when accessing the handler from multiple
/// threads.
```

kchibisov: it's not a GL error though.

MarijnS95: Can you explain what "the multiple threads" is though?

kchibisov: If you call swap buffers from 2 threads at the same time, and the error is generated by the Xlib or glx, it could overwrite the error each thread will get.

kchibisov: I haven't seen that you've commented when merging btw, it appeared after the fact.

kchibisov: But `the` before plural is fine, no?

MarijnS95: Directly before, so `the threads`, should be fine. I'm no language expert but `the multiple threads` reads and feels wrong, though I cannot tell you what exact grammar rule this is.

kchibisov: From what I can see there's nothing wrong with saying this, it's just puts different emphasis on the sentence.

MarijnS95: @kchibisov I don't think so because of the double `the`: `the handler from the multiple threads` tries to put emphasis in two places.

PR #1602, opened 2023-05-30
docs: use lists for Platform and Api sections

kchibisov:


glutin/src/context.rs
```diff
@@ -55,10 +55,10 @@ pub trait NotCurrentGlContext: Sealed {
     ///
     /// # Platform specific
     ///
-    /// **macOS:** - **This will block if your main thread is blocked.**
-    /// **Wayland:** - This call may latch the underlying back buffer (will do
-    /// with mesa drivers), meaning that all resize operations will apply after
-    /// the next [`GlSurface::swap_buffers`].
+    /// - **macOS:** - **This will block if your main thread is blocked.**
+    /// - **Wayland:** - This call may latch the underlying back buffer (will do
```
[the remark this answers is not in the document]
kchibisov: I'm not sure what you mean here?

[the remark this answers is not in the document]
MarijnS95: Doesn't it become clear from the suggested change?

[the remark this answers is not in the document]
kchibisov: I think I just used the style we had in winit. In general lists should be ` - ... ;` and the last entry with ` - ... .`. Maybe we should change to it instead?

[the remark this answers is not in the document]
MarijnS95: Winit has the requested `- ...: ...` format, not the `- ...: - ...` introduced by this PR.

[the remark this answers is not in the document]
MarijnS95: I agree that `;` for all entries and `.` for the last one is the way to go, but that's a different issue from what I'm bringing up here.

glutin/src/context.rs
```diff
@@ -55,10 +55,10 @@ pub trait NotCurrentGlContext: Sealed {
     ///
     /// # Platform specific
     ///
-    /// **macOS:** - **This will block if your main thread is blocked.**
-    /// **Wayland:** - This call may latch the underlying back buffer (will do
-    /// with mesa drivers), meaning that all resize operations will apply after
-    /// the next [`GlSurface::swap_buffers`].
+    /// - **macOS: this will block if your main thread is blocked;**
```
MarijnS95, line 58: Nit: this feels more natural (same below) but not sure if it's worth it.

```suggestion
    /// - **macOS: this will block if your main thread is blocked**;
```

glutin/src/context.rs
```diff
@@ -206,8 +206,8 @@ impl ContextAttributesBuilder {
     ///
     /// # Api-specific
     ///
-    /// **WGL:** - you must pass `raw_window_handle` for if you plan to use it
-    /// with window.
+    /// - **WGL:** you **must** pass `raw_window_handle` for if you plan to use
+    ///   it with the window.
```
MarijnS95, line 210:
```suggestion
    /// - **WGL:** you **must** pass a `raw_window_handle` if you plan to use
    ///   this context with that window.
```

glutin/src/display.rs
```diff
@@ -63,11 +63,11 @@ pub trait GlDisplay: Sealed {
...
-    /// **Wayland:** - This call may latch the underlying back buffer of
-    /// the currently active context (will do with mesa drivers), meaning
-    /// that all resize operations will apply for it after the next
-    /// [`GlSurface::swap_buffers`]. To workaround this behavior the current
-    /// context should be made [`not current`].
+    /// - **Wayland:** this call may latch the underlying back buffer of the
+    ///   currently active context (will do with mesa drivers), meaning that all
+    ///   resize operations will apply for it after the next
+    ///   [`GlSurface::swap_buffers`]. To workaround this behavior the current
+    ///   context should be made [`not current`].
```
MarijnS95, line 70: Was there any good documentation on what this "latching" actually means (i.e. as far as I understand it a resize event does not "grab on" to the current buffer if it was already drawed to but is instead postponed to the next, using roller coaster terminology where a chain may or may not latch into a train)?

glutin/src/display.rs
```diff
@@ -63,11 +63,11 @@ pub trait GlDisplay: Sealed {
...
     /// # Platform-specific
     ///
-    /// **Wayland:** - This call may latch the underlying back buffer of
-    /// the currently active context (will do with mesa drivers), meaning
-    /// that all resize operations will apply for it after the next
-    /// [`GlSurface::swap_buffers`]. To workaround this behavior the current
-    /// context should be made [`not current`].
+    /// - **Wayland:** this call may latch the underlying back buffer of the
+    ///   currently active context (will do with mesa drivers), meaning that all
+    ///   resize operations will apply for it after the next
```
MarijnS95, line 68: s/for/to or something like that?

glutin/src/surface.rs
```diff
@@ -476,10 +476,10 @@ impl<T: SurfaceTypeTrait> Sealed for Surface<T> {}
 ///
 /// # Platform specific
 ///
-/// **Wayland:** - when the window is hidden and [`SwapInterval::Wait`] is used
-/// [`GlSurface::swap_buffers`] and based on it functions may block until the
-/// window is visible again. Using this variant is not recommended on Wayland
-/// and instead the throttling should be performed by [`frame callbacks`].
+/// - **Wayland:** when the window is hidden and [`SwapInterval::Wait`] is used
+///   [`GlSurface::swap_buffers`] and based on it functions may block until the
```
MarijnS95, line 480: s/and based on it functions/and any functions based on it?

Not entirely sure what the comment is trying to say here.

MarijnS95 (approved):
UL changes are good, got some wording suggestions for the rest :+1:

glutin/src/display.rs
kchibisov: that's what spec uses, it simply means locked until next swap buffers. Similar to modifier latching where it's depressed until the next press or something.

glutin/src/surface.rs
```diff
@@ -476,10 +476,11 @@ impl<T: SurfaceTypeTrait> Sealed for Surface<T> {}
 ///
 /// # Platform specific
 ///
-/// **Wayland:** - when the window is hidden and [`SwapInterval::Wait`] is used
-/// [`GlSurface::swap_buffers`] and based on it functions may block until the
-/// window is visible again. Using this variant is not recommended on Wayland
-/// and instead the throttling should be performed by [`frame callbacks`].
+/// - **Wayland:** when the window is hidden and [`SwapInterval::Wait`] is used
+///   [`GlSurface::swap_buffers`] and any function functions based on it may
```
MarijnS95, line 480:
```suggestion
///   [`GlSurface::swap_buffers`] and any functions based on it may
```

PR #1606, opened 2023-06-10
Check for null window handles

### LINKED ISSUE #1604: Check the raw window handles for null values before using them
It is perfectly valid for the `RawWindowHandle` type to contain null or zero values. In some cases this crate doesn't check and assumes that the values are valid.

notgull:

Closes #1604

### DIFF
--- a/glutin/src/error.rs
+++ b/glutin/src/error.rs
@@ -133,9 +133,6 @@ pub enum ErrorKind {
     /// Bad native window was provided.
     BadNativeWindow,
 
-    /// Bad native display was provided.
-    BadNativeDisplay,
-
     /// The context was lost.
     ContextLost,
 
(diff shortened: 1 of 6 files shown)


kchibisov (changes requested):
Could you change all the `BadMatch` to `BadNativeWindow`? In general some functions internally check for that already.

The same should apply to Pixmap handling code.

glutin/src/error.rs
```diff
@@ -163,6 +166,7 @@ impl ErrorKind {
             BadMatch => "arguments are inconsistance",
             BadParameter => "one or more argument values are invalid",
             BadNativePixmap => "argument does not refer to a valid native pixmap",
+            BadNativeDisplay => "argument does not refer to a valid native display",
```
kchibisov, line 169: There's a `BadDisplay` for that already.

kchibisov (approved):

PR #1609, opened 2023-06-28
egl: Use EGL_KHR_display_reference

i509VCB:
This resolves a past issue where glutin would not be able to terminate a display due to risk of two displays being created from the same native display.

- [X] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [X] Created or updated an example program if it would help users understand this functionality

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - **Breaking:** `GlContext` trait is now a part of the `prelude`.
 - Fixed lock on SwapBuffers with some GLX drivers.
 - Fixed EGL's `Surface::is_single_buffered` being inversed.
+- Use `EGL_KHR_display_reference` if available for EGL.
 
 # Version 0.30.8
 
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -2,6 +2,7 @@
 
 use std::collections::HashSet;
 use std::ffi::{self, CStr};
+use std::mem::MaybeUninit;
 use std::ops::Deref;
 use std::os::raw::c_char;
 use std::sync::Arc;
@@ -126,6 +127,10 @@ impl Display {
 
         let mut attrs = Vec::<EGLint>::with_capacity(2);
 
+        if extensions.contains("EGL_KHR_display_reference") {
+            attrs.push(egl::TRACK_REFERENCES_KHR as _);
+        }
+
         // TODO: Some extensions exist like EGL_EXT_device_drm which allow specifying
         // which DRM master fd to use under the hood by the implementation. This would
         // mean there would need to be an unsafe equivalent to this function.
@@ -192,7 +197,8 @@ impl Display {
 
         let extensions = NO_DISPLAY_EXTENSIONS.get().unwrap();
 
-        let mut attrs = Vec::<EGLAttrib>::new();
+        // Preallocate space for the terminator and to track references.
+        let mut attrs = Vec::<EGLAttrib>::with_capacity(2);
         let (platform, mut display) = match display {
             #[cfg(wayland_platform)]
             RawDisplayHandle::Wayland(handle)
@@ -219,6 +225,10 @@ impl Display {
             },
         };
 
+        if extensions.contains("EGL_KHR_display_reference") {
+            attrs.push(egl::TRACK_REFERENCES_KHR as _);
+        }
+
         // Be explicit here.
         if display.is_null() {
             display = egl::DEFAULT_DISPLAY as *mut _;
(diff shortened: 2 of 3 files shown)


kchibisov:
I think you'd need this to make it work from how I read the extension.

```diff
diff --git a/glutin/src/api/egl/display.rs b/glutin/src/api/egl/display.rs
index 43d043a..94c3dbd 100644
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -125,10 +125,11 @@ impl Display {
             .into());
         }
 
-        let mut attrs = Vec::<EGLint>::with_capacity(2);
+        let mut attrs = Vec::<EGLint>::with_capacity(3);
 
         if extensions.contains("EGL_KHR_display_reference") {
             attrs.push(egl::TRACK_REFERENCES_KHR as _);
+            attrs.push(egl::TRUE as _);
         }
 
         // TODO: Some extensions exist like EGL_EXT_device_drm which allow specifying
@@ -198,7 +199,7 @@ impl Display {
         let extensions = NO_DISPLAY_EXTENSIONS.get().unwrap();
 
         // Preallocate space for the terminator and to track references.
-        let mut attrs = Vec::<EGLAttrib>::with_capacity(2);
+        let mut attrs = Vec::<EGLAttrib>::with_capacity(3);
         let (platform, mut display) = match display {
             #[cfg(wayland_platform)]
             RawDisplayHandle::Wayland(handle)
@@ -227,6 +228,7 @@ impl Display {
 
         if extensions.contains("EGL_KHR_display_reference") {
             attrs.push(egl::TRACK_REFERENCES_KHR as _);
+            attrs.push(egl::TRUE as _);
         }
 
         // Be explicit here.
@@ -251,7 +253,7 @@ impl Display {
         let extensions = NO_DISPLAY_EXTENSIONS.get().unwrap();
 
         // Preallocate space for the terminator and to track references.
-        let mut attrs = Vec::<EGLint>::with_capacity(2);
+        let mut attrs = Vec::<EGLint>::with_capacity(3);
         let (platform, mut display) = match display {
             #[cfg(wayland_platform)]
             RawDisplayHandle::Wayland(handle)
@@ -286,6 +288,7 @@ impl Display {
 
         if extensions.contains("EGL_KHR_display_reference") {
             attrs.push(egl::TRACK_REFERENCES_KHR as _);
+            attrs.push(egl::TRUE as _);
         }
 
         // Be explicit here.
@@ -509,15 +512,21 @@ impl Drop for DisplayInner {
         if self.client_extensions.contains("EGL_KHR_display_reference") {
             let mut track_references = MaybeUninit::<u32>::uninit();
             unsafe {
-                self.egl.QueryDisplayAttribEXT(
-                    *self.raw,
-                    egl::TRACK_REFERENCES_KHR as _,
-                    track_references.as_mut_ptr().cast(),
-                );
-
-                if track_references.assume_init() == egl::TRUE {
+                if match self.raw {
+                    EglDisplay::Khr(khr) => self.egl.QueryDisplayAttribKHR(
+                        khr,
+                        egl::TRACK_REFERENCES_KHR as _,
+                        track_references.as_mut_ptr().cast(),
+                    ),
+                    EglDisplay::Ext(ext) => self.egl.QueryDisplayAttribEXT(
+                        ext,
+                        egl::TRACK_REFERENCES_KHR as _,
+                        track_references.as_mut_ptr().cast(),
+                    ),
+                    EglDisplay::Legacy(_) => return,
+                } == egl::TRUE
+                {
                     self.egl.Terminate(*self.raw);
-                    return;
                 }
             }
         }

```

kchibisov:
@MarijnS95 could you test this? The extension is not present for me with mesa, but from what I've see it was default on Android.

With my patch applied likely and test that it actually calls `eglTerminate` in the drop?

MarijnS95:
Indeed, [the documentation](https://registry.khronos.org/EGL/extensions/KHR/EGL_KHR_display_reference.txt) seems rather clear on requiring a `TRUE`/`FALSE` argument, so I'm curious how this was "tested on all platforms changed".

Unfortunately my Android 13 phone does not advertise `EGL_KHR_display_reference`, even though the docs mention that on `EGL_KHR_platform_android` the default value is `EGL_TRUE` (but I assume that only counts when the extension is there).

I bluntly tried to retrieve the value, but since `eglQueryDisplayAttribKHR` is provided by this extension (and `eglQueryDisplayAttribEXT` by `EGL_EXT_device_query` which isn't supported either), both panic as they cannot be loaded.

Strangely there's no `BadAttribute` error when adding the attribute (with or without boolean) to `GetPlatformDisplay`.

kchibisov:
Yeah, it'll need more work in general.

How did you test it @i509VCB , do you have nvidia card? mesa doesn't provide this extension from what I can see.

MarijnS95:
Indeed, Mesa doesn't even support it on my Intel laptop nor on Freedreno for one of the many "mainlined" phones I have lining around (there's a PR open, though...).

It seems NVidia should support it?

i509VCB:
> Yeah, it'll need more work in general.
> 
> How did you test it @i509VCB , do you have nvidia card? mesa doesn't provide this extension from what I can see.

I plan to try to test on some nvidia hardware I have, I need to install a new distro on the machine though

kchibisov:
See #1632.

PR #1611, opened 2023-06-29
Add support for creating glutin EGL displays on Windows using Angle

### LINKED ISSUE #1508: ANGLE support
It would be nice to work with ANGLE, it should be pretty trivial from what I can see.

This would help providing EGL on every possible platform.

tronical:
Angle provides libEGL.dll, which glutin loads. The only change necessary is to get the platform display via EGL_PLATFORM_ANGLE_ANGLE and treating it as a legacy display.

Fixes #1508

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - **Breaking:** `GlContext` trait is now a part of the `prelude`.
 - Fixed lock on SwapBuffers with some GLX drivers.
 - Fixed EGL's `Surface::is_single_buffered` being inversed.
+- Added support for EGL on Windows using Angle. This assumes libEGL.dll/libGLESv2.dll present.
 
 # Version 0.30.8
 
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -241,6 +241,7 @@ impl Display {
         let extensions = NO_DISPLAY_EXTENSIONS.get().unwrap();
 
         let mut attrs = Vec::<EGLint>::new();
+        let mut legacy = false;
         let (platform, mut display) = match display {
             #[cfg(wayland_platform)]
             RawDisplayHandle::Wayland(handle)
@@ -266,6 +267,11 @@ impl Display {
             RawDisplayHandle::Gbm(handle) if extensions.contains("EGL_MESA_platform_gbm") => {
                 (egl::PLATFORM_GBM_MESA, handle.gbm_device)
             },
+            RawDisplayHandle::Windows(..) if extensions.contains("EGL_ANGLE_platform_angle") => {
+                // Only CreateWindowSurface appears to work with Angle.
+                legacy = true;
+                (egl::PLATFORM_ANGLE_ANGLE, egl::DEFAULT_DISPLAY as *mut _)
+            },
             _ => {
                 return Err(
                     ErrorKind::NotSupported("provided display handle is not supported").into()
@@ -284,7 +290,17 @@ impl Display {
         let display =
             unsafe { egl.GetPlatformDisplayEXT(platform, display as *mut _, attrs.as_ptr()) };
 
-        Self::check_display_error(display).map(EglDisplay::Ext)
+        Self::check_display_error(display).map(|display| {
+            if legacy {
+                // NOTE: For angle we use the Legacy code path, as that uses CreateWindowSurface
+                // instead of CreatePlatformWindowSurface*. The latter somehow
+                // doesn't work, only the former does. But Angle's own example also use the
+                // former: https://github.com/google/angle/blob/main/util/EGLWindow.cpp#L424
+                EglDisplay::Legacy(display)
+            } else {
+                EglDisplay::Ext(display)
+            }
+        })
     }
 
     fn get_display(egl: &Egl, display: RawDisplayHandle) -> Result<EglDisplay> {
--- a/glutin/src/display.rs
+++ b/glutin/src/display.rs
@@ -426,6 +426,11 @@ pub enum DisplayApiPreference {
     ///
     /// But despite this issues it should be preferred on at least Linux over
     /// GLX, given that GLX is phasing away.
+    ///
+    /// # Platform-specific
+    ///
+    /// **Windows:** ANGLE can be used if `libEGL.dll` and `libGLESv2.dll` are
+    ///              in the library search path.
     #[cfg(egl_backend)]
     Egl,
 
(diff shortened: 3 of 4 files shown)


glutin/src/api/egl/display.rs
```diff
@@ -212,6 +213,10 @@ impl Display {
             RawDisplayHandle::Android(_) if extensions.contains("EGL_KHR_platform_android") => {
                 (egl::PLATFORM_ANDROID_KHR, egl::DEFAULT_DISPLAY as *mut _)
             },
+            RawDisplayHandle::Windows(..) => {
+                legacy = true;
+                (egl::PLATFORM_ANGLE_ANGLE, core::ptr::null_mut())
```
kchibisov, line 218: That's the wrong function to do it, you need the one with `EXT` according to the extension you've linked. See `get_platform_display_ext`.

I'd assume when using ANGLE the regular extension to use EGL on Windows is hidden? Though, for whatever reason I don't see us building the display for windows here in any of the display path, but it's for the different issue.

glutin/src/api/egl/display.rs
```diff
@@ -193,6 +193,7 @@ impl Display {
         let extensions = NO_DISPLAY_EXTENSIONS.get().unwrap();
 
         let mut attrs = Vec::<EGLAttrib>::new();
+        let mut legacy = false;
```
kchibisov, line 196: Could you link to the sources for that? It's not obvious from the extension that it should work like that at all. You could link to other libraries code wrt ANGLE integration.

glutin_egl_sys/src/lib.rs
```diff
@@ -32,6 +32,8 @@ pub mod egl {
     pub const PLATFORM_XCB_SCREEN_EXT: super::EGLenum = 0x31DE;
     // EGL_EXT_device_query_name
     pub const RENDERER_EXT: super::EGLenum = 0x335F;
+    // EGL_ANGLE_platform_angle - https://chromium.googlesource.com/angle/angle/+/HEAD/extensions/EGL_ANGLE_platform_angle.txt
+    pub const PLATFORM_ANGLE_ANGLE: super::EGLenum = 0x3202;
```
kchibisov, line 36: Could we add other ANGLE constants from the spec while at it?

glutin/src/api/egl/display.rs
kchibisov, line 218:
```suggestion
                (egl::PLATFORM_ANGLE_ANGLE, egl::DEFAULT_DISPLAY as *mut _)
```

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - **Breaking:** `GlContext` trait is now a part of the `prelude`.
 - Fixed lock on SwapBuffers with some GLX drivers.
 - Fixed EGL's `Surface::is_single_buffered` being inversed.
+- Added support for EGL on Windows using Angle. This assumes libEGL.dll/libGLESv2.dll present.
```
kchibisov, line 9: Could we document on the `DisplayApiPreference` that Windows could use ANGLE?

kchibisov (changes requested):
Thanks.

It's sort of weird that Windows EGL was simply forgotten in the display creation, since I don't see any method touching RWHND for it...

glutin/src/api/egl/display.rs
tronical: Good point, that's not obvious - and perhaps I'm doing something wrong here. I ran into CreatePlatformWindowSurface failing in angle as not implemented - only CreateWindowSurface worked - hence the legacy variant. But even angle's own example uses that: https://github.com/google/angle/blob/16d65289e8681e8f9f6f372f1173951d56dca450/util/EGLWindow.cpp#L424

tronical:
I'll push fixes to the review comments as follow-up patches. In the event that this finds approval in the end, I'd squash them into one, if that's okay.

glutin/src/display.rs
```diff
@@ -426,6 +426,9 @@ pub enum DisplayApiPreference {
     ///
     /// But despite this issues it should be preferred on at least Linux over
     /// GLX, given that GLX is phasing away.
+    /// 
+    /// On Windows, EGL is available via Angle, if Angle built `libEGL.dll`
+    /// and `libGLESv2.dll` are in the library search path.
```
kchibisov, line 431: We have `## Platform-Specific` section to isolate such thins, You don't have to mention `EGL` in this comment at all, since you document `EGL` preference, just say `**Windows:**  ANGLE could be used as a provider when it's in search path`, or something like that.

glutin/src/api/egl/display.rs
```diff
@@ -284,7 +290,13 @@ impl Display {
         let display =
             unsafe { egl.GetPlatformDisplayEXT(platform, display as *mut _, attrs.as_ptr()) };
 
-        Self::check_display_error(display).map(EglDisplay::Ext)
+        Self::check_display_error(display).map(|display| {
+            if legacy {
+                EglDisplay::Legacy(display)
```
kchibisov, line 295: Add a `NOTE:` comment here why we're using `Legacy` and probably saying about ANGLE examples even using `Legacy` interfaces.

kchibisov (changes requested):
In addition, could you post the screenshot of default example running on ANGLE (want to see output of example), so I could see how it works, given that I can't test it.

tronical:
Proof ;)

Screenshot 2023-06-29 at 17 05 22

tronical:
> @tronical oh, could it work in qemu setup or only vmware specific? Or where you using gpu passthrough for that?

I'm using VMware's D3D drivers, those I think do pass-through to some extent. But in principle this should work absolutely anywhere, as D3D comes with the WARP software renderer. I think/hope that Angle falls back to selecting that if no other was found. Otherwise the angle extensions also allow to explicitly select a warp D3D device. This could be an interesting glutin API extension in the future.

Toolkits that use Warp often have some environment variable that allows users to explicitly opt into using warp.

kchibisov (approved):
Looks good after `cargo +nightly fmt`.

glutin/src/api/egl/display.rs
```diff
@@ -266,6 +267,11 @@ impl Display {
             RawDisplayHandle::Gbm(handle) if extensions.contains("EGL_MESA_platform_gbm") => {
                 (egl::PLATFORM_GBM_MESA, handle.gbm_device)
             },
+            RawDisplayHandle::Windows(..) => {
```
kchibisov, line 270: Oh, sorry, could we check for `ANGLE` extension here just in case?

tronical: Ah yes, good point! Will try to fix that tomorrow.

kchibisov (approved):

kchibisov:
Thanks!

PR #1617, opened 2023-07-31
Update `objc2` to v0.4

madsmtm:


rustfmt.toml
```diff
@@ -7,7 +7,7 @@ overflow_delimited_expr = true
 imports_granularity = "Module"
 use_small_heuristics = "Max"
 normalize_comments = true
-reorder_impl_items = true
+# reorder_impl_items = true # Currently breaks objc2::declare_class!
```
madsmtm, line 10: Tracking this in https://github.com/madsmtm/objc2/issues/479

glutin/src/api/cgl/appkit.rs
```diff
@@ -143,60 +112,3 @@ extern_methods!(
         );
     }
 );
-
```
madsmtm, line 146: This is moved to `icrate`, where it has been autogenerated instead of manually written like this.

glutin/src/api/cgl/surface.rs
```diff
@@ -85,8 +91,9 @@ impl Display {
 pub struct Surface<T: SurfaceTypeTrait> {
     display: Display,
     config: Config,
-    pub(crate) ns_view: Id<NSObject, Shared>,
-    ns_window: Id<NSObject, Shared>,
+    pub(crate) ns_view: MainThreadBound<Id<NSView>>,
+    ns_window: MainThreadBound<Id<NSWindow>>,
+    _nosendsync: PhantomData<*const std::ffi::c_void>,
```
madsmtm, line 96: I'm fairly sure this is now thread safe (by moving all calls on the view and window to the main thread, including dropping), though if you want, then I can make it usable from the main thread only? I'm unsure what makes sense.

kchibisov: It should be used from any thread, just not being able to send across.

madsmtm: Okay, I guess that part works now

glutin/src/api/cgl/appkit.rs
```diff
@@ -1,46 +1,19 @@
...
-/// Run closure on the main thread.
-pub(crate) fn run_on_main<R: Send>(f: impl FnOnce() -> R + Send) -> R {
-    if is_main_thread() {
-        f()
-    } else {
-        Queue::main().exec_sync(f)
-    }
+#[repr(C)]
+pub struct CGLContextObj {
+    __inner: [u8; 0],
```
kchibisov, line 16: What's the reason to have zst like that?

glutin/src/api/cgl/context.rs
```diff
@@ -216,11 +222,10 @@ impl ContextInner {
...
             self.update();
             self.raw.makeCurrentContext();
-            let raw = MainThreadSafe(&self.raw);
-            let ns_view = MainThreadSafe(&surface.ns_view);
 
-            run_on_main(move || unsafe {
-                raw.setView(Some(*ns_view));
+            let view = &surface.ns_view;
+            MainThreadMarker::run_on_main(|mtm| unsafe {
+                self.raw.setView(Some(&view.get(mtm)));
```
kchibisov, line 228:
```suggestion
                self.raw.setView(Some(view.get(mtm)));
```

glutin/src/api/cgl/surface.rs
```diff
@@ -54,13 +53,18 @@ impl Display {
             },
         };
 
+        // SAFETY: The objects below must have been created on the main thread
+        // in the first place, so we can safely "move" them back to that thread.
+        let mtm = unsafe { MainThreadMarker::new_unchecked() };
```
kchibisov, line 58: I guess it's just an indicator that the object belongs to the main thread?

rustfmt.toml
kchibisov: Could you instead of doing this, do `#[rustfmt(skip)]`? Or is the issue coming from the generated code?

glutin/src/api/cgl/appkit.rs
madsmtm: Ideally of course, this would be an extern type, but I changed it from an uninhabited enum to a zero-sized type to [avoid possible mistakes](https://doc.rust-lang.org/nomicon/ffi.html#representing-opaque-structs) - it was mostly to follow better practice, not that important for this specific code.

glutin/src/api/cgl/surface.rs
madsmtm: Yes, _and_ the `Drop` impl for the `T` in `MainThreadBound<T>` is called on the main thread, so the safety comment here is saying "yes, we do want to run that on the main thread" (and in fact that's the only sound thing to do).

kchibisov (approved):

Issue #1588, opened 2023-04-07
Support `EGL_KHR_display_reference` to clean up `EGL` properly


kchibisov:
This extension defines a ref counting for the `EGLDisplay` which we could use to properly call `eglTerminate` instead of leaking like we have it now.

kchibisov:
cc @i509VCB given that you had some concerns about not calling `eglTerminate` before.

i509VCB:
This is definitely a solution to the problem for displays which glutin has created and owns.

For glutin do we want to always set that attribute if supported? What about things which borrow the display, should cloning the display increment the ref count of the display reference or should we continue with the interior mutability in that case?

For displays which glutin does not own or takes ownership of it appears we can query if `EGL_TRACK_REFERENCES_KHR` is set for the display.

kchibisov:
it's not a driver extension, it's a platform extension, like a `buffer_age`.

We want to always set it if supported and don't anything otherwise.

We should increment on `Clone`, and we don't use interior mutability at all(we simply keep it inside the `Arc`).

The `GetGlDisplay` simply clones and rebuilds the display, so the reference must be incremented each time internal `Arc` is cloned.

PR #1626, opened 2023-08-16
egl: use EGL_KHR_display_reference

kchibisov:
This resolves a past issue where glutin would not be able to terminate a display due to risk of two displays being created from the same native display.

Besides not terminating EGL raises issue with nvidia on Wayland, thus the `api::egl::Display::terminate` is added to help clients call this function when calling `eglTerminate` is generally unsafe to do automatically when dropping the `Display`.

Suggested-by: i509VCB <git@i509.me>
Links: https://github.com/alacritty/alacritty/issues/7146
Fixes: #1588

### DIFF
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -2,6 +2,7 @@
 
 use std::collections::HashSet;
 use std::ffi::{self, CStr};
+use std::mem::MaybeUninit;
 use std::ops::Deref;
 use std::os::raw::c_char;
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


glutin/src/api/egl/display.rs
```diff
@@ -185,14 +191,32 @@ impl Display {
...
+    /// Terminate the EGL display.
+    ///
+    /// This function will destroy the global EGL state, even the one created
+    /// and managed by other libraries. Use this function only when you bring
+    /// everything down.
+    ///
+    /// When the display is managed by glutin with the
+    /// `EGL_KHR_display_reference` this function does nothing and
+    /// `eglTerminate` will be automatically invoked during display destruction.
+    pub unsafe fn terminate(&self) {
```
notgull, line 203: Shouldn't his take `self`, not `&self`?

kchibisov: Doesn't matter, it won't help.

kchibisov: To clarify on that, the display is `Clone`, and this function brings down global single ton, so taking `self` could look like saner, but it's all around the same issues. I'll mark it `self` just for consistency, but it doesn't help with anything.

kchibisov:
See #1632.

PR #1632, opened 2023-10-09
egl: use EGL_KHR_display_reference

Molytho:
This is a updated version of #1626 which intends to fix https://github.com/alacritty/alacritty/issues/7146

### DIFF
--- a/glutin/src/api/egl/device.rs
+++ b/glutin/src/api/egl/device.rs
@@ -9,7 +9,7 @@ use glutin_egl_sys::egl::types::EGLDeviceEXT;
 
 use crate::error::{ErrorKind, Result};
 
-use super::display::{extensions_from_ptr, get_extensions, NO_DISPLAY_EXTENSIONS};
+use super::display::{extensions_from_ptr, get_extensions, CLIENT_EXTENSIONS};
 use super::{Egl, EGL};
 
 /// Wrapper for `EGLDevice`.
@@ -34,7 +34,7 @@ impl Device {
         };
 
         let no_display_extensions =
-            NO_DISPLAY_EXTENSIONS.get_or_init(|| get_extensions(egl, egl::NO_DISPLAY));
+            CLIENT_EXTENSIONS.get_or_init(|| get_extensions(egl, egl::NO_DISPLAY));
 
         // Querying devices requires EGL_EXT_device_enumeration and
         // EGL_EXT_device_query.
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -30,7 +30,7 @@ use super::surface::Surface;
 use super::{Egl, EGL};
 
 /// Extensions that don't require any display.
-pub(crate) static NO_DISPLAY_EXTENSIONS: OnceCell<HashSet<&'static str>> = OnceCell::new();
+pub(crate) static CLIENT_EXTENSIONS: OnceCell<HashSet<&'static str>> = OnceCell::new();
 
 /// A wrapper for the `EGLDisplay` and its supported extensions.
 #[derive(Debug, Clone)]
@@ -54,7 +54,7 @@ impl Display {
             None => return Err(ErrorKind::NotFound.into()),
         };
 
-        NO_DISPLAY_EXTENSIONS.get_or_init(|| get_extensions(egl, egl::NO_DISPLAY));
+        CLIENT_EXTENSIONS.get_or_init(|| get_extensions(egl, egl::NO_DISPLAY));
 
         // Create a EGL display by chaining all display creation functions aborting on
         // `EGL_BAD_ATTRIBUTE`.
@@ -113,7 +113,7 @@ impl Display {
         // Okay to unwrap here because the client extensions must have been enumerated
         // while querying the available devices or the device was gotten from an
         // existing display.
-        let extensions = NO_DISPLAY_EXTENSIONS.get().unwrap();
+        let extensions = CLIENT_EXTENSIONS.get().unwrap();
 
         if !extensions.contains("EGL_EXT_platform_base")
             && !extensions.contains("EGL_EXT_platform_device")
@@ -150,7 +150,7 @@ impl Display {
     /// This function returns [`Err`] if the `EGL_EXT_device_query` or
     /// `EGL_EXT_device_base` extensions are not available.
     pub fn device(&self) -> Result<Device> {
-        let no_display_extensions = NO_DISPLAY_EXTENSIONS.get().unwrap();
+        let no_display_extensions = CLIENT_EXTENSIONS.get().unwrap();
 
         // Querying the device of a display only requires EGL_EXT_device_query, but we
         // also check if EGL_EXT_device_base is available since
@@ -190,7 +190,7 @@ impl Display {
             return Err(ErrorKind::NotSupported("eglGetPlatformDisplay is not supported").into());
         }
 
-        let extensions = NO_DISPLAY_EXTENSIONS.get().unwrap();
+        let extensions = CLIENT_EXTENSIONS.get().unwrap();
 
         let mut attrs = Vec::<EGLAttrib>::new();
         let (platform, mut display) = match display {
@@ -238,7 +238,7 @@ impl Display {
             return Err(ErrorKind::NotSupported("eglGetPlatformDisplayEXT is not supported").into());
         }
 
-        let extensions = NO_DISPLAY_EXTENSIONS.get().unwrap();
+        let extensions = CLIENT_EXTENSIONS.get().unwrap();
 
         let mut attrs = Vec::<EGLint>::new();
         let mut legacy = false;
@@ -382,15 +382,15 @@ impl Display {
         };
 
         // Load extensions.
-        let client_extensions = get_extensions(egl, *display);
-        let features = Self::extract_display_features(&client_extensions, version);
+        let display_extensions = get_extensions(egl, *display);
+        let features = Self::extract_display_features(&display_extensions, version);
 
         let inner = Arc::new(DisplayInner {
             egl,
             raw: display,
             _native_display: raw_display_handle.map(NativeDisplay),
             version,
-            client_extensions,
+            display_extensions,
             features,
         });
         Ok(Self { inner })
@@ -458,7 +458,7 @@ impl GlDisplay for Display {
 
 impl GetDisplayExtensions for Display {
     fn extensions(&self) -> &HashSet<&'static str> {
-        &self.inner.client_extensions
+        &self.inner.display_extensions
     }
 }
 
@@ -480,8 +480,8 @@ pub(crate) struct DisplayInner {
     /// The version of the egl library.
     pub(crate) version: Version,
 
-    /// Client EGL extensions.
-    pub(crate) client_extensions: HashSet<&'static str>,
+    /// Display EGL extensions.
+    pub(crate) display_extensions: HashSet<&'static str>,
 
     /// The features supported by the display.
     pub(crate) features: DisplayFeatures,
@@ -496,7 +496,7 @@ impl fmt::Debug for DisplayInner {
             .field("raw", &self.raw)
             .field("version", &self.version)
             .field("features", &self.features)
-            .field("extensions", &self.client_extensions)
+            .field("extensions", &self.display_extensions)
             .finish()
     }
 }
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -127,28 +127,49 @@ impl Display {
 
         let mut attrs = Vec::<EGLint>::with_capacity(3);
 
-        if extensions.contains("EGL_KHR_display_reference") {
-            attrs.push(egl::TRACK_REFERENCES_KHR as _);
-            attrs.push(egl::TRUE as _);
-        }
-
         // TODO: Some extensions exist like EGL_EXT_device_drm which allow specifying
         // which DRM master fd to use under the hood by the implementation. This would
         // mean there would need to be an unsafe equivalent to this function.
 
+        // Push at the end so we can pop it on failure
+        let mut has_display_reference = extensions.contains("EGL_KHR_display_reference");
+        if has_display_reference {
+            attrs.push(egl::TRACK_REFERENCES_KHR as _);
+            attrs.push(egl::TRUE as _);
+        }
+
         // Push `egl::NONE` to terminate the list.
         attrs.push(egl::NONE as EGLint);
 
-        let display = Self::check_display_error(unsafe {
-            egl.GetPlatformDisplayEXT(
-                egl::PLATFORM_DEVICE_EXT,
-                device.raw_device() as *mut _,
-                attrs.as_ptr(),
-            )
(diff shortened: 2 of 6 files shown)


kchibisov:
What's the reason of submitting my code again which you can't test? It's not like I'll use this over what I had, since I don't even understand what are you trying to do with your loops?

Molytho:
First there are some notable changes in the drop terminate path: it has to check the availability of EGL_KHR_display_reference against the client extensions and not the display extensions like it does in your version.
You also used a wrong type in the call to QueryDisplayAttribKHR/QueryDisplayAttribEXT.

Regarding the loop:
There is an issue with the egl dispatching done by libglvnd. It reports client extensions if one vendor library supports them. This can lead to display creation failure if the library exposing EGL_KHR_display_reference fails to create a display for another reason.
An example is a system with mesa and the nvidia libraries but without an nvidia card (I know this is rather theoretical but I suspect this to also happen in a nvidia+intel setup when trying to uses the intel card). I was able to confirm this issue with my laptop with intel graphics.
To address this I added this loop to try again without EGL_KHR_display_reference if it failed using it.

Regarding the testing:
I was able to test this on my ArchLinux setup through patching libglvnd as suggested here:
https://github.com/alacritty/alacritty/issues/7146#issuecomment-1701912925
i will try to get a fix upsteamed to libglvnd soon.
Currently I have a working alacritty build which does no longer crash.

kchibisov:
>it has to check the availability of EGL_KHR_display_reference against the client extensions and not the display extensions like it does in your version.

But you've just renamed the variables, you still query extensions the way it was before and you still check things the way they were before, you just changed a name in the codebase. 

> You also used a wrong type in the call to QueryDisplayAttribKHR/QueryDisplayAttribEXT.

`u32` which you're using is the wrong type, according to the specification. The `EGLAttrib` is what I've used and what spec uses, if it's not `EGLAttrib` the driver or the specification should be fixed. I won't be using the wrong type because that's what nvidia/libglvnd wants.
https://github.com/KhronosGroup/EGL-Registry/blob/b055c9b483e70ecd57b3cf7204db21f5a06f9ffe/extensions/KHR/EGL_KHR_display_reference.txt#L84 . The type is the same for both `KHR` and `EXT`.

>Regarding the loop:
There is an issue with the egl dispatching done by libglvnd. It reports client extensions if one vendor library supports them. This can lead to display creation failure if the library exposing EGL_KHR_display_reference fails to create a display for another reason.
An example is a system with mesa and the nvidia libraries but without an nvidia card (I know this is rather theoretical but I suspect this to also happen in a nvidia+intel setup when trying to uses the intel card). I was able to confirm this issue with my laptop with intel graphics.
To address this I added this loop to try again without EGL_KHR_display_reference if it failed using it.

Sounds like the bug should go upstream and not to glutin. Adding hacks like that means that no-one will ever fix it upstream, especially when nvidia is at play, since no-one will bother. I'd
rather have a way to naturally call `eglTerminate` even without the extension, since I can change semantics in the `0.31.0` version.

>I was able to test this on my ArchLinux setup through patching libglvnd as suggested here:

Good, that means we have something to work with, but I'd suggest to leave review comments on original PR or at least, explain what you're trying to do when submitting patches which were waiting for their testing. It's not like I can't review your comments to fix stuff, and the
way you've submitted looked like spam to me(since such things happen from time to time).

In general, I'm fine with the renaming, but not fine with the loop and use of `u32`.

kchibisov:
>I was able to test this on my ArchLinux setup through patching libglvnd as suggested here:

Also, if you haven't understood why I said that you haven't tested, it's because you have unchecked checkbox for 'testing stuff'. Which I have because mesa doesn't support this extension thus I can't test myself, but given that you can test and tested you should mark check it.

Molytho:
> > it has to check the availability of EGL_KHR_display_reference against the client extensions and not the display extensions like it does in your version.
>
> But you've just renamed the variables, you still query extensions the way it was before and you still check things the way they were before, you just changed a name in the codebase.

That's not true:
https://github.com/rust-windowing/glutin/blob/ff69874de6f68bdbbdf46fad69a8ec3934d2c7c3/glutin/src/api/egl/display.rs#L559
https://github.com/kchibisov/glutin/blob/b39853aaf68da3c9dbb14a2d69a8de7f22fe20e7/glutin/src/api/egl/display.rs#L548C74-L548C74
I use CLIENT_EXTENSIONS while you use self.client_extensions (renamed to self.display_extensions)

> > You also used a wrong type in the call to QueryDisplayAttribKHR/QueryDisplayAttribEXT.
> 
> `u32` which you're using is the wrong type, according to the specification. The `EGLAttrib` is what I've used and what spec uses, if it's not `EGLAttrib` the driver or the specification should be fixed. I won't be using the wrong type because that's what nvidia/libglvnd wants. https://github.com/KhronosGroup/EGL-Registry/blob/b055c9b483e70ecd57b3cf7204db21f5a06f9ffe/extensions/KHR/EGL_KHR_display_reference.txt#L84 . The type is the same for both `KHR` and `EXT`.

You misread that. You where using u32.
https://github.com/kchibisov/glutin/blob/b39853aaf68da3c9dbb14a2d69a8de7f22fe20e7/glutin/src/api/egl/display.rs#L549
I changed it to EGLAttrib

> > Regarding the loop:
> > There is an issue with the egl dispatching done by libglvnd. It reports client extensions if one vendor library supports them. This can lead to display creation failure if the library exposing EGL_KHR_display_reference fails to create a display for another reason.
> > An example is a system with mesa and the nvidia libraries but without an nvidia card (I know this is rather theoretical but I suspect this to also happen in a nvidia+intel setup when trying to uses the intel card). I was able to confirm this issue with my laptop with intel graphics.
> > To address this I added this loop to try again without EGL_KHR_display_reference if it failed using it.
> 
> Sounds like the bug should go upstream and not to glutin. Adding hacks like that means that no-one will ever fix it upstream, especially when nvidia is at play, since no-one will bother. I'd rather have a way to naturally call `eglTerminate` even without the extension, since I can change semantics in the `0.31.0` version.

I can drop this part but to clarify: I think this is not the fault of nvidia neither do I think this is fixable upstream. The way egl dispatching currently work in combination with EGL_KHR_display_reference being a client extension just doesn't allow libglvnd to do something else except joining all client extensions supported by any vendor.
I don't like this either but I can't think of a better solution anywhere.

> > I was able to test this on my ArchLinux setup through patching libglvnd as suggested here:
> 
> Good, that means we have something to work with, but I'd suggest to leave review comments on original PR or at least, explain what you're trying to do when submitting patches which were waiting for their testing. It's not like I can't review your comments to fix stuff, and the way you've submitted looked like spam to me(since such things happen from time to time).

Sorry for that. I didn't check the box because I didn't test it on windows which might also be effected by this patch.
If you prefer I can close this PR and add some comments in yours.

> In general, I'm fine with the renaming, but not fine with the loop and use of `u32`.

kchibisov:
>I use CLIENT_EXTENSIONS while you use self.client_extensions (renamed to self.display_extensions)

This part I know that it's wrong, the way you said it sounded like something more than that was wrong. I just hasn't bothered to fix, since after checking user's `eglinfo` the extension wasn't there in any instance.

Also, maybe we can merge all the extensions together? Because `NO_DISPLAY` and regular ones have no intersection, so if we can merge them it'll be more clear what extensions are available, since they are both available from what I can say.

>You misread that. You where using u32.

aye, sorry. I used that `cast()` thing again...

>I can drop this part but to clarify: I think this is not the fault of nvidia neither do I think this is fixable upstream. The way egl dispatching currently work in combination with EGL_KHR_display_reference being a client extension just doesn't allow libglvnd to do something else except joining all client extensions supported by any vendor.
I don't like this either but I can't think of a better solution anywhere.

I don't like the loop, but if there's no other way around it, I'd suggest to add the reference stuff in the end, and if adding them fails, just `pop` and retry.

>Sorry for that. I didn't check the box because I didn't test it on windows which might also be effected by this patch.
If you prefer I can close this PR and add some comments to yours.

Given that you're the one which can test you can continue with the patch. It was better initially to either comment or open a PR against my PR and ping me (github doesn't notify on forks by default).

glutin/src/api/egl/display.rs
```diff
@@ -185,122 +191,180 @@ impl Display {
...
     fn get_platform_display(egl: &Egl, display: RawDisplayHandle) -> Result<EglDisplay> {
         if !egl.GetPlatformDisplay.is_loaded() {
             return Err(ErrorKind::NotSupported("eglGetPlatformDisplay is not supported").into());
         }
 
-        let extensions = NO_DISPLAY_EXTENSIONS.get().unwrap();
+        let extensions = CLIENT_EXTENSIONS.get().unwrap();
+
+        let mut first_try = true;
+        loop {
```
kchibisov, line 221: We don't need this massive loop, try to structure in the following way

```rust
let mut attrs = Vec;
// Push display stuff

// Push reference extension

let mut remove_reference = false; // This is your first try logic.
let display = loop {
if remove_reference {
    attrs.pop();
    attrs.pop();
    attrs.pop();
    attrs.push(NONE);
};
display
}

```

glutin/src/api/egl/device.rs
```diff
@@ -34,7 +34,7 @@ impl Device {
         };
 
         let no_display_extensions =
-            NO_DISPLAY_EXTENSIONS.get_or_init(|| get_extensions(egl, egl::NO_DISPLAY));
+            CLIENT_EXTENSIONS.get_or_init(|| get_extensions(egl, egl::NO_DISPLAY));
```
kchibisov, line 37: Ack, and keep in the separate commit like it is now.

glutin/src/api/egl/display.rs
```diff
@@ -124,7 +125,12 @@ impl Display {
             .into());
         }
 
-        let mut attrs = Vec::<EGLint>::with_capacity(2);
+        let mut attrs = Vec::<EGLint>::with_capacity(3);
+
+        if extensions.contains("EGL_KHR_display_reference") {
+            attrs.push(egl::TRACK_REFERENCES_KHR as _);
+            attrs.push(egl::TRUE as _);
+        }
```
kchibisov, line 133: Have you tested the device platform (it has example)? Do we need the same logic like with normal displays?

glutin/src/api/egl/display.rs
```diff
@@ -185,122 +191,180 @@ impl Display {
...
     fn get_platform_display_ext(egl: &Egl, display: RawDisplayHandle) -> Result<EglDisplay> {
         if !egl.GetPlatformDisplayEXT.is_loaded() {
             return Err(ErrorKind::NotSupported("eglGetPlatformDisplayEXT is not supported").into());
         }
 
-        let extensions = NO_DISPLAY_EXTENSIONS.get().unwrap();
+        let extensions = CLIENT_EXTENSIONS.get().unwrap();
+
+        let mut first_try = true;
+        loop {
```
kchibisov, line 287: Same comment about the loop.

kchibisov (changes requested):

glutin/src/api/egl/display.rs
Molytho: I swapped to an or_else solution using pop and push which is definitely cleaner than the loop thing

glutin/src/api/egl/display.rs
Molytho: Yes we do. Missed that one

Molytho:
> Also, maybe we can merge all the extensions together? Because `NO_DISPLAY` and regular ones have no intersection, so if we can merge them it'll be more clear what extensions are available, since they are both available from what I can say.

We could but I think this isn't a good idea. Client extensions and display extensions are intentionally kept apart by the spec as they do serve different purposes. Most client extensions are useless after display creation so it's just waste of space to store them together with the display extensions.
I also don't see an issue with the current NO_DISPLAY_EXTENSIONS/CLIENT_EXTENSIONS global.

glutin/src/api/egl/display.rs
```diff
@@ -224,23 +268,45 @@ impl Display {
...
+
         // Push `egl::NONE` to terminate the list.
         attrs.push(egl::NONE as EGLAttrib);
 
-        let display =
+        let platform_display =
             unsafe { egl.GetPlatformDisplay(platform, display as *mut _, attrs.as_ptr()) };
 
-        Self::check_display_error(display).map(EglDisplay::Khr)
+        Self::check_display_error(platform_display)
```
kchibisov, line 284: I'm not sure I like this over the `loop` approach. It's less readable and we duplicate a lot of calls.

```rust
        // NOTE: EXPLAIN WHY ARE WE DOING THAT.
        let platform_display = loop {
            if !has_display_reference {
                attrs.pop();
                attrs.pop();
                attrs.pop();
                attrs.push(egl::NONE as EGLAttrib);
            }

            let platform_display =
                unsafe { egl.GetPlatformDisplay(platform, display as *mut _, attrs.as_ptr()) };

            match Self::check_display_error(platform_display) {
                Ok(display) => break Ok(display),
                Err(_) if has_display_reference => {
                    has_display_reference = false;
                },
                Err(err) => break Err(err),
            }
        };

        platform_display.map(EglDisplay::Khr)

```

I think something like this is more clear. Could you also add a `NOTE` explaining why we end up like that in more details? We'd also need this `NOTE` for each such loop.

glutin/src/api/egl/display.rs
```diff
@@ -124,33 +125,61 @@ impl Display {
...
+            if platform_display.is_err() && has_display_reference {
+                attrs.pop();
+                attrs.pop();
+                attrs.pop();
+                attrs.push(egl::NONE as EGLint);
+                has_display_reference = false;
+                continue;
+            }
+
+            break platform_display;
```
kchibisov, line 170: Could just do
```suggestion
            match Self::check_display_error(unsafe {
                egl.GetPlatformDisplay(platform, display as *mut _, attrs.as_ptr())
            }) {
                Err(_) if retry_on_failure => {
                    attrs.pop();
                    attrs.pop();
                    attrs.pop();
                    attrs.push(egl::NONE as EGLAttrib);
                    retry_on_failure = false;
                }
                platform_display => break platform_display,
            }
```

This or use `else`, there's no need to do `continue`.

glutin/src/api/egl/display.rs
```diff
@@ -536,21 +601,23 @@ impl DisplayInner {
         // terminate the display without worry for the instance being
         // reused elsewhere.
         let mut track_references = MaybeUninit::<EGLAttrib>::uninit();
-        unsafe {
-            (match self.raw {
-                EglDisplay::Khr(khr) => self.egl.QueryDisplayAttribKHR(
+        (match self.raw {
```
kchibisov, line 604: Could drop parens around the `match`, but other than that it should be good to go.

kchibisov (approved):

glutin/src/api/egl/display.rs
kchibisov: Hm, indeed. Pretty weird.

kchibisov:
Thanks, I'd assume that we still need libglvnd update?

Molytho:
Yes we do. I opened an issue upstream:
https://gitlab.freedesktop.org/glvnd/libglvnd/-/issues/244

PR #1637, opened 2023-10-21
chore: bump winit to 0.29.2

kchibisov:


MarijnS95:
Wasn't there a plan to bump `raw-window-handle` to `0.6` or is @notgull going to do something special with the new lifetimed handles here?

notgull:
I was going to try to implement borrowed handles here as well, see https://github.com/rust-windowing/glutin/pull/1582

kchibisov:
Basically yes, and we can add a `rwh_06` as a part of 0.31.1, it's not a big deal to convert them internally.

MarijnS95:
Heh, I wondered where the PR went and didn't see a PR along the lines of `Update raw-window-handle to 0.6`, that explains.

Issue #1640, opened 2023-10-24
Cannot create CGL context


tronical:
With the latest glutin-winit release (0.4.2), creating a (C)GL context does not work anymore. This is reproducible for me with the `window` example:

```
 % RUST_BACKTRACE=1 cargo run --example window
    Finished dev [unoptimized + debuginfo] target(s) in 0.26s
     Running `/Users/simon/src/glutin/target/debug/examples/window`
Picked a config with 0 samples
[...]

PR #1641, opened 2023-10-24
Relax encoding requirements for `CGLContextObj`

madsmtm:
Fixes https://github.com/rust-windowing/glutin/issues/1640. The issue is that `CGLContextObj` changed its encoding in macOS Sonoma 14.0. The encoding of a type is something that `objc2` uses to debug-assert that the definition is written correctly, but of course, if the Rust code assumes that the encoding is one thing, and it then changes, then the debug assertion will now fail.

The correct fix will land at some point in https://github.com/madsmtm/objc2/pull/526, which will also help with such a problem not happening again, but we can fix the issue in the interim using the `relax-void-encoding` escape-hatch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fixed `CGLContextObj` having an invalid encoding on newer macOS versions.
+
 # Version 0.31.0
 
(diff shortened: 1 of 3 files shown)


tronical:
Thank you, this fixes the crash for me!

kchibisov (approved):

Issue #1647, opened 2023-11-18
Unable to Force GLES2 Context (Still picks up the wrong GLES version)


p0ryae:
My device supports GLES2, but glutin instead uses GLES1 (OpenGL ES-CM 1.1), even when I force it via context:
```rust
let context_attributes = ContextAttributesBuilder::new()
            .with_context_api(ContextApi::Gles(Some(glutin::context::Version::new(2, 0))))
            .build(Some(raw_window_handle));
```

I also try to force the API level for the config:

```rs
let builder = ConfigTemplateBuilder::new()
            .with_api(glutin::config::Api::GLES2)
            .with_alpha_size(8)
            .compatible_with_native_window(raw_window_handle)
            .with_surface_type(ConfigSurfaceTypes::WINDOW);
```

Printing config.api() gives the following output, also indicating the existence of GLES2 support:
```
I/RustStdoutStderr( 3377): Config supports API: Api(GLES1 | GLES2)
```

The android device has an Ardeno 200 GPU, which has GLES2 support.

It's worth noting that I get an error about `EGL_BAD_DISPLAY` initially before anything:

```
validate_display:254 error 3008 (EGL_BAD_DISPLAY)
```

I have no idea what's causing it anymore. Any ideas are appreciated.

p0ryae:
Good news. Took me long but finally.

I managed to address this problem by adding `CONTEXT_CLIENT_VERSION` attr, following with the major version (2 in my case). 

```rust
        if let Some(version) = version {
            if version.major == 2 {
                attrs.push(egl::CONTEXT_CLIENT_VERSION as EGLint);
                attrs.push(version.major as EGLint);
            }
        }
```

Please note that this has to be right before the push for ```egl::NONE``` attr. Otherwise, it is pointless.

I purposed my changes here: https://github.com/rust-windowing/glutin/pull/1648

Please review on your appropriate time.

kchibisov:
This somehow fell off my radar.

Could you tell me the `EGL` version you have on that device? Because this API is an alias to

```rust
            if let Some(version) = version {
                attrs.push(egl::CONTEXT_MAJOR_VERSION as EGLint);
                attrs.push(version.major as EGLint);
                attrs.push(egl::CONTEXT_MINOR_VERSION as EGLint);
                attrs.push(version.minor as EGLint);
            }

```

We already have around.

MarijnS95:
To back that up with spec:

[...]

PR #1648, opened 2023-11-20
Workaround for fallback to GLES1 on supported GLES2 platforms

p0ryae:
This aims to safely fix the issue reported in https://github.com/rust-windowing/glutin/issues/1647. Tested on my device. Shouldn't break any other functionalities. 

Please note that I only purpose this workaround for major version `2` in order to decrease the risk of breakage, since I have no idea how this will preform on other major versions. Either ways, I don't believe major version `3` should be needing this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Added workaround for fallback to GLES1 on supported GLES2 platforms (See https://github.com/rust-windowing/glutin/issues/1647).
+
 # Version 0.31.1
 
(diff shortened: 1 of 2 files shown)


kchibisov:
Fixed by #1655.

Let me know if you need a patch release for that thing.

p0ryae:
> Fixed by #1655.
> 
> Let me know if you need a patch release for that thing.

Thanks. Wouldn't mind a patch release, for sure.

kchibisov:
done.

PR #1654, opened 2023-12-16
chore: bump raw-window-handle to 0.6.0

vially:
> [!WARNING]
> I'm still very new to Rust so please review these changes *very* carefully.

Marked as a draft until all platforms are implemented:
- [x] Linux
- [x] Android
- [ ] Windows
- [ ] macOS

Bump `raw-window-handle` to version `0.6.0` which introduced a couple of API changes.

- [ ] Tested on all platforms changed (these changes affect all platforms but I'm only able to test them on Wayland)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- **Breaking:** Bump `raw-window-handle` to version `0.6.0`
+
 # Version 0.31.1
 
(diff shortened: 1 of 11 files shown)


MarijnS95:
You might want to look at #1582 which is promising to do this upgrade while at the same time utilizing the new lifetime pattern.

vially:
Ah, I missed that part, thanks for mentioning it. In that case I'll close this pull-request and wait for #1582 to land.

PR #1656, opened 2023-12-21
Reuse some features and code, fix inverted intradoc links

MarijnS95:
Found some very minor code cases that could reuse some other code.  Some features also already imply/enable other features and don't need to be repeated.

Some "intradoc" links were having `[]` inside backticks, meaning they won't be links but actual brackets in monospace in the final doc.

Finally, the `wayland` feature needs the `egl` backend to compile and function, and was likely intended to be turned on as a dependency.

### DIFF
--- a/glutin_examples/src/lib.rs
+++ b/glutin_examples/src/lib.rs
@@ -23,10 +23,10 @@ pub mod gl {
 }
 
 pub fn main(event_loop: winit::event_loop::EventLoop<()>) -> Result<(), Box<dyn Error>> {
-    // Only windows requires the window to be present before creating the display.
+    // Only Windows requires the window to be present before creating the display.
     // Other platforms don't really need one.
     //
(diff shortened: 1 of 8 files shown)


glutin_examples/src/lib.rs
```diff
@@ -112,7 +112,7 @@ pub fn main(event_loop: winit::event_loop::EventLoop<()>) -> Result<(), Box<dyn
                         .unwrap()
                 });
 
-                let attrs = window.build_surface_attributes(<_>::default());
+                let attrs = window.build_surface_attributes(Default::default());
```
kchibisov, line 115: no fish, sadge.

kchibisov (approved):

glutin_examples/src/lib.rs
MarijnS95: This made it a bit harder to patch in a `.with_srgb(Some(t/f))` for local testing :)

PR #1657, opened 2023-12-21
examples: Exit when the user presses `Escape`

MarijnS95:


kchibisov:
Hm, maybe set it in title as well, since I think it's only relevant on stacking wms where user doesn't know the hotkey to close a window.

MarijnS95:
You mean the window title? Or enable the close button in the window?

I won't see those because I don't use window borders/decorations in `sway`.

kchibisov:
>You mean the window title?

Yes, window title.

>I won't see those because I don't use window borders/decorations in sway.

You should know how to close any window on sway with the hotkey though.

MarijnS95:
> You should know how to close any window on sway with the hotkey though.

Sure, but I've grown accustomed to graphics (test) apps supporting <kbd>Escape</kbd> specifically.

kchibisov (approved):

PR #1662, opened 2023-12-27
Change surface to be `Send`

vially:
As it was [recently discussed on matrix](https://matrix.to/#/!DGpLzJTRzBDTwZiogk:matrix.org/$6f7u-9EhwSFbqft-00FjVEUlREsgknYNG0v0D_OQ0Q8?via=matrix.org&via=gnome.org), it turns out the GL `Surface` can also be made `Send` (similar to the `Context`, which is `Send` already).

> [!Warning]
> I've only tested the example program on Wayland, so there's a good chance it might not work on other platforms (and I don't have an easy way to test it either).

Also, I'm just getting started with Rust, so the example code is definitely not very idiomatic, and most likely also incorrect, so please review it carefully. However, I don't mind iterating on it to get it in a better shape, if someone is willing to guide me in the right direction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- **Breaking:** Change `Surface` to be `Send`. This makes it consistent with the context, so now they are both `Send` but not `Sync`.
+- Change `Surface` to be `Send`. This makes it consistent with the context, so now they are both `Send` but not `Sync`.
 
 # Version 0.31.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- **Breaking:** Change `Surface` to be `Send`. This makes it consistent with the context, so now they are both `Send` but not `Sync`.
+
 # Version 0.31.2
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- **Breaking:** Change `Surface` to be `Send`. This makes it consistent with the context, so now they are both `Send` but not `Sync`.
```
kchibisov, line 3: it's not a breaking change, just say that it's now `Send`.

kchibisov (changes requested):
I'll look into example latter.

CHANGELOG.md
vially: Fixed in https://github.com/rust-windowing/glutin/pull/1662/commits/5641543019c39a9a403bafd0198681f77b57264c.

kchibisov:
macOS tested.
X11 tested.

kchibisov:
Windows tested.

kchibisov (approved):

kchibisov:
Pushed 0.31.3 with that, let's see how it survives.

PR #1668, opened 2024-03-06
Add extra EGL extensions

chrisduerr:
This patch adds multiple new EGL extensions which are useful for
implementing Wayland applications. These have been mostly implemented
manually, since gl_generator lacks support for these newer EGL
extensions (except for `EGL_KHR_image_base`).

The following extensions were added:
 - EGL_KHR_image_base
 - EGL_WL_bind_wayland_display
 - EGL_WL_create_wayland_buffer_from_image

---

Currently this PR doesn't really implement a good way to access this new functionality, instead it just exposes the `EGL` static publicly. Ideally there would be some abstraction that retrieves the `Egl` object, while automatically making sure it is initialized.

So I see two options, feedback appreciated:
 - Expose method that basically just does `EGL.as_ref()?` (does not ensure initialization)
 - Add method to `Display` (requires display creation)

### DIFF
--- a/glutin/src/api/egl/mod.rs
+++ b/glutin/src/api/egl/mod.rs
@@ -28,7 +28,7 @@ pub mod device;
 pub mod display;
 pub mod surface;
 
-pub(crate) static EGL: Lazy<Option<Egl>> = Lazy::new(|| {
+pub static EGL: Lazy<Option<Egl>> = Lazy::new(|| {
     #[cfg(windows)]
     let paths = ["libEGL.dll", "atioglxx.dll"];
 
@@ -67,6 +67,11 @@ impl SymLoading for egl::Egl {
                 (egl_proc_address)(sym_name.as_bytes_with_nul().as_ptr() as *const ffi::c_void)
             }
         };
+
+        egl::BindWaylandDisplayWL::load_with(loader);
+        egl::UnbindWaylandDisplayWL::load_with(loader);
+        egl::QueryWaylandBufferWL::load_with(loader);
+        egl::CreateWaylandBufferFromImageWL::load_with(loader);
 
         Self::load_with(loader)
     }
--- a/glutin_egl_sys/src/egl.rs
+++ b/glutin_egl_sys/src/egl.rs
@@ -1,0 +1,249 @@
+//! Manual implementation of EGL bindings.
+//!
+//! This is necesarry since `gl_generator` is unmaintaned and incapable of
+//! generating bindings for some of the newer extensions.
+
+use std::ffi::c_uint;
+
+pub type khronos_utime_nanoseconds_t = super::khronos_utime_nanoseconds_t;
+pub type khronos_uint64_t = super::khronos_uint64_t;
+pub type khronos_ssize_t = super::khronos_ssize_t;
+pub type EGLNativeDisplayType = super::EGLNativeDisplayType;
+pub type EGLNativePixmapType = super::EGLNativePixmapType;
+pub type EGLNativeWindowType = super::EGLNativeWindowType;
+pub type EGLint = super::EGLint;
+pub type NativeDisplayType = super::EGLNativeDisplayType;
+pub type NativePixmapType = super::EGLNativePixmapType;
+pub type NativeWindowType = super::EGLNativeWindowType;
+
+include!(concat!(env!("OUT_DIR"), "/egl_bindings.rs"));
+
+// EGL_EXT_platform_xcb
+pub const PLATFORM_XCB_EXT: super::EGLenum = 0x31DC;
+pub const PLATFORM_XCB_SCREEN_EXT: super::EGLenum = 0x31DE;
+// EGL_EXT_device_query_name
+pub const RENDERER_EXT: super::EGLenum = 0x335F;
+// EGL_ANGLE_platform_angle - https://chromium.googlesource.com/angle/angle/+/HEAD/extensions/EGL_ANGLE_platform_angle.txt
+pub const PLATFORM_ANGLE_ANGLE: super::EGLenum = 0x3202;
+pub const PLATFORM_ANGLE_TYPE_ANGLE: super::EGLenum = 0x3203;
+pub const PLATFORM_ANGLE_TYPE_VULKAN_ANGLE: super::EGLenum = 0x3450;
+pub const PLATFORM_ANGLE_MAX_VERSION_MAJOR_ANGLE: super::EGLenum = 0x3204;
+pub const PLATFORM_ANGLE_MAX_VERSION_MINOR_ANGLE: super::EGLenum = 0x3205;
+pub const PLATFORM_ANGLE_DEBUG_LAYERS_ENABLED: super::EGLenum = 0x3451;
+pub const PLATFORM_ANGLE_NATIVE_PLATFORM_TYPE_ANGLE: super::EGLenum = 0x348F;
+pub const PLATFORM_ANGLE_TYPE_DEFAULT_ANGLE: super::EGLenum = 0x3206;
+pub const PLATFORM_ANGLE_DEVICE_TYPE_HARDWARE_ANGLE: super::EGLenum = 0x320A;
+pub const PLATFORM_ANGLE_DEVICE_TYPE_NULL_ANGLE: super::EGLenum = 0x345E;
+
+mod wayland_storage {
+    use super::FnPtr;
+    use super::__gl_imports::raw;
+
+    // EGL_WL_create_wayland_buffer_from_image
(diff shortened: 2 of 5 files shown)


MarijnS95:
Fwiw for #1658 I'd also like to use updated EGL bindings from `gl_generator`, which is very trivial to achieve: https://github.com/brendanzab/gl-rs/pull/546

Unfortunately it seems that efforts to transition ownership of `gl-rs` (which is still an important cornerstone in the Rust ecosystem) have stalled: https://github.com/brendanzab/gl-rs/issues/524

kchibisov:
I'm not sure how we should expose the ref to `EGL` and similar types, since we should have it for all the platforms where we use loader (so, except macOS).

Maybe we should have a method to manually load(which will return `Option`) and just a method to get a `&'static Egl` from the `Display`(this will always work)? And the manual loading will be present only for `EGL`, because the rest is a bit _special_ and shouldn't be used anyway in the context of device extensions.

I'm also interested in solving the generator issue upstream, but I'll see what we can do

chrisduerr:
I added some functions to access the APIs for EGL and GLX. This works for my usecase at least, so maybe that's enough to just merge it?

Let me know if you want a cleaner/more elaborate solution.

glutin/src/api/egl/mod.rs
```diff
@@ -68,6 +70,11 @@ impl SymLoading for egl::Egl {
             }
         };
 
+        egl::BindWaylandDisplayWL::load_with(loader);
+        egl::UnbindWaylandDisplayWL::load_with(loader);
+        egl::QueryWaylandBufferWL::load_with(loader);
+        egl::CreateWaylandBufferFromImageWL::load_with(loader);
```
kchibisov, line 76: Could we maybe make it somehow done externally? So the user will have to load them if they want to, otherwise you'd have to add all the extensions here, which might be not great.

glutin/src/api/egl/mod.rs
```diff
@@ -41,7 +41,9 @@ pub(crate) static EGL: Lazy<Option<Egl>> = Lazy::new(|| {
 type EglGetProcAddress = unsafe extern "C" fn(*const ffi::c_void) -> *const ffi::c_void;
 static EGL_GET_PROC_ADDRESS: OnceCell<libloading_os::Symbol<EglGetProcAddress>> = OnceCell::new();
 
-pub(crate) struct Egl(pub SymWrapper<egl::Egl>);
+/// EGL interface.
+#[allow(missing_debug_implementations)]
+pub struct Egl(pub SymWrapper<egl::Egl>);
```
kchibisov, line 46: Do we really need the field `pub` and not `pub(crate)`.

glutin/src/display.rs
```diff
@@ -255,6 +257,24 @@ impl Display {
...
+    }
+
+    /// Get a reference to the initialized GLX API.
+    #[cfg(glx_backend)]
+    pub fn glx(&self) -> Option<&'static Glx> {
+        match self {
+            Self::Glx(glx_display) => Some(glx_display.glx()),
+            _ => None,
+        }
+    }
```
kchibisov, line 277: Users can match like that themselves.

glutin_egl_sys/src/egl.rs
```diff
@@ -0,0 +1,249 @@
...
+    if ptr.is_null() {
+        for &sym in fallbacks {
+            ptr = loadfn(sym);
+            if !ptr.is_null() {
+                break;
+            }
+        }
+    }
+    ptr
+}
```
kchibisov, line 249: I'd assume this was taken from the generated code? If so you should document that.

kchibisov (changes requested):
Not sure what to do with wgl...

glutin/src/api/egl/mod.rs
chrisduerr: Well this does load all the extensions which were added in this PR?

It seems odd to generate extensions, but then not load them?

glutin/src/api/egl/mod.rs
chrisduerr: This is what is exposed through the API, so it needs to be public?

Unless you mean the internal field, that can be changed. I just didn't modify it.

glutin/src/display.rs
chrisduerr: If they're forced to always do it anyway, why not offer the abstraction?

glutin/src/api/egl/mod.rs
kchibisov: No, I mean, that you can add extension to `egl_sys` but then don't update `glutin` to load it, so only bump to `egl_sys` will be required...

glutin/src/api/egl/mod.rs
kchibisov: Well, the `field` not the struct, so the `pub SymWrapper` thing.

glutin/src/display.rs
kchibisov: It just won't be consistent, though, maybe it's fine for these particular objects.

glutin/src/api/egl/mod.rs
chrisduerr: Yeah but then the functions will still be exposed, but when you try to call them it won't work? That just seems inconvenient to me from a user perspective.

IF you use the sys crate directly, it's more obvious that you need to load stuff.

glutin/src/display.rs
chrisduerr: What do you mean by "not consistent"?

kchibisov: All specific APIs in glutin are called via matching the backend first, we don't expose `as_egl`, etc on top-level stuff.

Though, maybe we should because it's shorter.

kchibisov: Oh, wait, you do the calls in a global fashion with `extern`, I don't think that it'll work, you'd need to load and allow to null check, because drivers sometimes advertise the extension but the pointer itself is not present...

We'd just need a way to load a pointer with prototype and then the user will use the notation we have interanlly with function pointers and not just by calling the prototypes straight away...

chrisduerr: I'm not sure what you mean? The current version in the PR works just fine for me?

The calls are done through the display.

chrisduerr: I've decided to remove these because for my usecase I can avoid an unnecessary `.unwrap()` by desugaring myself since I only have GLX available.

kchibisov: >I'm not sure what you mean? The current version in the PR works just fine for me?

>The calls are done through the display.

You don't `null-check` the pointer when calling your wayland functions, because even if you have extension the symbol could still be missing for whatever reason.

In glutin, for known functions that tend to bug, we check whether the function actually got loaded in addition to extension check, but in your case it'll just runtime panic, so users won't have an option to check whether something actually got loaded.

kchibisov: So I suggest that we just expose a way to load a specific type of functions, etc, and then user ensures that it actually loaded and call it.

chrisduerr: A user can easily verify if a function is loaded or not with the existing code:

```
glutin_egl_sys::egl::CreateWaylandBufferFromImageWL::is_loaded()
```

So for people that are concerned about this, it still seems entirely possible? There's also nothing stopping them from loading things themselves?

kchibisov (approved):

glutin_egl_sys/src/egl.rs
```diff
@@ -0,0 +1,252 @@
...
+pub const WAYLAND_BUFFER_WL: c_uint = 0x31D5;
+// Accepted in the <attrib_list> parameter of eglCreateImageKHR.
+pub const WAYLAND_PLANE_WL: c_uint = 0x31D6;
+// Possible values for EGL_TEXTURE_FORMAT.
+pub const TEXTURE_Y_U_V_WL: i32 = 0x31D7;
+pub const TEXTURE_Y_UV_WL: i32 = 0x31D8;
+pub const TEXTURE_Y_XUXV_WL: i32 = 0x31D9;
+pub const TEXTURE_EXTERNAL_WL: i32 = 0x31DA;
+// Accepted in the <attribute> parameter of eglQueryWaylandBufferWL.
+pub const EGL_TEXTURE_FORMAT: i32 = 0x3080;
```
MarijnS95, line 161: https://docs.rs/glutin_egl_sys/latest/glutin_egl_sys/egl/constant.TEXTURE_FORMAT.html this already exists, without the `EGL_` prefix?

MarijnS95:
For completeness, since I mentioned updating EGL headers in `gl-generator` above, here's a branch showcasing that everything is still there when generated with a newer generator:

https://github.com/MarijnS95/glutin/compare/gl-generator-update

PR #1670, opened 2024-03-13
Update `raw-window-handle` to v0.6

declantsien:
Only tested on devices using Linux
 None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **Breaking:** updated `raw-window-handle` dependency to `0.6`.
 - Bump MSRV from `1.65` to `1.70`.
 
 # Version 0.31.3
--- a/glutin-winit/Cargo.toml
+++ b/glutin-winit/Cargo.toml
@@ -20,8 +20,8 @@ wayland = ["glutin/wayland", "winit/wayland"]
 
 [dependencies]
 glutin = { version = "0.31.0", path = "../glutin", default-features = false }
-raw-window-handle = "0.5.2"
-winit = { version = "0.29.2", default-features = false, features = ["rwh_05"] }
+raw-window-handle = "0.6"
+winit = { version = "0.29.2", default-features = false, features = ["rwh_06"] }
 
 [build-dependencies]
 cfg_aliases = "0.1.1"
--- a/glutin-winit/src/lib.rs
+++ b/glutin-winit/src/lib.rs
@@ -21,9 +21,9 @@ use glutin::platform::x11::X11GlConfigExt;
 use glutin::prelude::*;
 
 #[cfg(wgl_backend)]
-use raw_window_handle::HasRawWindowHandle;
+use raw_window_handle::HasWindowHandle;
 
-use raw_window_handle::{HasRawDisplayHandle, RawWindowHandle};
+use raw_window_handle::{HasDisplayHandle, RawWindowHandle};
 use winit::error::OsError;
 use winit::event_loop::EventLoopWindowTarget;
 use winit::window::{Window, WindowBuilder};
@@ -104,7 +104,10 @@ impl DisplayBuilder {
         };
 
         #[cfg(wgl_backend)]
-        let raw_window_handle = window.as_ref().map(|window| window.raw_window_handle());
+        let raw_window_handle = window
+            .as_ref()
+            .and_then(|window| window.window_handle().ok())
+            .map(|handle| handle.as_raw());
         #[cfg(not(wgl_backend))]
         let raw_window_handle = None;
 
@@ -170,7 +173,8 @@ fn create_display<T>(
         ApiPreference::FallbackEgl => DisplayApiPreference::WglThenEgl(_raw_window_handle),
     };
 
-    unsafe { Ok(Display::new(window_target.raw_display_handle(), _preference)?) }
+    let handle = window_target.display_handle()?.as_raw();
+    unsafe { Ok(Display::new(handle, _preference)?) }
 }
 
 /// Finalize [`Window`] creation by applying the options from the [`Config`], be
--- a/glutin-winit/src/window.rs
+++ b/glutin-winit/src/window.rs
@@ -5,7 +5,7 @@ use glutin::surface::{
     GlSurface, ResizeableSurface, Surface, SurfaceAttributes, SurfaceAttributesBuilder,
     SurfaceTypeTrait, WindowSurface,
 };
-use raw_window_handle::HasRawWindowHandle;
+use raw_window_handle::{HandleError, HasWindowHandle};
 use winit::window::Window;
 
 /// [`Window`] extensions for working with [`glutin`] surfaces.
@@ -25,7 +25,7 @@ pub trait GlWindow {
     fn build_surface_attributes(
         &self,
         builder: SurfaceAttributesBuilder<WindowSurface>,
-    ) -> SurfaceAttributes<WindowSurface>;
+    ) -> Result<SurfaceAttributes<WindowSurface>, HandleError>;
 
     /// Resize the surface to the window inner size.
     ///
@@ -51,9 +51,10 @@ impl GlWindow for Window {
     fn build_surface_attributes(
         &self,
         builder: SurfaceAttributesBuilder<WindowSurface>,
-    ) -> SurfaceAttributes<WindowSurface> {
+    ) -> Result<SurfaceAttributes<WindowSurface>, HandleError> {
         let (w, h) = self.inner_size().non_zero().expect("invalid zero inner size");
-        builder.build(self.raw_window_handle(), w, h)
+        let handle = self.window_handle()?.as_raw();
+        Ok(builder.build(handle, w, h))
     }
 
     fn resize_surface(
--- a/glutin/src/api/egl/config.rs
+++ b/glutin/src/api/egl/config.rs
@@ -196,8 +196,8 @@ impl Display {
                 // XXX This can't be done by passing visual in the EGL attributes
                 // when calling `eglChooseConfig` since the visual is ignored.
                 match template.native_window {
-                    Some(RawWindowHandle::Xcb(xcb)) if xcb.visual_id > 0 => {
-                        xcb.visual_id as u32 == config.native_visual()
+                    Some(RawWindowHandle::Xcb(xcb)) => {
+                        xcb.visual_id.map_or(false, |id| id.get() == config.native_visual())
                     },
                     Some(RawWindowHandle::Xlib(xlib)) if xlib.visual_id > 0 => {
                         xlib.visual_id as u32 == config.native_visual()
@@ -386,7 +386,7 @@ impl X11GlConfigExt for Config {
         match *self.inner.display.inner._native_display? {
             raw_window_handle::RawDisplayHandle::Xlib(display_handle) => unsafe {
                 let xid = self.native_visual();
-                X11VisualInfo::from_xid(display_handle.display as *mut _, xid as _)
+                X11VisualInfo::from_xid(display_handle.display?.as_ptr() as *mut _, xid as _)
             },
             _ => None,
         }
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -14,6 +14,8 @@ use glutin_egl_sys::egl::types::{EGLAttrib, EGLDisplay, EGLint};
 use once_cell::sync::OnceCell;
 
 use raw_window_handle::RawDisplayHandle;
+#[cfg(x11_platform)]
+use raw_window_handle::XlibDisplayHandle;
 
 use crate::config::ConfigTemplate;
 use crate::context::Version;
@@ -239,21 +241,24 @@ impl Display {
         let extensions = CLIENT_EXTENSIONS.get().unwrap();
 
         let mut attrs = Vec::<EGLAttrib>::with_capacity(5);
-        let (platform, mut display) = match display {
+        let (platform, display) = match display {
             #[cfg(wayland_platform)]
             RawDisplayHandle::Wayland(handle)
                 if extensions.contains("EGL_KHR_platform_wayland") =>
             {
-                (egl::PLATFORM_WAYLAND_KHR, handle.display)
+                (egl::PLATFORM_WAYLAND_KHR, handle.display.as_ptr())
             },
             #[cfg(x11_platform)]
             RawDisplayHandle::Xlib(handle) if extensions.contains("EGL_KHR_platform_x11") => {
                 attrs.push(egl::PLATFORM_X11_SCREEN_KHR as EGLAttrib);
                 attrs.push(handle.screen as EGLAttrib);
-                (egl::PLATFORM_X11_KHR, handle.display)
+                (
+                    egl::PLATFORM_X11_KHR,
+                    handle.display.map_or(egl::DEFAULT_DISPLAY as *mut _, |d| d.as_ptr()),
+                )
             },
             RawDisplayHandle::Gbm(handle) if extensions.contains("EGL_KHR_platform_gbm") => {
-                (egl::PLATFORM_GBM_KHR, handle.gbm_device)
+                (egl::PLATFORM_GBM_KHR, handle.gbm_device.as_ptr())
             },
             RawDisplayHandle::Android(_) if extensions.contains("EGL_KHR_platform_android") => {
                 (egl::PLATFORM_ANDROID_KHR, egl::DEFAULT_DISPLAY as *mut _)
@@ -264,11 +269,6 @@ impl Display {
                 )
             },
         };
-
-        // Be explicit here.
-        if display.is_null() {
-            display = egl::DEFAULT_DISPLAY as *mut _;
-        }
 
         // Push at the end so we can pop it on failure
         let mut has_display_reference = extensions.contains("EGL_KHR_display_reference");
@@ -315,18 +315,21 @@ impl Display {
 
         let mut attrs = Vec::<EGLint>::with_capacity(5);
         let mut legacy = false;
-        let (platform, mut display) = match display {
+        let (platform, display) = match display {
             #[cfg(wayland_platform)]
             RawDisplayHandle::Wayland(handle)
                 if extensions.contains("EGL_EXT_platform_wayland") =>
             {
-                (egl::PLATFORM_WAYLAND_EXT, handle.display)
+                (egl::PLATFORM_WAYLAND_EXT, handle.display.as_ptr())
             },
             #[cfg(x11_platform)]
             RawDisplayHandle::Xlib(handle) if extensions.contains("EGL_EXT_platform_x11") => {
                 attrs.push(egl::PLATFORM_X11_SCREEN_EXT as EGLint);
                 attrs.push(handle.screen as EGLint);
-                (egl::PLATFORM_X11_EXT, handle.display)
+                (
+                    egl::PLATFORM_X11_EXT,
+                    handle.display.map_or(egl::DEFAULT_DISPLAY as *mut _, |d| d.as_ptr()),
+                )
             },
             #[cfg(x11_platform)]
             RawDisplayHandle::Xcb(handle)
@@ -335,10 +338,13 @@ impl Display {
             {
                 attrs.push(egl::PLATFORM_XCB_SCREEN_EXT as EGLint);
                 attrs.push(handle.screen as EGLint);
-                (egl::PLATFORM_XCB_EXT, handle.connection)
+                (
+                    egl::PLATFORM_XCB_EXT,
+                    handle.connection.map_or(egl::DEFAULT_DISPLAY as *mut _, |c| c.as_ptr()),
+                )
             },
             RawDisplayHandle::Gbm(handle) if extensions.contains("EGL_MESA_platform_gbm") => {
-                (egl::PLATFORM_GBM_MESA, handle.gbm_device)
+                (egl::PLATFORM_GBM_MESA, handle.gbm_device.as_ptr())
             },
             RawDisplayHandle::Windows(..) if extensions.contains("EGL_ANGLE_platform_angle") => {
                 // Only CreateWindowSurface appears to work with Angle.
@@ -351,11 +357,6 @@ impl Display {
                 )
             },
         };
-
-        // Be explicit here.
-        if display.is_null() {
-            display = egl::DEFAULT_DISPLAY as *mut _;
-        }
 
         // Push at the end so we can pop it on failure
         let mut has_display_reference = extensions.contains("EGL_KHR_display_reference");
@@ -404,10 +405,12 @@ impl Display {
     }
 
     fn get_display(egl: &Egl, display: RawDisplayHandle) -> Result<EglDisplay> {
-        let mut display = match display {
-            RawDisplayHandle::Gbm(handle) => handle.gbm_device,
+        let display = match display {
+            RawDisplayHandle::Gbm(handle) => handle.gbm_device.as_ptr(),
             #[cfg(x11_platform)]
-            RawDisplayHandle::Xlib(handle) => handle.display,
+            RawDisplayHandle::Xlib(XlibDisplayHandle { display, .. }) => {
+                display.map_or(egl::DEFAULT_DISPLAY as *mut _, |d| d.as_ptr())
+            },
             RawDisplayHandle::Android(_) => egl::DEFAULT_DISPLAY as *mut _,
             _ => {
                 return Err(
@@ -415,10 +418,6 @@ impl Display {
                 )
             },
         };
-
-        if display.is_null() {
-            display = egl::DEFAULT_DISPLAY as *mut _;
-        }
 
         let display = unsafe { egl.GetDisplay(display) };
         Self::check_display_error(display).map(EglDisplay::Legacy)
--- a/glutin/src/api/egl/surface.rs
+++ b/glutin/src/api/egl/surface.rs
@@ -472,14 +472,10 @@ impl NativeWindow {
         let native_window = match raw_window_handle {
             #[cfg(wayland_platform)]
             RawWindowHandle::Wayland(window_handle) => unsafe {
-                if window_handle.surface.is_null() {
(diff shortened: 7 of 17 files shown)


kchibisov:
We already have #1582 for that, though we have an option for `0.6` just for the time being unless I have time to look into the safe handle stuff and how it could be approached in glutin.

glutin/src/api/egl/config.rs
```diff
@@ -196,8 +196,10 @@ impl Display {
                 // XXX This can't be done by passing visual in the EGL attributes
                 // when calling `eglChooseConfig` since the visual is ignored.
                 match template.native_window {
-                    Some(RawWindowHandle::Xcb(xcb)) if xcb.visual_id > 0 => {
-                        xcb.visual_id as u32 == config.native_visual()
+                    Some(RawWindowHandle::Xcb(xcb))
+                        if xcb.visual_id.is_some() && xcb.visual_id.unwrap().get() > 0 =>
```
kchibisov, line 200: Do not unwrap here, just make it default to zero and follow the old logic.

glutin/src/api/egl/display.rs
```diff
@@ -244,16 +244,18 @@ impl Display {
...
             },
             #[cfg(x11_platform)]
-            RawDisplayHandle::Xlib(handle) if extensions.contains("EGL_KHR_platform_x11") => {
+            RawDisplayHandle::Xlib(handle)
+                if extensions.contains("EGL_KHR_platform_x11") && handle.display.is_some() =>
+            {
                 attrs.push(egl::PLATFORM_X11_SCREEN_KHR as EGLAttrib);
                 attrs.push(handle.screen as EGLAttrib);
-                (egl::PLATFORM_X11_KHR, handle.display)
+                (egl::PLATFORM_X11_KHR, handle.display.unwrap().as_ptr())
```
kchibisov, line 255: Do not unwrap here. IIRC passing `null` is fine and it'll just open a default display or error, we shouldn't crash.

glutin/src/api/egl/display.rs
```diff
@@ -320,25 +322,28 @@ impl Display {
...
             },
             #[cfg(x11_platform)]
-            RawDisplayHandle::Xlib(handle) if extensions.contains("EGL_EXT_platform_x11") => {
+            RawDisplayHandle::Xlib(handle)
+                if extensions.contains("EGL_EXT_platform_x11") && handle.display.is_some() =>
+            {
                 attrs.push(egl::PLATFORM_X11_SCREEN_EXT as EGLint);
                 attrs.push(handle.screen as EGLint);
-                (egl::PLATFORM_X11_EXT, handle.display)
+                (egl::PLATFORM_X11_EXT, handle.display.unwrap().as_ptr())
```
kchibisov, line 333: Same about `unwrap`.

glutin/src/api/egl/display.rs
```diff
@@ -320,25 +322,28 @@ impl Display {
...
-                if extensions.contains("EGL_MESA_platform_xcb")
-                    || extensions.contains("EGL_EXT_platform_xcb") =>
+                if (extensions.contains("EGL_MESA_platform_xcb")
+                    || extensions.contains("EGL_EXT_platform_xcb"))
+                    && handle.connection.is_some() =>
             {
                 attrs.push(egl::PLATFORM_XCB_SCREEN_EXT as EGLint);
                 attrs.push(handle.screen as EGLint);
-                (egl::PLATFORM_XCB_EXT, handle.connection)
+                (egl::PLATFORM_XCB_EXT, handle.connection.unwrap().as_ptr())
```
kchibisov, line 343: I don't remember the semantics here, but I think `null` is also fine here. In general I'd suggest to test.

glutin/src/api/glx/display.rs
```diff
@@ -45,11 +45,11 @@ impl Display {
         // Don't load GLX when unsupported platform was requested.
         let (display, screen) = match display {
             RawDisplayHandle::Xlib(handle) => {
-                if handle.display.is_null() {
+                if handle.display.is_none() {
                     return Err(ErrorKind::BadDisplay.into());
                 }
 
-                (GlxDisplay(handle.display as *mut _), handle.screen as i32)
+                (GlxDisplay(handle.display.unwrap().as_ptr() as *mut _), handle.screen as i32)
```
kchibisov, line 52: refactor it `if let else`

glutin/src/api/wgl/display.rs
```diff
@@ -59,9 +59,14 @@ impl Display {
...
         let (wgl_extra, client_extensions) =
             if let Some(RawWindowHandle::Win32(window)) = native_window {
+                if window.hinstance.is_none() {
+                    return Err(ErrorKind::BadDisplay.into());
+                }
                 unsafe {
-                    let (wgl_extra, client_extensions) =
-                        super::load_extra_functions(window.hinstance as _, window.hwnd as _)?;
+                    let (wgl_extra, client_extensions) = super::load_extra_functions(
+                        window.hinstance.unwrap().get() as _,
```
kchibisov, line 67: I don't remember the semantics when `0` is passed. Would suggest to check, probably it's also _fine_

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Unreleased
+- Updated raw-window-handle dependency to 0.6. See [raw-window-handle CHANGELOG](https://github.com/rust-windowing/raw-window-handle/blob/v0.6.0/CHANGELOG.md#060-2023-09-30) for more info.
```
kchibisov, line 2:
```suggestion

- **Breaking:** updated raw-window-handle dependency to 0.6.
```

glutin-winit/src/window.rs
```diff
@@ -53,7 +53,10 @@ impl GlWindow for Window {
         builder: SurfaceAttributesBuilder<WindowSurface>,
     ) -> SurfaceAttributes<WindowSurface> {
         let (w, h) = self.inner_size().non_zero().expect("invalid zero inner size");
-        builder.build(self.raw_window_handle(), w, h)
+        match self.window_handle() {
+            Ok(handle) => builder.build(handle.as_raw(), w, h),
+            Err(err) => panic!("Error {err:?}"),
```
kchibisov, line 58: We shouldn't panic here, probably it's better to propagate.

kchibisov (changes requested):

kchibisov:
I'd just post a review, since there's nothing wrong with just using `raw` version of 0.6, since it's just small update.

MarijnS95:
A lot of changes here may snoop the implementation from https://github.com/ash-rs/ash/pull/799 which - to my knowledge - treats (non)-nullable parameters accordingly.

glutin/src/api/egl/config.rs
declantsien: If I understand correctly
Compare `xcb.visual_id` with `config.native_visual` when `xcb.visual.id > 0`,
anyother case return false.

glutin/src/api/egl/display.rs
declantsien: Ok return `ptr::null_mut()` if None

glutin/src/api/glx/display.rs
declantsien: I can't make `if let else` working here. We are setting inline variable here.
Anyway hopes `match` fits here.

glutin/src/api/egl/display.rs
```diff
@@ -244,16 +244,19 @@ impl Display {
...
+                (egl::PLATFORM_WAYLAND_KHR, handle.display.as_ptr())
             },
             #[cfg(x11_platform)]
             RawDisplayHandle::Xlib(handle) if extensions.contains("EGL_KHR_platform_x11") => {
                 attrs.push(egl::PLATFORM_X11_SCREEN_KHR as EGLAttrib);
                 attrs.push(handle.screen as EGLAttrib);
-                (egl::PLATFORM_X11_KHR, handle.display)
+                (
+                    egl::PLATFORM_X11_KHR,
+                    handle.display.map(|d| d.as_ptr()).unwrap_or(ptr::null_mut()),
```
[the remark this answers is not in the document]
MarijnS95: There's a check below that turns this into `egl::DEFAULT_DISPLAY` (whose value is `0`) but it's also set explicitly in Android :)

(Maybe this section of code needs some cleanup to read more logically)

glutin/src/api/egl/display.rs
```diff
@@ -405,9 +414,11 @@ impl Display {
 
     fn get_display(egl: &Egl, display: RawDisplayHandle) -> Result<EglDisplay> {
         let mut display = match display {
-            RawDisplayHandle::Gbm(handle) => handle.gbm_device,
+            RawDisplayHandle::Gbm(handle) => handle.gbm_device.as_ptr(),
             #[cfg(x11_platform)]
-            RawDisplayHandle::Xlib(handle) => handle.display,
+            RawDisplayHandle::Xlib(handle) if handle.display.is_some() => {
+                handle.display.unwrap().as_ptr()
+            },
```
MarijnS95, line 421: It might be cleaner to write this `unwrap()` as an explicit `Some` match:

```suggestion
            RawDisplayHandle::Xlib(XlibDisplayHandle { display: Some(display), .. }) => {
                display.as_ptr()
            },
```

glutin/src/api/egl/surface.rs
```diff
@@ -498,36 +494,20 @@ impl NativeWindow {
             },
             #[cfg(x11_platform)]
             RawWindowHandle::Xcb(window_handle) => {
-                if window_handle.window == 0 {
+                if window_handle.window.get() == 0 {
                     return Err(ErrorKind::BadNativeWindow.into());
                 }
```
MarijnS95, line 499: Isn't this a `NonZeroU32`? This check makes no sense.

```suggestion
```

glutin_examples/src/lib.rs
```diff
@@ -53,7 +53,9 @@ pub fn main(event_loop: winit::event_loop::EventLoop<()>) -> Result<(), Box<dyn
 
     println!("Picked a config with {} samples", gl_config.num_samples());
 
-    let raw_window_handle = window.as_ref().map(|window| window.raw_window_handle());
+    let raw_window_handle = window
+        .as_ref()
+        .and_then(|window| window.window_handle().map(|handle| handle.as_raw()).ok());
```
MarijnS95, line 58: Same here.

glutin_examples/src/lib.rs
```diff
@@ -102,7 +104,10 @@ pub fn main(event_loop: winit::event_loop::EventLoop<()>) -> Result<(), Box<dyn
                         .unwrap()
                 });
 
-                let attrs = window.build_surface_attributes(Default::default());
+                let attrs = match window.build_surface_attributes(Default::default()) {
+                    Ok(attrs) => attrs,
+                    Err(err) => panic!("Error: {err:?}"),
```
MarijnS95, line 109: Same here, even though it's just an example.

glutin/src/api/egl/config.rs
```diff
@@ -196,9 +196,11 @@ impl Display {
                 // XXX This can't be done by passing visual in the EGL attributes
                 // when calling `eglChooseConfig` since the visual is ignored.
                 match template.native_window {
-                    Some(RawWindowHandle::Xcb(xcb)) if xcb.visual_id > 0 => {
-                        xcb.visual_id as u32 == config.native_visual()
-                    },
+                    Some(RawWindowHandle::Xcb(xcb)) => xcb
+                        .visual_id
+                        .map(|id| id.get() as u32)
+                        .map(|id| id > 0 && id == config.native_visual())
```
MarijnS95, line 202: This is already a `NonZeroU32` (otherwise `visual_id` is `None`)

glutin/src/api/egl/display.rs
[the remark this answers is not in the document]
declantsien: > There's a check below that turns this into egl::DEFAULT_DISPLAY (whose value is 0) but it's also set explicitly in Android :)

[AndroidDisplayHandle](https://docs.rs/raw-window-handle/latest/raw_window_handle/struct.AndroidDisplayHandle.html) is empty, not sure what are you asking for...

glutin_examples/src/lib.rs
declantsien: Nah, this will not. I saw an `unwrap()` just above the line. It's inside the closure for `event_loop`

sarthakvk:
Is there a plan to merge this PR?

kchibisov:
In general, re-request the review with github UI when it's ready or explicitly indicate the state if you're not interested anymore.

kchibisov (approved):

glutin-winit/src/lib.rs
```diff
@@ -170,7 +173,8 @@ fn create_display<T>(
         ApiPreference::FallbackEgl => DisplayApiPreference::WglThenEgl(_raw_window_handle),
     };
 
-    unsafe { Ok(Display::new(window_target.raw_display_handle(), _preference)?) }
+    let handle = window_target.display_handle()?;
+    unsafe { Ok(Display::new(handle.as_raw(), _preference)?) }
```
MarijnS95, line 177: Nit: `let handle` is different here than in `build_surface_attributes()`.

glutin/src/api/egl/display.rs
```diff
@@ -405,9 +407,9 @@ impl Display {
 
     fn get_display(egl: &Egl, display: RawDisplayHandle) -> Result<EglDisplay> {
         let mut display = match display {
-            RawDisplayHandle::Gbm(handle) => handle.gbm_device,
+            RawDisplayHandle::Gbm(handle) => handle.gbm_device.as_ptr(),
             #[cfg(x11_platform)]
-            RawDisplayHandle::Xlib(handle) => handle.display,
+            RawDisplayHandle::Xlib(XlibDisplayHandle { display: Some(d), .. }) => d.as_ptr(),
```
MarijnS95, line 412: Previously setting `display` to `null`/`None` would cause `egl::DEFAULT_DISPLAY` to be used below, now this causes it to return `NotSupported`?

If there's an explicit case to be made that no default display exists for `Xlib`, that should be documented as the previous code was allowing this.

glutin/src/api/egl/display.rs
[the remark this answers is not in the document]
MarijnS95: Asking to make the code flow less silly/hard-to-read.

glutin/src/api/wgl/display.rs
```diff
@@ -60,8 +60,10 @@ impl Display {
         let (wgl_extra, client_extensions) =
             if let Some(RawWindowHandle::Win32(window)) = native_window {
                 unsafe {
-                    let (wgl_extra, client_extensions) =
-                        super::load_extra_functions(window.hinstance as _, window.hwnd as _)?;
+                    let (wgl_extra, client_extensions) = super::load_extra_functions(
+                        window.hinstance.map(|i| i.get()).unwrap_or(0) as _,
```
MarijnS95, line 64: `unwrap_or_else`?

I'm unfamiliar with this code to say if it's okay to pass `0` for `HINSTANCE`. There are multiple getters but don't know if any should be used.

glutin_examples/examples/switch_render_thread.rs
```diff
@@ -191,7 +193,10 @@ fn create_window_with_render_context(
             .expect("failed to create context")
     };
 
-    let attrs = window.build_surface_attributes(<_>::default());
+    let attrs = match window.build_surface_attributes(<_>::default()) {
+        Ok(attrs) => attrs,
+        Err(err) => panic!("Error {err:?}"),
```
MarijnS95, line 198: That's equivalent to `.expect("Error")` (except that you don't expect there to be an error, but a more sensible message).

glutin_examples/src/lib.rs
MarijnS95: Then make it a proper `.expect()` with a clear message.

MarijnS95 (commented):
Overall seeing a bit too many semantical changes that I'm not immediately content with. `raw-window-handle 0.6` gives us the ability to know that values are intentionally `0`/`null()` by setting an `Option` to `None`, but we're not taking this opportunity to validate whether the underlying API can and should handle that value, or that we should bail. In some cases there are unwraps, or `None` propagation, or `?` propagation making things harder to follow and validate.

glutin/src/api/wgl/display.rs
kchibisov: it's the same as before, because it could be 0, which is default for rwh 0.5 and lower.

glutin/src/api/egl/display.rs
declantsien: Nice catch

declantsien:
@kchibisov Can you help with the CI error. Wired cfg errors.

declantsien:
> @kchibisov Can you help with the CI error. Wired cfg errors.

The error is from the rust nightly. Tried with an old build `nightly-2024-04-05` the errors are gone. Since we don't have a `rust-toolchain.toml` file, no idea how to deal with it.
Another question, why don't we use Cargo.lock? Cargo.lock is useful for reproducible build.

kchibisov:
>Another question, why don't we use Cargo.lock? Cargo.lock is useful for reproducible build.

The old default behavior for rust was to not include `Cargo.lock` for libraries. They changed that to include last year iirc.

kchibisov (approved):
Should be good after rebase.

declantsien:
> Should be good after rebase.

Nice

declantsien:
> > Another question, why don't we use Cargo.lock? Cargo.lock is useful for reproducible build.
> 
> The old default behavior for rust was to not include `Cargo.lock` for libraries. They changed that to include last year iirc.

Right. Good to know. Thanks.

MarijnS95:
> Another question, why don't we use Cargo.lock? Cargo.lock is useful for reproducible build.

A reproducible build (for libraries) is "useless" when a published crate is on crates.io. Only semver version (range) dependencies from `Cargo.toml` are taken into account when your crate is used in another project. `cargo` doesn't magically "merge" its `Cargo.lock` with the lockfile of the current project:

https://doc.rust-lang.org/cargo/faq.html#:~:text=However%2C%20this%20determinism%20can%20give%20a%20false%20sense%20of%20security%20because%20Cargo.lock%20does%20not%20affect%20the%20consumers%20of%20your%20package%2C%20only%20Cargo.toml%20does%20that.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Bump MSRV from `1.65` to `1.70`.
+- **Breaking:** updated `raw-window-handle` dependency to `0.6.1`.
```
MarijnS95, line 4: Question for @kchibisov: the changelog seems to have a mix of past- and present-tense. What should we pick?

glutin-winit/Cargo.toml
```diff
@@ -20,8 +20,8 @@ wayland = ["glutin/wayland", "winit/wayland"]
 
 [dependencies]
 glutin = { version = "0.31.0", path = "../glutin", default-features = false }
-raw-window-handle = "0.5.2"
-winit = { version = "0.29.2", default-features = false, features = ["rwh_05"] }
+raw-window-handle = "0.6.1"
```
MarijnS95, line 23: Is there anything new in the `0.6.1` release that is not in the `0.6.0` release, that is required by `glutin`?

glutin/src/api/egl/display.rs
```diff
@@ -405,9 +407,11 @@ impl Display {
 
     fn get_display(egl: &Egl, display: RawDisplayHandle) -> Result<EglDisplay> {
         let mut display = match display {
-            RawDisplayHandle::Gbm(handle) => handle.gbm_device,
+            RawDisplayHandle::Gbm(handle) => handle.gbm_device.as_ptr(),
             #[cfg(x11_platform)]
-            RawDisplayHandle::Xlib(handle) => handle.display,
+            RawDisplayHandle::Xlib(XlibDisplayHandle { display, .. }) => {
+                display.map_or(egl::DEFAULT_DISPLAY as *mut _, |d| d.as_ptr())
```
MarijnS95, line 413: Fwiw I meant to just map it to `null()` since this special value is selected below if it was `is_null()`.

MarijnS95: But then Android could have done the same. As said in earlier review, this function is a tad messy.

MarijnS95: In fact, since `gbm_device` is `NonNull`, I'd request to remove the now-unreachable thing below:

```rust
        if display.is_null() {
            display = egl::DEFAULT_DISPLAY as *mut _;
        }
```

MarijnS95 (approved):
One more cleanup request for an unreachable `if display.is_null()`, but seems like we're getting somewhere otherwise.

glutin/src/api/wgl/display.rs
MarijnS95: @kchibisov yes that is the point: it could have been passed before, but now that we _know_ the `raw-window-handle` API _allows zero_, we should take a second to realize whether that is actually an allowed case.

MarijnS95: _Our own_ safety docs already say that this must be "valid":

https://github.com/rust-windowing/glutin/blob/052399de0a4a7237b4dc12dd0fcdeb63f1c2af41/glutin/src/api/wgl/display.rs#L40-L44

So I think this should be a `panic!()`/`.unwrap()`/`ok_or...()?`.

glutin/src/api/egl/display.rs
```diff
@@ -265,11 +267,6 @@ impl Display {
             },
         };
 
-        // Be explicit here.
-        if display.is_null() {
-            display = egl::DEFAULT_DISPLAY as *mut _;
```
MarijnS95, line 270: Not this one, unless you change all the `.map_or()`'s above to return `egl::DEFAULT_DISPLAY` instead.

glutin/src/api/egl/display.rs
```diff
@@ -405,9 +407,11 @@ impl Display {
 
     fn get_display(egl: &Egl, display: RawDisplayHandle) -> Result<EglDisplay> {
         let mut display = match display {
-            RawDisplayHandle::Gbm(handle) => handle.gbm_device,
+            RawDisplayHandle::Gbm(handle) => handle.gbm_device.as_ptr(),
             #[cfg(x11_platform)]
-            RawDisplayHandle::Xlib(handle) => handle.display,
+            RawDisplayHandle::Xlib(XlibDisplayHandle { display, .. }) => {
+                display.map_or(std::ptr::null(), |d| d.as_ptr())
```
MarijnS95, line 413: :shrug: 

```suggestion
                display.map_or(egl::DEFAULT_DISPLAY as *mut _, |d| d.as_ptr())
```

glutin/src/api/egl/display.rs
```diff
@@ -320,13 +322,13 @@ impl Display {
...
             {
-                (egl::PLATFORM_WAYLAND_EXT, handle.display)
+                (egl::PLATFORM_WAYLAND_EXT, handle.display.as_ptr())
             },
             #[cfg(x11_platform)]
             RawDisplayHandle::Xlib(handle) if extensions.contains("EGL_EXT_platform_x11") => {
                 attrs.push(egl::PLATFORM_X11_SCREEN_EXT as EGLint);
                 attrs.push(handle.screen as EGLint);
-                (egl::PLATFORM_X11_EXT, handle.display)
+                (egl::PLATFORM_X11_EXT, handle.display.map_or(ptr::null_mut(), |d| d.as_ptr()))
```
MarijnS95, line 331: `if display.is_null()` can be removed if you set `egl::DEFAULT_DISPLAY as *mut _` here... 1/2

glutin/src/api/egl/display.rs
```diff
@@ -335,10 +337,10 @@ impl Display {
             {
                 attrs.push(egl::PLATFORM_XCB_SCREEN_EXT as EGLint);
                 attrs.push(handle.screen as EGLint);
-                (egl::PLATFORM_XCB_EXT, handle.connection)
+                (egl::PLATFORM_XCB_EXT, handle.connection.map_or(ptr::null_mut(), |c| c.as_ptr()))
```
MarijnS95, line 340: And `egl::DEFAULT_DISPLAY as *mut _` here 2/2.

MarijnS95:
One item remaining, the rest looks good:

- https://github.com/rust-windowing/glutin/pull/1670#discussion_r1590681604

glutin/src/api/wgl/display.rs
declantsien: > So I think this should be a `panic!()`/`.unwrap()`/`ok_or...()?`.

I go with `.unwrap()`.

declantsien:
> One item remaining, the rest looks good:
> 
>     * [Update `raw-window-handle` to v0.6 #1670 (comment)](https://github.com/rust-windowing/glutin/pull/1670#discussion_r1590681604)

@MarijnS95 I think I am done with your requests.

glutin/src/api/wgl/display.rs
MarijnS95: Yes that is a good idea. If people were previously implicitly passing `0` (mismatching the `# Safety` docs), we'll catch that and can always amend the code with a new patch release.

glutin_examples/Cargo.toml
```diff
@@ -22,11 +22,11 @@ wayland = ["glutin-winit/wayland", "winit/wayland-dlopen", "winit/wayland-csd-ad
 glutin = { path = "../glutin", default-features = false }
 glutin-winit = { path = "../glutin-winit", default-features = false }
 png = { version = "0.17.6", optional = true }
-raw-window-handle = "0.5"
+raw-window-handle = "0.6"
 winit = { version = "0.29.2", default-features = false, features = ["rwh_05"] }
```
marc2332, line 26: Any reason this version of `winit` was left unchanged with `rwh_05` instead of `rwh_06` ?

I just noticed it when resolving conflicts in https://github.com/rust-windowing/glutin/pull/1675

kchibisov: it's example, doesn't really matter. Could use 06.

marc2332: Yeah, seems to work just fine, I simply found it odd 😄

PR #1680, opened 2024-05-28
build: `windows-sys` 0.48.0 → 0.52.0

ErichDonGubler:
- [x] ~~Created or updated an example program if it would help users understand this functionality~~


MarijnS95:
What does `build` mean in the title? Isn't this part of the `glutin` runtime crate?

ErichDonGubler:
@MarijnS95: Ah, that's a convention from certain communities that use Conventional Commits; [this cheat sheet](https://kapeli.com/cheat_sheets/Conventional_Commits.docset/Contents/Resources/Documents/index) describes well how I've seen it used, which is:

> Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)

I can alter or remove the prefix if you want. Just LMK!

MarijnS95:
I've never liked "conventional commits" for e.g. this reason. What does "build" have to do with "external dependencies"?

ErichDonGubler:
I mean, Rust developers in particular quite literally `cargo build` dependencies into a binary. It doesn't sound conceptually distant to me, but I understand if you prefer a different prefix (or a system of authoring history outside of Conventional Commits).

My offer to change the commit message still stands.

kchibisov (approved):

kchibisov:
Generally, the title prefix should reflect the subsystem, in this case it was primary wgl.

PR #1681, opened 2024-06-06
wgl: add GLUTIN_WGL_OPENGL_DLL

### LINKED ISSUE #1671: Add a variable to set opengl.dll filename on WGL
Can a variable be added to set filename? Like `QT_OPENGL_DLL`.
https://github.com/rust-windowing/glutin/blob/92281268669b4e5db4569b92ff23b6d2c01cd325/glutin/src/api/wgl/display.rs#L53

kchibisov:
Fixes #1671.


MarijnS95:
@kchibisov sidenote: this PR says that the source branch was removed, but it seems to still be there. Is that expected?

PR #1683, opened 2024-06-07
egl_sys: Drop duplicate manual definition of `(EGL_)TEXTURE_FORMAT`

MarijnS95:
For https://github.com/rust-windowing/glutin/pull/1668#discussion_r1631607958

The generated code already defines `TEXTURE_FORMAT` to `0x3080` with the usual `EGL_` prefix stripped; we shouldn't be redefining it (with this unnecessary prefix) in manual code.


MarijnS95:
Right above there's a comment:

```rust
// Possible values for EGL_TEXTURE_FORMAT.
```

I'm carefully considering to rename that to `glutin_egl_sys::TEXTURE_FORMAT`. Thoughts?

MarijnS95:
I think that there's only `EGL_TEXTURE_FORMAT`. As with everything that's output by `gl-generator`, the common `EGL_` prefix is stripped off, and @chrisduerr in #1668 simply missed that it was already available in the autogenerated bindings? The value is the same too.

kchibisov (approved):

chrisduerr (approved):

PR #1686, opened 2024-07-01
Fix logic in egl Device::query_device

quaternic:
The previous test would mistakenly return an error if the extensions contain "EGL_EXT_device_base" but not the two others. As stated by the comment, and in https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_device_base.txt , it is supposed to contain the other two.

The cases that should not return an error are:
 - contains EGL_EXT_device_base (`B`)
 - contains both EGL_EXT_device_enumeration (`E`) and EGL_EXT_device_query (`Q`)
 
The overall effects of this patch in terms of those three extensions are:
   -  `!B &  E & !Q` leads to a different error message than before
   - ` B & !E & !Q`  no longer errors
   - `!B &  E &  Q`  no longer errors

### DIFF
--- a/glutin/src/api/egl/device.rs
+++ b/glutin/src/api/egl/device.rs
@@ -41,9 +41,9 @@ impl Device {
         //
         // Or we can check for the EGL_EXT_device_base extension since it contains both
         // extensions.
-        if (!no_display_extensions.contains("EGL_EXT_device_enumeration")
-            && !no_display_extensions.contains("EGL_EXT_device_query"))
-            || !no_display_extensions.contains("EGL_EXT_device_base")
+        if !no_display_extensions.contains("EGL_EXT_device_base")
+        && ( !no_display_extensions.contains("EGL_EXT_device_enumeration")
+          || !no_display_extensions.contains("EGL_EXT_device_query"))
         {
             return Err(ErrorKind::NotSupported("EGL does not support EGL_EXT_device_base").into());
         }
--- a/glutin/src/api/egl/device.rs
+++ b/glutin/src/api/egl/device.rs
@@ -4,7 +4,6 @@ use std::collections::HashSet;
 use std::ffi::{c_void, CStr};
 use std::ptr;
 
-use bitflags::Flags;
 use glutin_egl_sys::egl;
 use glutin_egl_sys::egl::types::EGLDeviceEXT;
 
--- a/glutin/src/api/egl/device.rs
+++ b/glutin/src/api/egl/device.rs
@@ -37,19 +37,20 @@ impl Device {
             CLIENT_EXTENSIONS.get_or_init(|| get_extensions(egl, egl::NO_DISPLAY));
 
         // Querying devices requires EGL_EXT_device_enumeration or EGL_EXT_device_base.
-        //
-        // The former depends on EGL_EXT_device_query, so also check that just in case.
         if !client_extensions.contains("EGL_EXT_device_base") {
-            let query = client_extensions.contains("EGL_EXT_device_query");
-            let enumr = client_extensions.contains("EGL_EXT_device_enumeration");
-            if !(query && enumr) {
-                let msg = if !enumr {
-                    "Enumerating devices is not supported by the EGL instance"
-                } else {
-                    // This should not happen as device_enumeration depends on device_query.
-                    "EGL_EXT_device_enumeration without EGL_EXT_device_query, buggy driver?"
-                };
-                return Err(ErrorKind::NotSupported(msg).into());
+            if !client_extensions.contains("EGL_EXT_device_enumeration") {
+                return Err(ErrorKind::NotSupported(
+                    "Enumerating devices is not supported by the EGL instance",
+                )
+                .into());
+            }
+            // EGL_EXT_device_enumeration depends on EGL_EXT_device_query,
+            // so also check that just in case.
+            if !client_extensions.contains("EGL_EXT_device_query") {
+                return Err(ErrorKind::NotSupported(
+                    "EGL_EXT_device_enumeration without EGL_EXT_device_query, buggy driver?",
+                )
+                .into());
             }
         }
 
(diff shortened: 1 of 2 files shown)


kchibisov:
cc @i509VCB I think the test was similar to the one in smithay?

kchibisov (changes requested):
Should add a CHANGELOG entry.

i509VCB:
> The previous test would mistakenly return an error if the extensions contain "EGL_EXT_device_base" but not the two others. As stated by the comment, and in https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_device_base.txt , it is supposed to contain the other two.
>
> The cases that should not return an error are:
> 
> * contains EGL_EXT_device_base
> * contains both EGL_EXT_device_enumeration and EGL_EXT_device_query

The changes look correct the specification for EGL_EXT_device_base states the following:

```
Add to section "3.2 Devices"

    "EGL_EXT_device_base is equivalent to the combination of the
    functionality defined by EGL_EXT_device_query and
    EGL_EXT_device_enumeration."
```

glutin/src/api/egl/device.rs
```diff
@@ -41,9 +41,9 @@ impl Device {
...
         // Or we can check for the EGL_EXT_device_base extension since it contains both
         // extensions.
-        if (!no_display_extensions.contains("EGL_EXT_device_enumeration")
-            && !no_display_extensions.contains("EGL_EXT_device_query"))
-            || !no_display_extensions.contains("EGL_EXT_device_base")
+        if !no_display_extensions.contains("EGL_EXT_device_base")
+            && (!no_display_extensions.contains("EGL_EXT_device_enumeration")
+                || !no_display_extensions.contains("EGL_EXT_device_query"))
         {
             return Err(ErrorKind::NotSupported("EGL does not support EGL_EXT_device_base").into());
```
i509VCB, line 48: This error is not correct, maybe just say enumerating/querying devices is unsupported.

quaternic: It seems that `EGL_EXT_device_enumeration` actually requires `EGL_EXT_device_query`, so I believe this code doesn't need to check for the latter at all.

kchibisov: it's better to check just _in case_, like it could be redundant, but some drivers may have bugs.

quaternic: So, that would be a driver that claims to support `EGL_EXT_device_enumeration` but actually doesn't. I suppose it makes sense to check for that and report it as a bug in the driver.

kchibisov: It could also be a _middle layer_ bug, like if something breaks in `libglvnd` because it also has to wire everything...

quaternic: Oh, yeah. I'll make that case return an error with a different message. Would panicking in debug mode make sense?

kchibisov: Library should really not panic and just propagate.

glutin/src/api/egl/device.rs
```diff
@@ -33,19 +33,24 @@ impl Device {
...
-            || !no_display_extensions.contains("EGL_EXT_device_base")
-        {
-            return Err(ErrorKind::NotSupported("EGL does not support EGL_EXT_device_base").into());
+        // The former depends on EGL_EXT_device_query, so also check that just in case.
+        if !client_extensions.contains("EGL_EXT_device_base") {
+            let query = client_extensions.contains("EGL_EXT_device_query");
+            let enumr = client_extensions.contains("EGL_EXT_device_enumeration");
+            if !(query & enumr) {
+                let msg = if !enumr {
+                    "enumerating devices is not supported by the EGL instance"
```
MarijnS95, line 47:
```suggestion
                    "Enumerating devices is not supported by the EGL instance"
```

glutin/src/api/egl/device.rs
```diff
@@ -33,19 +33,24 @@ impl Device {
...
+        // Querying devices requires EGL_EXT_device_enumeration or EGL_EXT_device_base.
         //
-        // Or we can check for the EGL_EXT_device_base extension since it contains both
-        // extensions.
-        if (!no_display_extensions.contains("EGL_EXT_device_enumeration")
-            && !no_display_extensions.contains("EGL_EXT_device_query"))
-            || !no_display_extensions.contains("EGL_EXT_device_base")
-        {
-            return Err(ErrorKind::NotSupported("EGL does not support EGL_EXT_device_base").into());
+        // The former depends on EGL_EXT_device_query, so also check that just in case.
```
MarijnS95, line 41: For readability (especially inside this diff), perhaps this last line of comment belongs _inside_ the `if`?

MarijnS95 (commented):
Nice catch, the original logic doesn't match what's described in the comment :)

glutin/src/api/egl/device.rs
```diff
@@ -33,19 +33,24 @@ impl Device {
...
+        if !client_extensions.contains("EGL_EXT_device_base") {
+            let query = client_extensions.contains("EGL_EXT_device_query");
+            let enumr = client_extensions.contains("EGL_EXT_device_enumeration");
+            if !(query & enumr) {
```
kchibisov, line 45:
```suggestion
            if !(query && enumr) {
```

glutin/src/api/egl/device.rs
```diff
@@ -33,19 +33,24 @@ impl Device {
...
+                let msg = if !enumr {
+                    "Enumerating devices is not supported by the EGL instance"
+                } else {
+                    // this should not happen as device_enumeration depends on device_query
```
kchibisov, line 49:
```suggestion
                    // This should not happen as device_enumeration depends on device_query.
```

kchibisov (approved):

glutin/src/api/egl/device.rs
quaternic: Well, trying to reposition that comment led to also reorganizing the code, but I think it's better overall now.

MarijnS95 (approved):

PR #1678, opened 2024-05-14
Upgrade `winit` 0.29.2 → 0.30.0

ErichDonGubler:
- [x] Tested on all platforms changed
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [x] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [x] Created or updated an example program if it would help users understand this functionality: I've updated examples, which I think might be sufficient?

Resolves https://github.com/rust-windowing/glutin/issues/1674.

PR #1687, opened 2024-07-02
examples: Reuse `not_current_context` in `resumed()` handler

MarijnS95:
When testing the `glutin` example on Android, after suspending and resuming the app no triangle appears.  As it turns out the `winit 0.30` bump in #1678 never reuses the `not_current_context` that was set apart in `suspended()`, but creates a new one, without re-running initialization code such as creating the `Renderer`.

Partially restore original code before the `winit 0.30` bump, which never created a context in `resumed()` because it was already available before the loop started (and moved back in place in `suspended()`). Note also that more initialization code that now (unnecessarily) lives in `resumed()` used to be outside of the loop.

### DIFF
--- a/glutin_examples/src/lib.rs
+++ b/glutin_examples/src/lib.rs
@@ -92,7 +92,10 @@ impl ApplicationHandler for App {
             .with_context_api(ContextApi::OpenGl(Some(Version::new(2, 1))))
             .build(raw_window_handle);
 
-        self.not_current_gl_context.replace(unsafe {
+        // Reuse the uncurrented context from a suspended() call if it exists, otherwise
+        // this is the first time resumed() is called, where the context still
+        // has to be created.
+        let not_current_gl_context = self.not_current_gl_context.take().unwrap_or_else(|| unsafe {
             gl_display.create_context(&gl_config, &context_attributes).unwrap_or_else(|_| {
                 gl_display.create_context(&gl_config, &fallback_context_attributes).unwrap_or_else(
                     |_| {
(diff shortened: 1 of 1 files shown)


MarijnS95:
CC @marc2332 @ErichDonGubler, note also that things from outside the loop were moved into `fn resume()` even though they probably shouldn't have been there to begin with?

ErichDonGubler:
Ah, this was my bad, I think. I must have mistakenly interpreted the old code to re-init every time, but if the purpose of the not-current context is to be reused, then…😅 that's obviously incorrect.

glutin_examples/src/lib.rs
```diff
@@ -92,7 +92,10 @@ impl ApplicationHandler for App {
             .with_context_api(ContextApi::OpenGl(Some(Version::new(2, 1))))
             .build(raw_window_handle);
 
-        self.not_current_gl_context.replace(unsafe {
+        // Reuse the context from a suspend() call if it exists, otherwise this is the
+        // first time resume() is called, where the context still has to be
+        // created.
```
MarijnS95, line 97:
```suggestion
        // Reuse the uncurrented context from a suspended() call if it exists, otherwise this is the
        // first time resumed() is called, where the context still has to be
        // created.
```

MarijnS95:
@ErichDonGubler yes, it looks like everyone missed this in review. Probably clobbered by the fact that, on Windows/WGL, a window needs to be created **before** "making a connection to the compositor" (which is creating the display), which requires `ActiveEventLoop` since `winit 0.30`. In turn a lot of that setup code moved from outside the loop, into the loop (specifically into the `resumed()` call) to be able to do this.

---

There's a whole side-issue on `winit` where we're discussing and figuring out that `resumed()` has a lot more (platform-specific) semantics to it and isn't the right place to create windows either (unless the actual current usage is split out into a differently-named callback).

kchibisov (approved):

PR #1688, opened 2024-08-08
Fixed crash in `EglGetProcAddress` on Win32-x86 platform

varphone:

# Summary

Change the calling convention of the `EglGetProcAddress` declaration to `extern "system"` to automatically use the correct calling convention for the platform.

# Details

The `EglGetProcAddress` function uses the `stdcall` calling convention on the `Win32-x86` platform, but it was previously declared as `extern "C"` in the code.

On the `Win32-x86` platform, `extern "C"` functions use the `cdecl` calling convention by default.

This mismatch in calling conventions caused a crash.

By changing the declaration to `extern "system"`, the correct calling convention for the platform will be used automatically, preventing the crash.

# Testing

- Verified that the application no longer crashes on the `Win32-x86` platform.
- Ensured that the change does not affect other platforms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Fixed EGL's `Device::query_devices()` being too strict about required extensions
+- Fixed crash in `EglGetProcAddress` on Win32-x86 platform
 
 # Version 0.32.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Unreleased
 
 - Fixed EGL's `Device::query_devices()` being too strict about required extensions
-- Fixed crash in `EglGetProcAddress` on Win32-x86 platform
+- Fixed crash in `EglGetProcAddress` on Win32-x86 platform due to wrong calling convention
 
 # Version 0.32.0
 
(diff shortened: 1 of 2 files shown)


kchibisov (changes requested):
Could you add a changelog entry for that? should be fine after that.

varphone:
> Could you add a changelog entry for that? should be fine after that.

Already added, please review again.

MarijnS95:
In general I'm not a fan of "Fixed crash" or other context-less titles using the word "fix", can we come up with anything more descriptive (i.e. _"use the correct calling convention for `EglGetProcAddress()` on Win32-x86 (optionally: preventing a crash)"_)? Same for the PR title.

Probably not a requirement to merge though, depending on what Kirill thinks of this.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Fixed EGL's `Device::query_devices()` being too strict about required extensions
+- Fixed crash in `EglGetProcAddress` on Win32-x86 platform
```
kchibisov, line 4:
```suggestion
- Fixed crash in `EglGetProcAddress` on Win32-x86 platform due to wrong calling convention
```

kchibisov (changes requested):
Should be fine after that.

kchibisov (approved):

Issue #1689, opened 2024-08-25
Glutin fails to create a surface when Mesa's EGL 1.5 implementation for GBM is present alongside native (X11) EGL 1.4


janKilika:
**Issue**: In my understanding, per presence of Mesa's EGL 1.5 implementation for GBM, symbols for `Khr` platform functions are actually properly loaded. This in and of itself would not be an issue, but `eglGetPlatformDisplay` actually fails to error and, from what I've managed to gather through testing, yields a display that is valid as `EglDisplay::Ext` (note: the `EGL_EXT_platform_base` extension is present; I'm not sure whether such an interaction would occur if it isn't, nor is it clear what would happen EGL 1.5 and implementations older than EGL 1.4 collide). That, in turn, creates problems down the line, as EGL stably outputs a `BadDisplay` error when a `Khr` surface is attempted to be created.

**Potential Solution**: Though I don't think this circumstance is something that a library can sanely check for before obtaining the display connection, initializing the EGL display does yield the actual version of the active implementation. As such, having a check that downgrades `EglDisplay` from `Khr` to `Ext` provided the yielded version is 1.4 (note that (a) `EGL_EXT_platform_base` is only available for 1.4 according to the specification and (b) a `Khr` display wouldn't even make sense for any version that is not 1.5) and `EGL_EXT_platform_base` is present among client extensions could likely negate the issue.

janKilika:
Version originally tested is the latest on Cargo, v0.32.0, however the problem does also occur in the master branch, as the git clone exhibits the same issue, outputting a `Khr` display in spite of it belonging to EGL 1.4, which can only happen when `egl.GetPlatformDisplay` (a) is properly loaded (and it is, due to EGL 1.5 being present for GBM) and (b) doesn't actually fail (and failure here would be expected due to disparity between EGL versions for native X11 and GBM, yet somehow it does not occur, which I would suppose points to some bizarre behavior on part of Mesa's EGL implementation).

In my case, downgrading the illegal `Khr` display to `Ext` fixed the issue, as the subsequent surface creation proceeds as normal, which means that the `egl.GetPlatformDisplay` call returns a valid `Ext` display when it encounters a native display that only supports EGL 1.4.

kchibisov:
[...]

PR #1690, opened 2024-08-26
Implement a workaround for `libglvnd` aliasing `eglGetPlatformDisplay` and `eglGetPlatformDisplayEXT` and causing errors at surface creation due to use of KHR functions when the EGL version for the dispaly is 1.4.

janKilika:
**Issue addressed**: EGL's `Display::new` making a `Khr` display (those are only valid for 1.5) when EGL version for the platform is actually 1.4, which causes errors when 1.5 surface creation is attempted afterwards. Caused by `libglvnd` "aliasing" `eglGetPlatformDisplay` and `eglGetPlatformDisplayEXT` (as in both functions call the same vendor function internally), which makes the former succeed and create a display that is only valid as an `Ext` display despite what otherwise would be expected according to specification.

**Solution**: Check for the version after performing an `eglInitialize` call, as it returns the actual version for the platform's implementation of EGL and therefore enables us to detect whether the aliasing issue occurred and downgrade the display from `Khr` to `Ext` if it did.

**Changes**: The only actual change is that it is checked in `Display::initialize_display` with the `Display::sanitize_libglvnd_aliasing` function whether (a) `EglDisplay` is of variant `Khr` and (b) version is equal to 1.4, and if both conditions are true, `EglDisplay` is downgraded from `Khr` to `Ext`. This seems to be safe, as `eglGetPlatformDisplay` and `eglGetPlatformDisplayEXT` in practice are made to be aliases of each other by `libglvnd`, and if any of the attributes passed are invalid for the underlying function, it is presumed that it would return with an error.

See issue #1689 for further details.
Also see [issue #251](https://gitlab.freedesktop.org/glvnd/libglvnd/-/issues/251) on `libglvnd`'s repository.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 - Fixed EGL's `Device::query_devices()` being too strict about required extensions
 - Fixed crash in `EglGetProcAddress` on Win32-x86 platform due to wrong calling convention
-- Fixed EGL's `Display::new` making an `EGLDisplay::Khr` when the underlying platform only supports `EGLDisplay::Ext` due to `libglvnd` using the same vendor functions for obtaining a display connection both for `eglGetPlatformDisplay` and `eglGetPlatformDisplayEXT`.
+- Implement a workaround for EGL's `Display::new` making an `EGLDisplay::Khr` when the underlying platform only supports `EGLDisplay::Ext` due to `libglvnd` using the same vendor functions for obtaining a display connection both for `eglGetPlatformDisplay` and `eglGetPlatformDisplayEXT`.
 
 # Version 0.32.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 - Fixed EGL's `Device::query_devices()` being too strict about required extensions
 - Fixed crash in `EglGetProcAddress` on Win32-x86 platform due to wrong calling convention
-- Implement a workaround for EGL's `Display::new()` making an `EGLDisplay::Khr` when the underlying platform only supports `EGLDisplay::Ext` due to `libglvnd` using the same vendor functions for obtaining a display connection for both `eglGetPlatformDisplay()` and `eglGetPlatformDisplayEXT()`.
+- Fixed EGL's `Display::new()` making an `EGLDisplay::Khr` when the EGL version for the display is 1.4 or lower.
 
 # Version 0.32.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Fixed EGL's `Device::query_devices()` being too strict about required extensions
 - Fixed crash in `EglGetProcAddress` on Win32-x86 platform due to wrong calling convention
+- Fixed EGL's `Display::new` making an `EGLDisplay::Khr` when the underlying platform only supports `EGLDisplay::Ext` due to `libglvnd` using the same vendor functions for obtaining a display connection both for `eglGetPlatformDisplay` and `eglGetPlatformDisplayEXT`.
 
 # Version 0.32.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 - Fixed EGL's `Device::query_devices()` being too strict about required extensions
 - Fixed crash in `EglGetProcAddress` on Win32-x86 platform due to wrong calling convention
-- Implement a workaround for EGL's `Display::new` making an `EGLDisplay::Khr` when the underlying platform only supports `EGLDisplay::Ext` due to `libglvnd` using the same vendor functions for obtaining a display connection both for `eglGetPlatformDisplay` and `eglGetPlatformDisplayEXT`.
+- Implement a workaround for EGL's `Display::new()` making an `EGLDisplay::Khr` when the underlying platform only supports `EGLDisplay::Ext` due to `libglvnd` using the same vendor functions for obtaining a display connection for both `eglGetPlatformDisplay()` and `eglGetPlatformDisplayEXT()`.
 
 # Version 0.32.0
 
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -512,6 +512,9 @@ impl Display {
             // for a platform that does not support KHR functions and only supports EXT
             // functions. As such, we check if the version is equal to 1.4 and downgrade to
             // `EglDisplay::Ext` if so.
+            // See: https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/src/EGL/libegl.c#L270-358
+            // https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/src/EGL/libegl.c#L409-461
+            // https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/include/glvnd/libeglabi.h#L231-232
             (EglDisplay::Khr(display), Version { major: 1, minor: 4 }) => EglDisplay::Ext(display),
             // We do not do anything otherwise.
             (display, _) => display,
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -507,11 +507,11 @@ impl Display {
     #[cfg(free_unix)]
     fn sanitize_libglvnd_aliasing(display: EglDisplay, version: Version) -> EglDisplay {
         match (display, version) {
-            // libglvnd has an unsavoury quirk of unconditionally aliasing `getPlatformDisplay` (unavailable on EGL 1.4) and
-            // `getPlatformDisplayEXT`, which may result in an `EglDisplay::Khr` being created
-            // for a platform that does not support KHR functions and only supports EXT
-            // functions. As such, we check if the version is equal to 1.4 and downgrade to
-            // `EglDisplay::Ext` if so.
+            // libglvnd has an unsavoury quirk of unconditionally aliasing `getPlatformDisplay`
+            // (unavailable on EGL 1.4) and `getPlatformDisplayEXT`, which may result in
+            // an `EglDisplay::Khr` being created for a platform that does not support
+            // KHR functions and only supports EXT functions. As such, we check if the
+            // version is equal to 1.4 and downgrade to `EglDisplay::Ext` if so.
             // See: https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/src/EGL/libegl.c#L270-358
             // https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/src/EGL/libegl.c#L409-461
             // https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/include/glvnd/libeglabi.h#L231-232
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -487,18 +487,18 @@ impl Display {
 
         // Sanitize the display provider, as the received EGL 1.4 display could have
         // been marked incorrectly as `EglDisplay::Khr`.
-        #[cfg(free_unix)]
+        // See pull request #1690.
         let display = match display {
-            // Since according to libglvnd, `eglGetPlatformDisplay` and `GetPlatformDisplayEXT`
-            // aren't really differentiated under the hood, we must check if the version of the
-            // initialized display is not sensible for the EglDisplay type and downgrade it if so
-            // See: https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/src/EGL/libegl.c#L270-358
-            // https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/src/EGL/libegl.c#L409-461
-            // https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/include/glvnd/libeglabi.h#L231-232
-            // https://gitlab.freedesktop.org/glvnd/libglvnd/-/issues/251
-            EglDisplay::Khr(display) if version == Version { major: 1, minor: 4 } => {
+            // `eglGetPlatformDisplay` and `GetPlatformDisplayEXT` aren't really differentiated,
+            // we must check if the version of the initialized display is not sensible for the
+            // EglDisplay type and downgrade it if so.
+            EglDisplay::Khr(display) if version <= Version { major: 1, minor: 4 } => {
                 let client_extensions = CLIENT_EXTENSIONS.get().unwrap();
-                if client_extensions.contains("EGL_EXT_platform_base") {
+                if client_extensions.contains("EGL_EXT_platform_base")
+                    && (version == Version { major: 1, minor: 4 })
+                {
+                    // `EGL_EXT_platform_base` requires EGL 1.4 per specification; we cannot safely
+                    // presume that an `Ext` display would be valid for older versions.
                     EglDisplay::Ext(display)
                 } else {
                     EglDisplay::Legacy(display)
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -485,9 +485,28 @@ impl Display {
             Version::new(major as u8, minor as u8)
         };
 
-        // Sanitize for libglvnd's aliasing.
+        // Sanitize the display provider, as the received EGL 1.4 display could have
+        // been marked incorrectly as `EglDisplay::Khr`.
         #[cfg(free_unix)]
-        let display = Self::sanitize_egl_display_provider(display, version);
+        let display = match display {
+            // Since according to libglvnd, `eglGetPlatformDisplay` and `GetPlatformDisplayEXT`
+            // aren't really differentiated under the hood, we must check if the version of the
+            // initialized display is not sensible for the EglDisplay type and downgrade it if so
+            // See: https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/src/EGL/libegl.c#L270-358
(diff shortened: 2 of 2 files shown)


glutin/src/api/egl/display.rs
```diff
@@ -499,6 +502,19 @@ impl Display {
         });
         Ok(Self { inner })
     }
+
+    fn sanitize_libglvnd_aliasing(display: EglDisplay, version: Version) -> EglDisplay {
+        match (display, version) {
+            // libglvnd has an unsavoury quirk of aliasing `getPlatformDisplay` and
```
MarijnS95, line 508: Can you link the source code that describes this "unsavoury quirk", and how it's gone in 1.5?

MarijnS95 (commented):
We can probably fit a more descriptive title than "aliasing issue" without going over any character limit.

glutin/src/api/egl/display.rs
janKilika: [Relevant portion here](https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/master/src/EGL/libegl.c#L270-358). `libglvnd` uses the same underlying function of `GetPlatformDisplayCommon` for both of the aliased functions. Although it passes a parameter [`funcName`](https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/master/src/EGL/libegl.c#L272) to it with the name of the function that called `GetPlatformDisplayCommon`, it only uses it for error reporting.

janKilika: Also of note is [this comment](https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/master/include/glvnd/libeglabi.h#L231-232) for `getPlatformDisplay` field of `__EGLapiImports`.

janKilika: > and how it's gone in 1.5?

As for the problem not being relevant for EGL 1.5, it's because `Khr` displays are actually valid for 1.5 platforms, which means KHR platform functions (i.e. `eglCreatePlatform`..`Surface` functions) won't subsequently crash with `BadDisplay`. For 1.4 platforms, these functions cannot be used, as they error out with `BadDisplay` when provided with a display that was returned by the aliased function; EXT platform functions must be used instead (i.e. `eglCreatePlatform`..`SurfaceEXT`).

glutin/src/api/egl/display.rs
```diff
@@ -499,6 +502,19 @@ impl Display {
...
+            // `getPlatformDisplayEXT`, which may result in an `EglDisplay::Khr` being created
+            // for a platform that does not support KHR functions and only supports EXT
+            // functions. As such, we check if the version is equal to 1.4 and downgrade to
+            // `EglDisplay::Ext` if so.
+            (EglDisplay::Khr(display), Version { major: 1, minor: 4 }) => EglDisplay::Ext(display),
```
kchibisov, line 513: I was thinking to ensure that the function pointers for `egl.GetPlatformDisplay` and `egl.GetPlatformDisplayExt` are the same before doing such thing, so you can ensure that the cast is _safe_ because the same functions were used to produce the pointer.

Just doing a raw cast might be not a great idea if the functions are _different_?

kchibisov (changes requested):
Also need a changelog entry for this.

glutin/src/api/egl/display.rs
janKilika: > I was thinking to ensure that the function pointers for `egl.GetPlatformDisplay` and `egl.GetPlatformDisplayExt` are the same before doing such thing, so you can ensure that the cast is _safe_ because the same functions were used to produce the pointer.

The problem with that is that they are only aliased by `libglvnd` internally. The functions requested through the dynamic library are technically distinct, but otherwise call the same underlying vendor function.

> Just doing a raw cast might be not a great idea if the functions are different?

I'd say that considering the version is explicitly checked here to be 1.4 before the downgrade, this isn't much of a problem per `libglvnd`'s aliasing and the fact that `EglDisplay::Khr` shouldn't even be possible for 1.4. Although, I suppose this could cause EGL to error out at surface creation if `EGL_EXT_platform_base` isn't actually available for the platform, but that hinges on the KHR display connection request not failing beforehand, and I'm not entirely sure how that would play out when the underlying vendor function is `eglGetDisplay` or whether that's even possible.

kchibisov: >The problem with that is that they are only aliased by libglvnd internally. The functions requested through the dynamic library are technically distinct, but otherwise call the same underlying vendor function.

We can not assume that, if pointers are different then we not do a casting since they could in fact be different. The problem is that libglvnd tells us that 1.5 exists due to one vendor providing it, while it's not due to issue in how it's done in libglvnd. When using only a vendor library you won't ever have this issue in the first place from what I understand.

So, ensure the pointers are the same and that's the only time we can do such a cast safely.

janKilika: > We can not assume that, if pointers are different then we not do a casting since they could in fact be different.

`libglvnd`'s public functions, the functions that are actually used through the dynamic library, are distinct (as the aliasing is only done internally in `libglvnd`; the public functions call the same underlying vendor function), so the check would inevitably fail and the downgrade wouldn't happen.

> When using only a vendor library you won't ever have this issue in the first place from what I understand.

Likely not, as vendor libraries would, I presume, only have the functions that are relevant to them.

> The problem is that libglvnd tells us that 1.5 exists due to one vendor providing it

Well, not quite. Firstly, `libglvnd` advertises client extensions if at least one implementation supports them, so fully discriminating based on them is out of the question. Secondly, `eglGetPlatformDisplay` and `eglGetPlatformDisplayEXT` are functions that exist in `libglvnd` and are loaded regardless of whether any of the platforms support them, and both internally call the same underlying vendor function, so fully discriminating based on whether they are loaded is also out of the question for `libglvnd`. If the display creation manages to succeed, we are basically only left to do a check for the version after display initialization, which the PR already introduces.

janKilika:
> Also need a changelog entry for this.

Done!

glutin/src/api/egl/display.rs
kchibisov: Hm, indeed they'll be different. I was thinking they just provide the same pointer when loading but no, they use different pointers but call to the same implementation.

In this case, I'd suggest to this only for `#[cfg(free_unix)]`.

I'll also open an issue upstream.

kchibisov (changes requested):

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - Fixed EGL's `Device::query_devices()` being too strict about required extensions
 - Fixed crash in `EglGetProcAddress` on Win32-x86 platform due to wrong calling convention
+- Fixed EGL's `Display::new` making an `EGLDisplay::Khr` when the underlying platform only supports `EGLDisplay::Ext` due to `libglvnd` using the same vendor functions for obtaining a display connection both for `eglGetPlatformDisplay` and `eglGetPlatformDisplayEXT`.
```
kchibisov, line 5: That's not really a fix, it's a _workaround_, since it's a libglvnd bug, obviously.

MarijnS95: And the same applies to the PR & commit title which also need to be improved, as requested before.

glutin/src/api/egl/display.rs
MarijnS95: Let's make these permalinks so that they keep referencing the same lines of code: https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/src/EGL/libegl.c#L270-358.

And the request is to provide this information within the code comment, so that folks don't have to go find this discussion thread on a PR via git blame.

MarijnS95:
https://registry.khronos.org/EGL/sdk/docs/man/html/eglQueryString.xhtml

It's unfortunate that we cannot use `eglQueryString` with `NO_DISPLAY` to "supposedly" (?) read the EGL version before calling `eglInitialize()`, until version 1.5? It's unclear to me if that version requirement applies to `EGL_EXTENSIONS` because that's the only query listed to support `NO_DISPLAY` - but it would mean our call is invalid on your given `libglvnd` too?

MarijnS95:
Reading somewhat further, we don't ever check if https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_client_extensions.txt works and following that, https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_platform_base.txt is available before calling `eglGetPlatformDisplayEXT()`?

janKilika:
> https://registry.khronos.org/EGL/sdk/docs/man/html/eglQueryString.xhtml
> 
> It's unfortunate that we cannot use `eglQueryString` with `NO_DISPLAY` to "supposedly" (?) read the EGL version before calling `eglInitialize()`, until version 1.5? It's unclear to me if that version requirement applies to `EGL_EXTENSIONS` because that's the only query listed to support `NO_DISPLAY` - but it would mean our call is invalid on your given `libglvnd` too?

If by call you mean querying the client extensions, then, as far I can see, `libglvnd` provides querying for client extensions through `eglQueryString` (the function exists in `libglvnd` and [specifically checks](https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/master/src/EGL/libegl.c#L925) for `NO_DISPLAY`) in accord with both the 1.5 specification and `EGL_EXT_client_extensions`, so the call is technically valid.

MarijnS95:
Yes I mean us calling `eglQueryString` for `NO_DISPLAY`. According to https://registry.khronos.org/EGL/sdk/docs/man/html/eglQueryString.xhtml that's only supported from 1.5 onwards but `EGL_EXT_client_extensions` (which requires 1.4) specifies that it's possible as long you handle the `NULL`+`BAD_DISPLAY` case gracefully.

I was wondering if we could use something like that to detect and prevent the failure case from happening (by never calling `eglGetPlatformDisplay()`) but it seems like we have a chicken-egg problem here anyway by not knowing if we have a 1.5 display before calling functions that "rely on" knowing the version and/or extensions 😅 

---

Also, reading back the PR description, can you reword it to provide a summary of both the problem and proposed solution, rather than linking through to an 18-comment-dense issue thread "for further details"?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - Fixed EGL's `Device::query_devices()` being too strict about required extensions
 - Fixed crash in `EglGetProcAddress` on Win32-x86 platform due to wrong calling convention
+- Implement a workaround for EGL's `Display::new` making an `EGLDisplay::Khr` when the underlying platform only supports `EGLDisplay::Ext` due to `libglvnd` using the same vendor functions for obtaining a display connection both for `eglGetPlatformDisplay` and `eglGetPlatformDisplayEXT`.
```
MarijnS95, line 5:
```suggestion
- Implement a workaround for EGL's `Display::new()` making an `EGLDisplay::Khr` when the underlying platform only supports `EGLDisplay::Ext` due to `libglvnd` using the same vendor functions for obtaining a display connection for both `eglGetPlatformDisplay()` and `eglGetPlatformDisplayEXT()`.
```

kchibisov:
Could also link to upstream issue I've just opened.

janKilika:
> Also, reading back the PR description, can you reword it to provide a summary of both the problem and proposed solution, rather than linking through to an 18-comment-dense issue thread "for further details"?

Sure, relevant information provided.

glutin/src/api/egl/display.rs
MarijnS95: Separate question: is EGL `1.3` or older something that even exists out in the wild and that `glutin` could possibly run on (via `libglvnd`)? If so, should we do this?


```suggestion
            (EglDisplay::Khr(display), Version { major, minor }) if major <= 1 && minor <= 4 => EglDisplay::Ext(display),
```

janKilika: > If so, should we do this?

Probably not, since [`EGL_EXT_platform_base`](https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_platform_base.txt), the extension that provides the functions for `Ext` displays, is said in the specification to require 1.4. Downgrading to `Legacy` if the version number is even lower than 1.4 could be done, but I'm not certain how 1.3 or older would behave in this situation.

glutin/src/api/egl/display.rs
```diff
@@ -499,6 +503,24 @@ impl Display {
...
+            // libglvnd has an unsavoury quirk of unconditionally aliasing `getPlatformDisplay`
+            // (unavailable on EGL 1.4) and `getPlatformDisplayEXT`, which may result in
+            // an `EglDisplay::Khr` being created for a platform that does not support
+            // KHR functions and only supports EXT functions. As such, we check if the
+            // version is equal to 1.4 and downgrade to `EglDisplay::Ext` if so.
+            // See: https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/src/EGL/libegl.c#L270-358
+            // https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/src/EGL/libegl.c#L409-461
+            // https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/include/glvnd/libeglabi.h#L231-232
+            // https://gitlab.freedesktop.org/glvnd/libglvnd/-/issues/251
+            (EglDisplay::Khr(display), Version { major: 1, minor: 4 }) => EglDisplay::Ext(display),
```
kchibisov, line 519: I'd call this `sanitize_egl_display_provider` and always converted `Khr` to `Ext` if the version is `1.4` and the `ext` extension is present. Otherwise downgrade to legacy.

For more context https://gitlab.freedesktop.org/glvnd/libglvnd/-/issues/251 and also add an explanation telling that they are all basically the same, since it's generally, a `EGLDisplay`.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - Fixed EGL's `Device::query_devices()` being too strict about required extensions
 - Fixed crash in `EglGetProcAddress` on Win32-x86 platform due to wrong calling convention
+- Implement a workaround for EGL's `Display::new()` making an `EGLDisplay::Khr` when the underlying platform only supports `EGLDisplay::Ext` due to `libglvnd` using the same vendor functions for obtaining a display connection for both `eglGetPlatformDisplay()` and `eglGetPlatformDisplayEXT()`.
```
kchibisov, line 5:
```suggestion
- Fixed EGL's `Display::new()` making an `EGLDisplay::Khr` when the EGL version is below 1.4.
```

glutin/src/api/egl/display.rs
```diff
@@ -485,6 +485,10 @@ impl Display {
             Version::new(major as u8, minor as u8)
         };
 
+        // Sanitize for libglvnd's aliasing.
+        #[cfg(free_unix)]
```
kchibisov, line 489:
```suggestion
        // Sanitize the display version.
```

glutin/src/api/egl/display.rs
```diff
@@ -499,6 +503,29 @@ impl Display {
         });
         Ok(Self { inner })
     }
+
+    #[cfg(free_unix)]
```
kchibisov, line 507:
```suggestion
```

glutin/src/api/egl/display.rs
```diff
@@ -499,6 +503,29 @@ impl Display {
...
+        let client_extensions = CLIENT_EXTENSIONS.get().unwrap();
+        match display {
+            // Since according to libglvnd, `eglGetPlatformDisplay` and `GetPlatformDisplayEXT`
+            // aren't really differentiated under the hood, we must check if the version of the
+            // initialized display is not sensible for the EglDisplay type and downgrade it if so
+            // See: https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/src/EGL/libegl.c#L270-358
+            // https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/src/EGL/libegl.c#L409-461
+            // https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/include/glvnd/libeglabi.h#L231-232
+            // https://gitlab.freedesktop.org/glvnd/libglvnd/-/issues/251
+            EglDisplay::Khr(display) if version == Version { major: 1, minor: 4 } => {
```
kchibisov, line 518: Shouldn't this be

```suggestion
            EglDisplay::Khr(display) if version <= Version { major: 1, minor: 4 } => {
```

kchibisov (changes requested):
I hope that this is the last one.

CHANGELOG.md
MarijnS95: @kchibisov can we correct that to `when the EGL version is 1.4 or lower`. Otherwise, when the `EGL version is below 1.4`?

glutin/src/api/egl/display.rs
```diff
@@ -499,6 +503,29 @@ impl Display {
...
+    fn sanitize_egl_display_provider(display: EglDisplay, version: Version) -> EglDisplay {
+        let client_extensions = CLIENT_EXTENSIONS.get().unwrap();
+        match display {
+            // Since according to libglvnd, `eglGetPlatformDisplay` and `GetPlatformDisplayEXT`
+            // aren't really differentiated under the hood, we must check if the version of the
+            // initialized display is not sensible for the EglDisplay type and downgrade it if so
+            // See: https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/src/EGL/libegl.c#L270-358
+            // https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/src/EGL/libegl.c#L409-461
+            // https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/606f6627cf481ee6dcb32387edc010c502cdf38b/include/glvnd/libeglabi.h#L231-232
+            // https://gitlab.freedesktop.org/glvnd/libglvnd/-/issues/251
```
MarijnS95, line 517: Let's drop all this too and simplify/abbreviate it?

CHANGELOG.md
kchibisov: sure.

glutin/src/api/egl/display.rs
```diff
@@ -485,6 +485,10 @@ impl Display {
             Version::new(major as u8, minor as u8)
         };
 
+        // Sanitize for libglvnd's aliasing.
+        #[cfg(free_unix)]
+        let display = Self::sanitize_egl_display_provider(display, version);
```
MarijnS95, line 490: What if we do this directly in the tail of `get_platform_display()`? That should make the code a lot more compact since you can simply change the `.map(EglDisplay::Khr)` to use `Egl` when the version is lower.

We have an exact same `if-else` in `get_platform_display_ext()` already.

kchibisov: yeah, not a bad idea.

kchibisov (approved):

PR #1691, opened 2024-08-26
EGL: Implement interfaces for `EGL_EXT_device_drm`

MarijnS95:
This extension (together with the optional `EGL_EXT_device_drm_render_node` extension) allows querying the DRM primary and render device node filenames from `EGLDevice`, to help associating them with DRM devices.

Additionally a platform display _can_ be created with a DRM file descriptor with master permissions, but this is only useful in situations where EGL might fail to open a file descriptor itself or will fail to acquire master permissions (which are only required when the implementation needs to "modify output attributes": but no behaviour is defined by this extension that requires this).

This file descriptor will be duplicated by the implementation if it needs to use it beyond the call to `eglGetPlatformDisplayEXT()`, and does not need to be kept alive by the caller.

Note that `EGL_EXT_output_drm` is omitted for now, because it relies on the equally unimplemented `EGL_EXT_output_base` extension (neither of which are available on my hardware for testing).


kchibisov (changes requested):

glutin/src/api/egl/device.rs
```diff
@@ -114,18 +119,57 @@ impl Device {
...
+    /// Requires the [`EGL_EXT_device_drm`] extension.
+    ///
+    /// If the [`EGL_EXT_device_drm_render_node`] extension is supported, this
+    /// is guaranteed to return the primary device node file, or [`None`].
+    /// Consult [`Self::drm_render_device_node_file()`] to retrieve the render
+    /// device node file.
+    ///
+    /// [`EGL_EXT_device_drm`]: https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_device_drm.txt
+    /// [`EGL_EXT_device_drm_render_node`]: https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_device_drm_render_node.txt
+    pub fn drm_device_node_file(&self) -> Option<String> {
```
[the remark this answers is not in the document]
MarijnS95: This is what the EGL constants name them, which I stuck with, but we can surely change.

glutin_examples/Cargo.toml
```diff
@@ -19,11 +19,13 @@ x11 = ["glutin-winit/x11"]
 wayland = ["glutin-winit/wayland", "winit/wayland-dlopen", "winit/wayland-csd-adwaita"]
 
 [dependencies]
+anyhow = "1"
```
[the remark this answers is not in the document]
MarijnS95: I prefer `?` but sure.

glutin/src/api/egl/device.rs
```diff
@@ -114,18 +120,61 @@ impl Device {
...
+    pub fn drm_render_device_node_path(&self) -> Option<PathBuf> {
+        if !self.extensions.contains("EGL_EXT_device_drm_render_node") {
+            return None;
+        }
+
+        const EGL_DRM_RENDER_NODE_PATH_EXT: egl::types::EGLenum = 0x3377;
+        // SAFETY: We pass a valid EGLDevice pointer, and validated that the enum name
+        // is valid because the extension is present.
+        unsafe { Self::query_string(self.raw_device(), EGL_DRM_RENDER_NODE_PATH_EXT) }
+            .map(PathBuf::from)
```
MarijnS95, line 170: Should we let `query_string` return a `&'static CStr` or `&'static OsStr` since `eglQueryDeviceStringEXT()` is guaranteed to return a static string (which would disappear if the implementation library is unloaded...)?

MarijnS95:
> Note that `EGL_EXT_output_drm` is omitted for now, because it relies on the equally unimplemented `EGL_EXT_output_base` extension (neither of which are available on my hardware for testing).

The `output` extensions only appear to be implemented on Nvidia drivers (and my laptop unfortunately has _no outputs_ connected on the dedicated GPU), and don't seem to be used for anything else than creating `EGLStreams` which is a completely separate feature I don't want to build in `glutin`.

glutin/src/api/egl/device.rs
kchibisov: Well, you can not _unload_ the library and keep using anything created by glutin, since we never unload.

MarijnS95: Good point (but the user could hold on to the `'static` str). Perhaps I'll refactor it to return `'_` for both `query_string()` and `&Path` here.

kchibisov (approved):

glutin/src/api/egl/device.rs
```diff
@@ -152,7 +206,10 @@ impl Device {
             // SAFETY: RENDERER_EXT and VENDOR are valid strings for device string queries
             // if EGL_EXT_device_query_name.
             unsafe {
-                (Self::query_string(ptr, egl::RENDERER_EXT), Self::query_string(ptr, egl::VENDOR))
+                (
+                    Self::query_string(ptr, egl::RENDERER_EXT).map(String::from),
+                    Self::query_string(ptr, egl::VENDOR).map(String::from),
```
MarijnS95, line 211: Theoretically this returns a lifetime constrained by `&'this_one Egl`, but that would make us carry the lifetime on all of `Device` which we shouldn't do :)

glutin/src/api/egl/device.rs
MarijnS95: Or right, since we have `static EGL: Lazy` this _could_ actually be `'static`?

MarijnS95: We already did this for `extensions_from_ptr()` anyway :)

kchibisov (approved):
From what I can see, it doesn't have any breaking changes, since `static` is more _broad_ lifetime here?

kchibisov:
@MarijnS95 it was `&str`, I'm not returning `String` anywhere before?

MarijnS95:
@kchibisov never mind you're right, I misread.

MarijnS95:
> Also, we don't really have to add `&'static str`, since `&str` is more forward compat.

That is true, do you think we'll ever unload or replace the library?

kchibisov:
>That is true, do you think we'll ever unload or replace the library?

I'm not sure about this, probably not, but having `&'static str` is more _restricting_ to what we can do, so, maybe changing back to regular `&str` is better?

Like if someone has a need for `&'static` there they could probably request it, though, we do have `'static` inside the extensions `HashMap` already, so that's why I don't really mind changing to `'static` other stuff that behaves like that.

MarijnS95:
More restrictive to what we can do, but less restrictive to what the caller can do with it once we've given a borrow to them.

PR #1695, opened 2024-09-05
Fix EGL device pointer

chrisduerr:
This fixes an issue where `eglQueryDisplayAttribEXT` would always fail since the pointer passod to it didn't match its expectations.

### DIFF
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -1,7 +1,7 @@
 //! Everything related to `EGLDisplay`.
 
 use std::collections::HashSet;
-use std::ffi::{self, CStr};
+use std::ffi::{self, c_void, CStr};
 use std::mem::MaybeUninit;
 use std::ops::Deref;
 use std::os::raw::c_char;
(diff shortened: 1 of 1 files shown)


glutin/src/api/egl/display.rs
```diff
@@ -211,7 +211,8 @@ impl Display {
             }));
         }
 
-        Device::from_ptr(self.inner.egl, device)
+        let device_ptr = unsafe { ptr::null_mut::<c_void>().offset(device_offset) };
```
chrisduerr, line 214: There's probably a better way to do this? But that's the first thing I could come up with as someone who doesn't do Rust pointers much.

MarijnS95:
Thanks for reporting this @chrisduerr! That's indeed a bug, but this (using `offset()` and all) isn't the right way to deal with passing a "pointer to a to-be-initialized pointer". The meat of your change is in passing `&mut` to `QueryDisplayAttribEXT()`. I've opened #1696 to demonstrate how to do this, while also rewriting the commit description to explain the situation. Do you want a `Reported-by:` tag?

kchibisov:
See https://github.com/rust-windowing/glutin/pull/1696

PR #1696, opened 2024-09-05
egl/display: Pass *pointer to* output variable when querying `DEVICE_EXT`

MarijnS95:
Supersedes #1695

By casting a `null_mut()` ptr and passing it directly _by value_ to `QueryDisplayAttribEXT()`, the function ignores writing the output as it didn't receive an address (it received `NULL`) to write the device to.  Instead we should take the _pointer address of this `null_mut()` pointer_ and provide that to the function instead so that it can _overwrite_ it with a pointer to the requested `eglDevice`.

  - Tested by adding `dbg!(display.device())` to `examples/egl_device.rs`.  This returns `BadParameter` before this patch, and the device afterwards.
- [x] (N/A) Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [x] (N/A) Created or updated an example program if it would help users understand this functionality

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Fixed EGL's `Device::query_devices()` being too strict about required extensions
 - Fixed crash in `EglGetProcAddress` on Win32-x86 platform due to wrong calling convention
+- Fixed EGL's `Display::device()` always returning an error due to invalid pointer-argument passing inside
 
 # Version 0.32.0
 
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -2,11 +2,11 @@
 
 use std::collections::HashSet;
 use std::ffi::{self, CStr};
+use std::fmt;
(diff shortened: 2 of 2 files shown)


MarijnS95:
All other calls to `QueryDisplayAttribEXT()` (despite not for ptrs) use `MaybeUninit` which is an even cleaner way :)

MarijnS95:
@kchibisov I also noticed that `DisplayInner::uses_display_reference()` doesn't check the _returned value_ of the `MaybeUninit` query for `TRACK_REFERENCES`, but only whether `eglQueryDisplayAttrib{KHR,EXT}` did _not_ error out (and returned `TRUE`).

I think we need something like https://github.com/MarijnS95/glutin/compare/egl-display-attrib-value but I don't have any driver to test it on.

kchibisov (approved):

chrisduerr:
Would you mind cutting a new release? Would make my life a little easier.

Issue #1694, opened 2024-09-04
egl: NV_context_priority_realtime support


i509VCB:
Realtime priority to match Vulkan.

https://registry.khronos.org/EGL/extensions/NV/EGL_NV_context_priority_realtime.txt

Issue #1645, opened 2023-11-08
Context priority


i509VCB:
From what I can tell, this is available in EGL and GLX.

PR #1700, opened 2024-09-15
Implement priority hints for EGL

valaphee:
Fixes https://github.com/rust-windowing/glutin/issues/1645
Fixes https://github.com/rust-windowing/glutin/issues/1694

This PR adds a way to specify the priority for contexts. I mostly used the term hint, as its not guaranteed for various reasons that the context actually gets the priority.

Don't know if it makes sense to guarantee the priority. In Chromium it's also only used as a hint.

I also probably should extend the extension check as it is not correctly reported on Android.
https://github.com/chromium/chromium/blob/main/ui/gl/gl_display.cc#L814

Theoretically the extension also exists for GLX but Mesa for example doesn't support it.

### DIFF
--- a/glutin/src/api/egl/context.rs
+++ b/glutin/src/api/egl/context.rs
@@ -9,7 +9,8 @@ use glutin_egl_sys::{egl, EGLContext};
 
 use crate::config::{Api, GetGlConfig};
 use crate::context::{
-    self, AsRawContext, ContextApi, ContextAttributes, GlProfile, RawContext, Robustness, Version,
+    self, AsRawContext, ContextApi, ContextAttributes, GlProfile, Priority, RawContext, Robustness,
+    Version,
 };
 use crate::display::{DisplayFeatures, GetGlDisplay};
 use crate::error::{ErrorKind, Result};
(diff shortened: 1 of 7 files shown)


glutin/src/context.rs
```diff
@@ -228,6 +241,8 @@ pub struct ContextAttributes {
 
     pub(crate) api: Option<ContextApi>,
 
+    pub(crate) priority: Priority,
```
kchibisov, line 244: Given that drivers could confuse things, can we use `Option` internally here? So when it's `None` we don't push, but when it's e.g. `Some(Medium)` it's pushed to ensure the `Medium`.

glutin/src/context.rs
```diff
@@ -201,6 +201,19 @@ impl ContextAttributesBuilder {
...
     }
 
+    /// Sets the priority hint, which might not be honored if the API does not
+    /// support it, if there are constraints on the number of high priority
+    /// contexts available in the system, or system policy limits access to
+    /// high priority contexts to appropriate system privilege level
+    ///
+    /// # Api-specific
+    ///
+    /// - Only EGL at the moment implements context priorities.
```
kchibisov, line 211: I'd probably use the notation with not supported we use everywhere else, so say that `glx`, `wgl`, and `cgl` is not supported.

glutin/src/api/egl/context.rs
```diff
@@ -126,6 +127,24 @@ impl Display {
...
+                Priority::High => egl::CONTEXT_PRIORITY_HIGH_IMG,
+                Priority::Realtime => {
+                    if self.inner.display_extensions.contains("EGL_NV_context_priority_realtime") {
+                        egl::CONTEXT_PRIORITY_REALTIME_NV
+                    } else {
+                        egl::CONTEXT_PRIORITY_HIGH_IMG
+                    }
+                },
+                _ => unreachable!(),
+            } as EGLint);
```
kchibisov, line 145: Put a `match` on a separate line please and then just push with `as` casting.

kchibisov (changes requested):

kchibisov:
You need to add a `CHANGELOG` entry, since you change the public API.

kchibisov:
>I also probably should extend the extension check as it is not correctly reported on Android.
https://github.com/chromium/chromium/blob/main/ui/gl/gl_display.cc#L814

I'm fine with adding such a check just for android target.

kchibisov (approved):

kchibisov:
I've rabased it and added a getter, which could be used to log what level of priority context is using.

glutin/src/api/egl/context.rs
```diff
@@ -126,6 +127,38 @@ impl Display {
...
+
+            // Some android versions don't report support for this extension, even though
+            // it's supported.
+            //
+            // https://github.com/googlevr/gvr-android-sdk/issues/330
+            #[cfg(android_platform)]
+            let android = extensions.contains("EGL_ANDROID_front_buffer_auto_refresh")
+                && extensions.contains("EGL_ANDROID_create_native_client_buffer");
+            #[cfg(not(android_platform))]
+            let android = false;
```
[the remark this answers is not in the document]
kchibisov: I thought about that, and decided to do it that way for now, since we already do things like that...

PR #1701, opened 2024-09-19
wgl: add support for pixel buffer surfaces

SK83RJOSH:
- [x] ~Created or updated an example program if it would help users understand this functionality~

I didn't create an example, since this is standard functionality that was previously unexposed.

### DIFF
--- a/glutin/src/api/wgl/context.rs
+++ b/glutin/src/api/wgl/context.rs
@@ -22,7 +22,7 @@ use crate::prelude::*;
 use crate::private::Sealed;
 use crate::surface::SurfaceTypeTrait;
 
-use super::config::Config;
+use super::{config::Config, surface::WGLSurface};
 use super::display::Display;
 use super::surface::Surface;
 
@@ -387,7 +387,12 @@ impl ContextInner {
 
     fn make_current<T: SurfaceTypeTrait>(&self, surface: &Surface<T>) -> Result<()> {
         unsafe {
-            if wgl::MakeCurrent(surface.hdc as _, self.raw.cast()) == 0 {
+            let hdc = match surface.raw {
+                WGLSurface::Window(_, hdc) => hdc  as _,
+                WGLSurface::PBuffer(_, hdc) => hdc as _,
+            };
+
+            if wgl::MakeCurrent(hdc, self.raw.cast()) == 0 {
                 Err(IoError::last_os_error().into())
             } else {
                 Ok(())
--- a/glutin/src/api/wgl/context.rs
+++ b/glutin/src/api/wgl/context.rs
@@ -24,7 +24,7 @@ use crate::surface::SurfaceTypeTrait;
 
 use super::config::Config;
 use super::display::Display;
-use super::surface::{Surface, WglSurface};
+use super::surface::Surface;
 
 impl Display {
     pub(crate) unsafe fn create_context(
@@ -387,12 +387,7 @@ impl ContextInner {
 
     fn make_current<T: SurfaceTypeTrait>(&self, surface: &Surface<T>) -> Result<()> {
         unsafe {
-            let hdc = match surface.raw {
-                WglSurface::Window(_, hdc) => hdc as _,
-                WglSurface::PBuffer(_, hdc) => hdc as _,
-            };
-
-            if wgl::MakeCurrent(hdc, self.raw.cast()) == 0 {
+            if wgl::MakeCurrent(surface.raw.hdc() as _, self.raw.cast()) == 0 {
                 Err(IoError::last_os_error().into())
             } else {
                 Ok(())
--- a/glutin/src/api/wgl/context.rs
+++ b/glutin/src/api/wgl/context.rs
@@ -22,9 +22,9 @@ use crate::prelude::*;
 use crate::private::Sealed;
 use crate::surface::SurfaceTypeTrait;
 
-use super::{config::Config, surface::WGLSurface};
+use super::config::Config;
 use super::display::Display;
-use super::surface::Surface;
+use super::surface::{Surface, WGLSurface};
 
 impl Display {
     pub(crate) unsafe fn create_context(
@@ -388,7 +388,7 @@ impl ContextInner {
     fn make_current<T: SurfaceTypeTrait>(&self, surface: &Surface<T>) -> Result<()> {
         unsafe {
             let hdc = match surface.raw {
-                WGLSurface::Window(_, hdc) => hdc  as _,
+                WGLSurface::Window(_, hdc) => hdc as _,
                 WGLSurface::PBuffer(_, hdc) => hdc as _,
             };
 
--- a/glutin/src/api/wgl/context.rs
+++ b/glutin/src/api/wgl/context.rs
@@ -24,7 +24,7 @@ use crate::surface::SurfaceTypeTrait;
 
 use super::config::Config;
 use super::display::Display;
-use super::surface::{Surface, WGLSurface};
+use super::surface::{Surface, WglSurface};
 
 impl Display {
     pub(crate) unsafe fn create_context(
@@ -388,8 +388,8 @@ impl ContextInner {
     fn make_current<T: SurfaceTypeTrait>(&self, surface: &Surface<T>) -> Result<()> {
         unsafe {
             let hdc = match surface.raw {
-                WGLSurface::Window(_, hdc) => hdc as _,
-                WGLSurface::PBuffer(_, hdc) => hdc as _,
+                WglSurface::Window(_, hdc) => hdc as _,
+                WglSurface::PBuffer(_, hdc) => hdc as _,
             };
 
             if wgl::MakeCurrent(hdc, self.raw.cast()) == 0 {
--- a/glutin/src/api/wgl/surface.rs
+++ b/glutin/src/api/wgl/surface.rs
@@ -69,12 +69,12 @@ impl Display {
                 attrs.as_ptr(),
             )
         };
-        if hbuf == std::ptr::null() {
+        if hbuf.is_null() {
             return Err(IoError::last_os_error().into());
         }
 
         let hdc = unsafe { extra.GetPbufferDCARB(hbuf) };
-        if hdc == std::ptr::null() {
+        if hdc.is_null() {
             return Err(IoError::last_os_error().into());
         }
 
--- a/glutin/src/api/wgl/surface.rs
+++ b/glutin/src/api/wgl/surface.rs
@@ -11,6 +11,7 @@ use glutin_wgl_sys::wgl_extra::types::HPBUFFEREXT;
 use glutin_wgl_sys::wgl_extra::{self};
 use raw_window_handle::RawWindowHandle;
 use windows_sys::Win32::Foundation::{HWND, RECT};
+use windows_sys::Win32::Graphics::Gdi::HDC;
 use windows_sys::Win32::Graphics::{Gdi as gdi, OpenGL as gl};
 use windows_sys::Win32::UI::WindowsAndMessaging::GetClientRect;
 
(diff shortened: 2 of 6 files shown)


glutin/src/api/wgl/surface.rs
```diff
@@ -61,29 +97,71 @@ impl Display {
...
 // Impl only `Send` for Surface.
 unsafe impl<T: SurfaceTypeTrait> Send for Surface<T> {}
 
+impl<T: SurfaceTypeTrait> Surface<T> {
+    /// # Safety
+    ///
+    /// The caller must ensure that the attribute could be present.
+    unsafe fn raw_attribute(&self, attr: GLenum) -> c_int {
+        let mut value = 0;
+        unsafe {
```
kchibisov, line 137: Should use unsafe around the actual call to unsafe function.

glutin/src/api/wgl/surface.rs
```diff
@@ -129,6 +226,10 @@ impl<T: SurfaceTypeTrait> GlSurface<T> for Surface<T> {
     }
 
     fn set_swap_interval(&self, _context: &Self::Context, interval: SwapInterval) -> Result<()> {
+        let WGLSurface::Window(..) = self.raw else {
+            return Ok(());
+        };
```
kchibisov, line 231: we don't `let else` in general, use `match`.

glutin/src/api/wgl/surface.rs
```diff
@@ -61,29 +97,71 @@ impl Display {
...
 
         Ok(surface)
     }
 }
 
-/// A Wrapper around `HWND`.
+/// A wrapper around WGL surfaces.
+#[derive(Debug)]
+pub enum WGLSurface {
```
kchibisov, line 113: In general rust wants the `WglSurface` spelling. Also, this type should be defined below other types in this file.

glutin/src/api/wgl/surface.rs
```diff
@@ -36,10 +39,43 @@ impl Display {
...
+        let hdc = config.inner.hdc;
+        let width = surface_attributes.width.unwrap().get() as c_int;
+        let height = surface_attributes.height.unwrap().get() as c_int;
+
+        let mut attrs = Vec::<c_int>::with_capacity(2);
+        if surface_attributes.largest_pbuffer {
+            attrs.push(wgl_extra::PBUFFER_LARGEST_ARB as c_int);
+            attrs.push(1 as c_int);
+        }
+        attrs.push(0);
```
kchibisov, line 54: can use array of 3 zero init c_int, no need to allocate, also capacity should be 3, not 2, just fyi.

glutin/src/api/wgl/surface.rs
```diff
@@ -61,29 +97,71 @@ impl Display {
...
+    /// # Safety
+    ///
+    /// The caller must ensure that the attribute could be present.
+    unsafe fn raw_attribute(&self, attr: GLenum) -> c_int {
+        let mut value = 0;
+        unsafe {
+            match self.raw {
+                WGLSurface::Window(..) => unreachable!(),
+                WGLSurface::PBuffer(hbuf, _) => {
+                    if let Some(extra) = self.display.inner.wgl_extra {
```
kchibisov, line 141: In general, `pbuffer` can not be created without `wgl_extra` being present, so in here and in `drop` you can `unwrap`.

glutin/src/api/wgl/surface.rs
```diff
@@ -61,29 +97,71 @@ impl Display {
...
 }
 
 // Impl only `Send` for Surface.
 unsafe impl<T: SurfaceTypeTrait> Send for Surface<T> {}
 
+impl<T: SurfaceTypeTrait> Surface<T> {
+    /// # Safety
+    ///
+    /// The caller must ensure that the attribute could be present.
+    unsafe fn raw_attribute(&self, attr: GLenum) -> c_int {
```
kchibisov, line 135: I don't think that this function is unsafe by any means? Like if you pass wrong attribute it'll just return err or zero iirc.

glutin/src/api/wgl/surface.rs
```diff
@@ -61,29 +97,71 @@ impl Display {
...
+        let mut value = 0;
+        unsafe {
+            match self.raw {
+                WGLSurface::Window(..) => unreachable!(),
+                WGLSurface::PBuffer(hbuf, _) => {
+                    if let Some(extra) = self.display.inner.wgl_extra {
+                        extra.QueryPbufferARB(hbuf, attr as _, &mut value);
```
kchibisov, line 142: you should check whether the function actually succeeded.

glutin/src/api/wgl/surface.rs
```diff
@@ -36,10 +39,43 @@ impl Display {
...
+        let (hbuf, hdc) = match self.inner.wgl_extra {
+            Some(extra) if extra.CreatePbufferARB.is_loaded() => unsafe {
+                let hbuf = extra.CreatePbufferARB(
+                    hdc as _,
+                    config.inner.pixel_format_index,
+                    width,
+                    height,
+                    attrs.as_ptr(),
+                );
+                let hdc = extra.GetPbufferDCARB(hbuf);
```
kchibisov, line 65: You should ensure that the create succeeded.

See https://registry.khronos.org/OpenGL/extensions/ARB/WGL_ARB_pbuffer.txt
on how error is propagated and detected. The same for query function you're using.

kchibisov (changes requested):

glutin/src/api/wgl/surface.rs
SK83RJOSH: Perhaps not, I just noticed a trend that most `raw_attribute` helpers were marked as unsafe and continued the trend. Will fix this up though, as I agree, it doesn't make much sense. 🙂

glutin/src/api/wgl/surface.rs
```diff
@@ -61,29 +97,71 @@ impl Display {
...
 }
 
-/// A Wrapper around `HWND`.
+/// A wrapper around WGL surfaces.
+#[derive(Debug)]
+pub enum WGLSurface {
+    /// Surface backed by a window surface.
+    Window(HWND, gdi::HDC),
+    /// Surface backed by a pixel buffer.
+    PBuffer(HPBUFFEREXT, wgl_extra::types::HDC),
```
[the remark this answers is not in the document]
SK83RJOSH: Do you mean in that they both likely point to the same device context? Or that the types should be the same? 🙂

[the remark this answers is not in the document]
SK83RJOSH: Going to assume you meant the type definitions as say, unfortunately, they are not; one is defined as `void*` and the other `unsigned int` - but I suppose it's OK to treat them as the same in the interest of making things a bit cleaner.

glutin/src/api/wgl/surface.rs
SK83RJOSH: Indeed, I'm assuming it's okay to `.unwrap` inside of `Surface::raw_attribute` as well, in that case?

glutin/src/api/wgl/surface.rs
```diff
@@ -36,10 +40,52 @@ impl Display {
...
+        let hdc = config.inner.hdc;
+        let width = surface_attributes.width.unwrap().get() as c_int;
+        let height = surface_attributes.height.unwrap().get() as c_int;
+        let attrs: [c_int; 3] = {
+            if surface_attributes.largest_pbuffer {
+                [wgl_extra::PBUFFER_LARGEST_ARB as _, 1, 0]
+            } else {
+                [0; 3]
+            }
+        };
```
kchibisov, line 61: you can use the old logic like with `vec` and assign. Should be cleaner.

```rust
let mut attrs = [0; 3];
if surface_attributes.largest_pbuffer {
    attrs[0] = wgl_extra::PBUFFER_LARGEST_ARB;
    attrs[1] = 1;
}
```

glutin/src/api/wgl/surface.rs
```diff
@@ -173,15 +263,17 @@ impl<T: SurfaceTypeTrait> fmt::Debug for Surface<T> {
...
             .finish()
     }
 }
 
 impl<T: SurfaceTypeTrait> AsRawSurface for Surface<T> {
     fn raw_surface(&self) -> RawSurface {
-        RawSurface::Wgl(self.hwnd as _)
+        match self.raw {
+            WglSurface::Window(hwnd, _) => RawSurface::Wgl(hwnd as _),
+            WglSurface::PBuffer(..) => RawSurface::Wgl(0 as _),
```
kchibisov, line 275: returning `zero` pointer just looks wrong. I think it should be the `HPBUFFEREXT` here?

glutin/src/api/wgl/context.rs
```diff
@@ -387,7 +387,12 @@ impl ContextInner {
 
     fn make_current<T: SurfaceTypeTrait>(&self, surface: &Surface<T>) -> Result<()> {
         unsafe {
-            if wgl::MakeCurrent(surface.hdc as _, self.raw.cast()) == 0 {
+            let hdc = match surface.raw {
+                WglSurface::Window(_, hdc) => hdc as _,
+                WglSurface::PBuffer(_, hdc) => hdc as _,
+            };
```
kchibisov, line 393: you can use your `hdc()` method here.

kchibisov (changes requested):

glutin/src/api/wgl/surface.rs
SK83RJOSH: That was my first pass ironically, though the let was a bit cleaner since it keeps attrs const, but will address now 🙂

glutin/src/api/wgl/surface.rs
SK83RJOSH: I thought so too, but I was a bit worried that people might be relying on the old behavior, so returning null at least means there shouldn't be any breakage if people were being sane and checking validity of handles 🙂 

With that said, I can make that change if you'd like.

glutin/src/api/wgl/context.rs
SK83RJOSH: Nice catch, knew I missed something ^^

glutin/src/api/wgl/surface.rs
kchibisov: Returning `null` is not sane because you created an _object_ and user would expect a valid pointer, but they'll get `null`. You generally shouldn't null check an object you've created and know that is non null.

SK83RJOSH: I suppose it's ultimately a matter of taste, I lean towards being particularly defensive, especially with external libraries, and thus sanity check these sorts of things at every opportunity - but I'm happy to make the change regardless. Users can/should probably be expected to check the underlying surface type anyway, and no one will run into this change unless they opt-in to using the new functionality. 🙂

kchibisov (changes requested):
Could you update doc here https://github.com/rust-windowing/glutin/blob/master/glutin/src/surface.rs#L522 with the pbuffer that could be present there?

Other than that looks good.

SK83RJOSH:
Done! Thank you for the thorough review, and my apologies for the sorry state of this initially. Have a great rest of your weekend. 🙂

glutin/src/surface.rs
```diff
@@ -519,7 +519,7 @@ pub enum RawSurface {
     #[cfg(glx_backend)]
     Glx(u64),
 
-    /// HWND
+    /// Either an `HWND` or `HPBUFFEREXT` depending on `SurfaceType`.
```
kchibisov, line 522: You can [`SurfaceType`] link directly, so users can click on the link.

SK83RJOSH: TIL 😄

glutin/src/surface.rs
```diff
@@ -519,7 +519,7 @@ pub enum RawSurface {
     #[cfg(glx_backend)]
     Glx(u64),
 
-    /// HWND
+    /// Either an `HWND` or `HPBUFFEREXT` depending on [`SurfaceType`].
```
kchibisov, line 522:
```suggestion
    /// Either a `HWND` or `HPBUFFEREXT` depending on [`SurfaceType`].
```

sorry, I hope that's the last one.

SK83RJOSH: This one is actually a point of contention - in American English the `h` in `haitch` is silent... thus my decision to use an here, as I'd pronounce it as "an aitch wind"... but this is not a hill I want to die on :)

kchibisov: I mean, I know this rule, it's not silent here since it's an abbreviation and all syllables are pronounced as the should. `an` should be used when the `h` is like in `hour`, so not pronounced, here it's `hardware` iirc if you want to be more pedantic, but it's still pronounced.

kchibisov (approved):

glutin/src/surface.rs
SK83RJOSH: Aye, you're correct that's how it *should* pronounced given it's an acronym/abbreviation - but I don't think I've ever heard anyone put emphasis on an `H` when it's not followed by a vowel.

Ultimately this is kind of a `gif` vs. `jif` thing (`gif` being "correct", yet the one to coin the moniker still insists it's `jif`) so it's not worth losing any sleep over. Down to personal taste, so as I said, not a hill worth dying on today. 😛

kchibisov: I think `hardware`(`hard` in general) is a good example of it, since it's not `ardware` or `ard`.

SK83RJOSH: https://learn.microsoft.com/en-us/windows/apps/develop/ui-input/retrieve-hwnd

PR #1702, opened 2024-09-21
feat: add openharmony support

richerfu:

Add openharmony platform support, now we can build it successfully with [XComponent](https://developer.huawei.com/consumer/en/doc/harmonyos-guides-V5/napi-xcomponent-guidelines-V5). It doesn't effect any other code, so i think i needn't to add more docs or tests.

And you can find a full example with [example](https://github.com/ohos-rs/ohos-native-bindings/tree/master/examples/xcomponent). If you want more info, please let me know, thanks !

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Added `Device::drm_device_node_path()` and `Device::drm_render_device_node_path()` getters to EGL via `EGL_EXT_device_drm`.
 - Added support for `DrmDisplayHandle` in EGL's `Display::with_device()` using `EGL_DRM_MASTER_FD_EXT` from `EGL_EXT_device_drm`.
 - Properly set up OpenGL-specific stuff on the `NSView`, instead of relying on Winit to do it.
-- Add `OpenHarmony` platform support.
+- Added `OpenHarmony` platform support with EGL.
 
 # Version 0.32.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Added `Device::drm_device_node_path()` and `Device::drm_render_device_node_path()` getters to EGL via `EGL_EXT_device_drm`.
 - Added support for `DrmDisplayHandle` in EGL's `Display::with_device()` using `EGL_DRM_MASTER_FD_EXT` from `EGL_EXT_device_drm`.
 - Properly set up OpenGL-specific stuff on the `NSView`, instead of relying on Winit to do it.
+- Add `OpenHarmony` platform support.
 
 # Version 0.32.0
 
--- a/glutin/build.rs
+++ b/glutin/build.rs
@@ -5,11 +5,12 @@ fn main() {
     cfg_aliases! {
         // Systems.
         android_platform: { target_os = "android" },
+        ohos_platform: { target_env = "ohos" },
         wasm_platform: { target_family = "wasm" },
         macos_platform: { target_os = "macos" },
         ios_platform: { target_os = "ios" },
         apple: { any(ios_platform, macos_platform) },
-        free_unix: { all(unix, not(apple), not(android_platform)) },
+        free_unix: { all(unix, not(apple), not(android_platform), not(ohos_platform)) },
 
         // Native displays.
         x11_platform: { all(feature = "x11", free_unix, not(wasm_platform)) },
--- a/glutin/Cargo.toml
+++ b/glutin/Cargo.toml
@@ -23,7 +23,7 @@ wayland = ["wayland-sys", "egl"]
 bitflags = "2.2.1"
 libloading = { version = "0.8.0", optional = true }
 once_cell = "1.13"
-raw-window-handle = "0.6"
+raw-window-handle = "0.6.2"
 
 [target.'cfg(windows)'.dependencies]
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Added `Device::drm_device_node_path()` and `Device::drm_render_device_node_path()` getters to EGL via `EGL_EXT_device_drm`.
 - Added support for `DrmDisplayHandle` in EGL's `Display::with_device()` using `EGL_DRM_MASTER_FD_EXT` from `EGL_EXT_device_drm`.
 - Properly set up OpenGL-specific stuff on the `NSView`, instead of relying on Winit to do it.
+- Add `OpenHarmony` platform support.
```
kchibisov, line 10: This will need a rebase, and state that it's for EGL.

kchibisov (changes requested):
You should also explicitly set that we depend on `raw-window-handle 0.6.2`, since the build will break otherwise.

richerfu:
> raw-window-handle

Should i update the `raw-window-handle` version to 0.6.2? Looks not..

kchibisov:
>Should i update the raw-window-handle version to 0.6.2? Looks not..

open harmony was added in 0.6.2, so you should explicitly specify 0.6.2

richerfu:
> > Should i update the raw-window-handle version to 0.6.2? Looks not..
> 
> open harmony was added in 0.6.2, so you should explicitly specify 0.6.2

Done

glutin-winit/Cargo.toml
```diff
@@ -20,7 +20,7 @@ wayland = ["glutin/wayland", "winit/wayland"]
 
 [dependencies]
 glutin = { version = "0.32.0", path = "../glutin", default-features = false }
-raw-window-handle = "0.6"
+raw-window-handle = "0.6.2"
```
kchibisov, line 23: You don't need to update it here.

glutin_examples/Cargo.toml
```diff
@@ -22,7 +22,7 @@ wayland = ["glutin-winit/wayland", "winit/wayland-dlopen", "winit/wayland-csd-ad
 glutin = { path = "../glutin", default-features = false }
 glutin-winit = { path = "../glutin-winit", default-features = false }
 png = { version = "0.17.6", optional = true }
-raw-window-handle = "0.6"
+raw-window-handle = "0.6.2"
```
kchibisov, line 25: Don't need to update here as well.

kchibisov (changes requested):

glutin_examples/Cargo.toml
richerfu: Reverted

kchibisov (approved):

glutin/build.rs
```diff
@@ -5,11 +5,12 @@ fn main() {
     cfg_aliases! {
         // Systems.
         android_platform: { target_os = "android" },
+        ohos_platform: { target_env = "ohos" },
```
MarijnS95, line 8: Glutin's CI is quite extensive. @richerfu would you be up for adding one or two `ohos` target triples to our CI with the `egl` feature, so that we can always build-test this functionality and make sure this target environment is not accidentally broken in the future?

richerfu: OpenHarmony currently has no environment to directly run `cargo test`, and all code testing is forcibly dependent on real machines. So there is no way to do some testing or verification in the CI environment for the time being. According to official news, this capability will be supported as early as Q1 2025.

MarijnS95: @richerfu I don't think we need run-time tests, just build-testing (i.e. `cargo check/build/clippy`) is enough _for now_ to at least guarantee a minimum level of source compatibility. We can incrementally move towards testing on real machines _if that possibility is provided_.

For example, this works for me:

```console
$ rustup target add x86_64-unknown-linux-ohos
$ cargo clippy --all-features --target x86_64-unknown-linux-ohos -p glutin
```

And if I make a silly change, like commenting out a `Self::Ohos` `match` arm, the compiler/CI could complain that the `ohos` target no longer even compiles:

```console
$ cargo clippy --all-features --target x86_64-unknown-linux-ohos -p glutin
error[E0004]: non-exhaustive patterns: `api::egl::surface::NativeWindow::Ohos(_)` not covered
   --> glutin/src/api/egl/surface.rs:541:15
    |
541 |         match *self {
    |               ^^^^^ pattern `api::egl::surface::NativeWindow::Ohos(_)` not covered
    |
note: `api::egl::surface::NativeWindow` defined here
   --> glutin/src/api/egl/surface.rs:446:6
    |
446 | enum NativeWindow {
    |      ^^^^^^^^^^^^
...
460 |     Ohos(*mut ffi::c_void),
    |     ---- not covered
    = note: the matched value is of type `api::egl::surface::NativeWindow`
help: ensure that all possible cases are being handled by adding a match arm with a wildcard pattern or an explicit pattern as shown
    |
541 ~         match *self {
542 +             api::egl::surface::NativeWindow::Ohos(_) => todo!(),
543 +         }
    |

For more information about this error, try `rustc --explain E0004`.
error: could not compile `glutin` (lib) due to 1 previous error
```

richerfu: Ok, i'll try to add some code and upgrade CI config later.

PR #1703, opened 2024-09-24
feat(ci): add OpenHarmony ci support

richerfu:
Add minimal OpenHarmony CI testing as requested in https://github.com/rust-windowing/glutin/pull/1702#discussion_r1771122157.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,6 +117,6 @@ jobs:
 
     - name: Lint with clippy
       shell: bash
-      if: matrix.rust_version == '1.70.0' && !contains(matrix.platform.options, '--no-default-features')
+      if: matrix.rust_version == '1.78.0' && !contains(matrix.platform.options, '--no-default-features')
       run: cargo clippy --workspace --all-targets --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES -- -Dwarnings
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,8 +55,8 @@ jobs:
           - { target: x86_64-unknown-linux-gnu,   os: ubuntu-latest, options: --no-default-features, features: "egl,wayland" }
           - { target: x86_64-unknown-linux-gnu,   os: ubuntu-latest, options: --no-default-features, features: "egl,x11" }
           - { target: x86_64-unknown-linux-gnu,   os: ubuntu-latest, options: --no-default-features, features: glx }
-          - { target: aarch64-unknown-linux-ohos, os: ubuntu-latest, options: --no-default-features, features: egl }
-          - { target: x86_64-unknown-linux-ohos,  os: ubuntu-latest, options: --no-default-features, features: egl }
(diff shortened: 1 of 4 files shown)


richerfu:
The CI of ohos needs to do some extra processing, and I will explain why it is implemented in this way:

1. First of all, the construction of ohos is not natively supported, we need to do some environment initialization.[ohrs](https://ohos.rs/en/docs/cli/init.html) is a tool that can help us simplify the use of cargo commands under the ohos build environment.
2. After the rust version of 1.78.0, ohos is only a tier2 stage target in rust, so it is difficult for us to simply run cargo commands in rust 1.70.0 for ohos, so we will skip the corresponding clippy detection, unless we update them all to versions after 1.78.0.

Could you provide some advice? @MarijnS95

MarijnS95:
1. That is only needed for compilation and linking right? I suggest we only run `clippy` to make sure things are in a compilable state, which seems to work in a cross-compilation environment as outlined in https://github.com/rust-windowing/glutin/pull/1702#discussion_r1771122157.
2. Hmm that's unforunate. We could try running OHOS clippy on 1.78 though newer linter warnings may show up. It looks like this version-lock may be used to enforce our MSRV though I'm not opposed to warn on and address lints on newer Rust versions (which may be nontrivial when `clippy` suggests changes that require a higher MSRV).

richerfu:
Yep, so should we upgrade MSRV to 1.78.0? If just run `clippy`, some steps can be removed.

richerfu:
Evenything seems ok now. 

I upgrade rust version to 1.78.0 for CI and keep MSRV for crates.  `wgl_binding` gets a warning in `1.78.0` and fixed it.
image

@MarijnS95

kchibisov:
tbf, instead of adding more targets, I'd rather remove `cfg`, since rwh doesn't need any cfg at all. Like there's nothing wrong in always compiling a bit more code, but you'd know that it's _properly tested_.

If you pass e.g. Android handle on general linux stuff, nothing wrong will happen, you'll just get an error and that's about it.

MarijnS95:
@kchibisov I've previously tried to add `cfg` guards consistently because they were only put in place "randomly", and found that it was also necessary to disambiguate i.e. the type of a global `let` binding (IIRC a `Vec`). In the end I ended up [removing most `cfg` guards](https://github.com/rust-windowing/glutin/commit/f27377a88fe55426dc412c7dbce38a27e00812b5) that were hiding code which would compile on all platforms. Glad to see/hear that trend continue.

---

> I upgrade rust version to 1.78.0 for CI and keep MSRV for crates. wgl_binding gets a warning in 1.78.0 and fixed it.

Please don't post screenshots but copy-paste logs/code/links, as they are hard to find/reference/quote otherwise.
Maybe we should check why the generator emits unused bindings?

MarijnS95:
@kchibisov I think you are saying the same thing. See the following example, where the `cfg`s for X11/Wayland are not needed because the code can compile without it anyway as it does not reference any of the deps:

https://github.com/rust-windowing/glutin/blob/dae89a5a9145853220385e0d5036319689afa13c/glutin/src/api/egl/display.rs#L257-L271

kchibisov:
yeah, that's fine then.

kchibisov:
Handled in #1707 

We just always build the code for harmony now.

PR #1705, opened 2024-10-02
Add PossiblyCurrentContext::make_not_current_as_possibly_current

### LINKED ISSUE #1704: Expose non-moving `PossiblyCurrentContext::make_not_current`
I want to make the context "not current" but I do not need to send the context to another thread. Can we have a non-moving variant of `PossiblyCurrentContext::make_not_current` + `NotCurrentContext::treat_as_possibly_current` that can be called on a reference?

I believe none of the context implementations need the context by-value, it's just the type-level change from `PossiblyCurrentContext` to `NotCurrentContext` that does. It seems like there could exist a method `PossiblyCurrentContext::make_not_current_in_place(&self) -> Result<()>`.

I understand that there are work-arounds for this. The easy way around this is to stuff the context into an `Option`. A probably unsound workaround is:

```rust
unsafe {
    let context = std::ptr::read(std::ptr::from_ref(&self.context));
    let Ok(context) = context.make_not_current() else {
        eprintln!("failed to make context not current");
        std::process::abort();
    };
    std::mem::forget(std::mem::replace(
        &mut self.context,
        context.treat_as_possibly_current(),
    ))
}
```

My motivation is as follows. I built an abstraction over winit and glutin that makes it a bit easier to create throw-away demos. The application lifecycle is abstracted as illustrated below:

image

Because the resumed and suspended states have their own types, I can use `PossiblyCurrentContext` in the resumed state type and `NotCurrentContext` in the suspended state type. 

However, the application lifecycle is generic over the rendering back-end (OpenGL, Vulkan, ...) and the back-end stores the context. This means that I am forcing the creation of two separate state types for the rendering-back end implementation, just because `PossiblyCurrentContext::make_not_current_in_place(&self)` is not available.

I understand that there is a cost to expanding the API surface of glutin, and that this change would restrict future backends by enforcing that the un-currenting of contexts can no longer take the context by value. However, it does seem like the combination of "un-current" and "treat as possibly current" is fairly common.

mickvangelderen:
Fixes #1704 

Allow making the context not current by reference.

### DIFF
--- a/glutin/src/context.rs
+++ b/glutin/src/context.rs
@@ -97,6 +97,11 @@ pub trait PossiblyCurrentGlContext: Sealed {
     /// - **macOS: this will block if your main thread is blocked.**
     fn make_not_current(self) -> Result<Self::NotCurrentContext>;
 
+    /// Make the context not current to the current thread. If you need to
+    /// send the context toanother thread, use [`Self::make_not_current`]
+    /// instead.
+    fn make_not_current_in_place(&self) -> Result<()>;
+
     /// Make [`Self::Surface`] current on the calling thread.
     ///
     /// # Platform specific
@@ -515,6 +520,10 @@ impl PossiblyCurrentGlContext for PossiblyCurrentContext {
         Ok(
             gl_api_dispatch!(self; Self(context) => context.make_not_current()?; as NotCurrentContext),
         )
+    }
+
+    fn make_not_current_in_place(&self) -> Result<()> {
+        Ok(gl_api_dispatch!(self; Self(context) => context.make_not_current_in_place()?))
     }
 
     fn make_current<T: SurfaceTypeTrait>(&self, surface: &Self::Surface<T>) -> Result<()> {
--- a/glutin/src/context.rs
+++ b/glutin/src/context.rs
@@ -98,7 +98,7 @@ pub trait PossiblyCurrentGlContext: Sealed {
     fn make_not_current(self) -> Result<Self::NotCurrentContext>;
 
     /// Make the context not current to the current thread. If you need to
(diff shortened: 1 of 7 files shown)


kchibisov (changes requested):
I feel like `in_place` is a bit better, since it's just shorter...

Other than that should be good.

mickvangelderen:
I rebased to redo the commit message after the name-change

kchibisov (changes requested):
I completely forgot, could you change the `src/lib.rs` in examples to also use this function and not `Option`?

mickvangelderen:
I think the example code can be reworked to match the diagram I sketched in #1704. There are a number of `Option<T>`'s that together define a state machine. One thing I noticed is that it is possible that the renderer will be asked to resize while the application is suspended and the context is not current. I am not sure this is a problem, but the fact that this is not obvious is a result of the current type-level design.

If you are curious what I mean, I'm happy to take a shot at improving the example code and showing you what I mean. For this PR however, I think this is good enough.

glutin/src/context.rs
```diff
@@ -97,6 +97,11 @@ pub trait PossiblyCurrentGlContext: Sealed {
     /// - **macOS: this will block if your main thread is blocked.**
     fn make_not_current(self) -> Result<Self::NotCurrentContext>;
 
+    /// Make the context not current to the current thread. If you need to
+    /// send the context toanother thread, use [`Self::make_not_current`]
```
kchibisov, line 101:
```suggestion
    /// send the context to another thread, use [`Self::make_not_current`]
```

Just noticed when decided to re-read.

kchibisov (changes requested):

PR #1706, opened 2024-10-02
Rework glutin example

mickvangelderen:
In my opinion the main application implementation of the example is hard to follow due to the combinatory state machine. This refactor eliminates many possible combinations from the state space through a simple explicit state machine. 

### DIFF
--- a/glutin_examples/src/lib.rs
+++ b/glutin_examples/src/lib.rs
@@ -1,6 +1,6 @@
 use std::error::Error;
 use std::ffi::{CStr, CString};
-use std::num::NonZeroU32;
+use std::num::{NonZero, NonZeroU32};
 use std::ops::Deref;
 
 use gl::types::GLfloat;
@@ -15,7 +15,7 @@ use glutin::config::{Config, ConfigTemplateBuilder, GetGlConfig};
 use glutin::context::{
     ContextApi, ContextAttributesBuilder, NotCurrentContext, PossiblyCurrentContext, Version,
 };
-use glutin::display::GetGlDisplay;
+use glutin::display::{Display, GetGlDisplay};
 use glutin::prelude::*;
 use glutin::surface::{Surface, SwapInterval, WindowSurface};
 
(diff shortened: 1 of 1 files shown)


kchibisov:
I'm not sure about this one, since it gets more involved.

I think I'd tune the glutin API a bit for 0.33 to explicitly state that it owns the display and that way I can remove a lot of code from the example.

glutin_examples/src/lib.rs
```diff
@@ -167,32 +116,65 @@ impl ApplicationHandler for App {
...
         #[cfg(egl_backend)]
         #[allow(irrefutable_let_patterns)]
-        if let glutin::display::Display::Egl(display) = _gl_display {
+        if let glutin::display::Display::Egl(display) = self.0.display() {
             unsafe {
                 display.terminate();
             }
         }
     }
+}
```
mickvangelderen, line 168: @kchibisov do you mean that this egl specific display terminate code should be handled by glutin by default?

glutin_examples/src/lib.rs
```diff
@@ -140,23 +72,40 @@ impl ApplicationHandler for App {
...
+                            AppState::Uninitialized { .. } => {
+                                return Err("invalid transition".into())
+                            },
+                            AppState::Resumed(state) => {
+                                state.resize(width, height);
+                            },
+                            AppState::Suspended(AppStateSuspended { renderer, .. }) => {
+                                // TODO: Should we call resize while suspended or should we not do
+                                // anything here?
+                                renderer.resize(size.width as i32, size.height as i32);
```
mickvangelderen, line 90: Should we call resize while suspended or should we not do anything here?

glutin_examples/src/lib.rs
```diff
@@ -167,35 +116,75 @@ impl ApplicationHandler for App {
...
         #[cfg(egl_backend)]
         #[allow(irrefutable_let_patterns)]
-        if let glutin::display::Display::Egl(display) = _gl_display {
+        if let glutin::display::Display::Egl(display) = self.0.display() {
             unsafe {
                 display.terminate();
             }
         }
     }
+}
```
mickvangelderen, line 171: These 50 lines can be removed if glutin can just do this for the egl backend. I am not sure about when `display.terminate()` can be called though, does the window have to be dropped before? After?

glutin_examples/src/lib.rs
```diff
@@ -237,41 +237,133 @@ fn window_attributes() -> WindowAttributes {
...
+    // NOTE: Window should be dropped after all resources created using its raw-window-handle.
     gl_surface: Surface<WindowSurface>,
-    // NOTE: Window should be dropped after all resources created using its
-    // raw-window-handle.
     window: Window,
+    // TODO: Does the context need to be dropped after the window? Specifically, the egl display
+    // terminate code, when should it run?
+    gl_context: ContextTerminateDisplay<PossiblyCurrentContext>,
+    renderer: Renderer,
+}
```
mickvangelderen, line 301: What is the correct drop order for all of this?

glutin_examples/src/lib.rs
```diff
@@ -237,41 +237,133 @@ fn window_attributes() -> WindowAttributes {
...
+    fn initialize(self, event_loop: &ActiveEventLoop) -> Result<AppStateResumed, Box<dyn Error>> {
+        let Self { template_builder, display_builder } = self;
+        let (window, gl_config) =
+            display_builder.build(event_loop, template_builder, gl_config_picker)?;
+        let window = window.ok_or("failed to create window")?;
+        println!("Picked a config with {} samples", gl_config.num_samples());
+        // TODO: Can we create the context and the surface in any order?
+        let gl_context = create_gl_context(&window, &gl_config);
+        let gl_surface = create_gl_surface(&window, &gl_config)?;
+        let gl_context = gl_context.make_current(&gl_surface)?;
```
mickvangelderen, line 318: Can the context and surface be created in any order?

mickvangelderen:
> I'm not sure about this one, since it gets more involved.

@kchibisov I understand. I do think that there are some things in this PR that could lead to improvements in glutin or the examples.

glutin_examples/src/lib.rs
kchibisov: Yes, it should.

mickvangelderen: It seems like the terminate call is only necessary if the reference counting extensions are not available: https://github.com/rust-windowing/glutin/issues/1588. 

The `DisplayInner` implementation will terminate the display only if reference counting is supported and the reference is the last one.

The terminate call in the example is only correct if reference counting is not supported AND we guarantee that there are no uses of the display after termination.

mickvangelderen: The only solution I can think of is to emulate reference counting (through a global static lookup table) if the extension is not available.

kchibisov: >The terminate call in the example is only correct if reference counting is not supported AND we guarantee that there are no uses of the display after termination.

Which we do, and even if we use ref-counting, calling terminate twice is safe.

mickvangelderen: Something like

```rust
use std::{collections::HashMap, sync::{atomic::{AtomicIsize, Ordering}, Arc, Mutex, OnceLock, Weak}};

// Pretend this is the external library.
mod external {
    use std::sync::atomic::{AtomicIsize, Ordering};

    static DISPLAY_GENERATOR: AtomicIsize = AtomicIsize::new(1);
    
    pub fn create() -> isize {
        DISPLAY_GENERATOR.fetch_add(1, Ordering::SeqCst)
    }

    pub fn destroy(_id: isize) {
        // TODO: Enable the external library to hand out the same id multiple times.
    }
}

struct Display(Arc<DisplayInner>);

impl Display {
    pub fn new() -> Self {
        let id = DisplayId(external::create());
        let inner = Arc::new(DisplayInner { id });
        assert!(display_map().lock().unwrap().insert(id, Arc::downgrade(&inner)).is_none());
        Self(inner)
    }

    pub fn from_id(id: DisplayId) -> Option<Display> {
        display_map().lock().unwrap().get(&id).and_then(|weak| weak.upgrade().map(Self))
    }
    
    pub fn id(&self) -> DisplayId {
        self.0.id
    }
}


struct DisplayInner {
    id: DisplayId,
}

impl Drop for DisplayInner {
    fn drop(&mut self) {
        // Terminate

        // Remove entry from hashmap, optional.
        display_map().lock().unwrap().remove(&self.id);
    }
}

#[derive(Debug, Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Hash)]
struct DisplayId(isize);

type DisplayMap = HashMap<DisplayId, Weak<DisplayInner>>;

static DISPLAY_MAP: OnceLock<Mutex<DisplayMap>> = OnceLock::new();

fn display_map() -> &'static Mutex<DisplayMap> {
    DISPLAY_MAP.get_or_init(|| Mutex::new(DisplayMap::new()))
}

fn main() {
    
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test() {
        let d1 = Display::new();
        let id = d1.id(); 
        let d2 = Display::from_id(id).unwrap();
        assert_eq!(std::ptr::from_ref(&*d1.0), std::ptr::from_ref(&*d2.0));
        drop(d2);
        let d2 = Display::from_id(id).unwrap();
        assert_eq!(std::ptr::from_ref(&*d1.0), std::ptr::from_ref(&*d2.0));
        drop(d1);
        drop(d2);
        assert!(Display::from_id(id).is_none());
    }
}
```

kchibisov: Maybe something like that is not a bad idea, since I don't see why we should deny the ability to create display from non-main thread if you don't want to pass things around from the main thread.

And once you have have `Display::from_raw(ptr)` we'd mark that we mustn't terminate this pointer.

glutin_examples/src/lib.rs
MarijnS95: Theoretically there should:

- Not be a resized event after `Suspended` (or before `Resumed`);
- Not be a render `Surface` to resize to begin with (but perhaps user code keeps persistent backing buffers around to reuse on a `Resumed`).

kchibisov: All the buffers are in the context, so during suspend -> resume there's no surface to issue a resize.

glutin_examples/src/lib.rs
mickvangelderen: I think we need `unsafe from_raw_assume_ownership(raw_window_handle::RawDisplayHandle)` for this to work.

kchibisov: if you want like that, sure, but generally building a display from raw pointer means that some other lib created it as well.

glutin_examples/src/lib.rs
mickvangelderen: The current implementation could call glViewport while the context is definitely not current, but allegedly the resized event is not triggered after suspend and so the problem does not actually occur. The proposed implementation can be updated to mark the transition as invalid.

glutin_examples/src/lib.rs
mickvangelderen: How many `EGLDisplay`'s can exist per "native display" (`NativeDisplayType` for `eglGetDisplay` and `void *` for `eglGetPlatformDisplay`)?

kchibisov: It's generally one, `EGLDisplay` is a global singletone.

PR #1707, opened 2024-10-03
egl: remove cfg from code not needing them

kchibisov:
--

@MarijnS95 I think something like that?

### DIFF
--- a/glutin/src/api/egl/config.rs
+++ b/glutin/src/api/egl/config.rs
@@ -327,21 +327,15 @@ impl GlConfig for Config {
         unsafe { self.raw_attribute(egl::CONFIG_CAVEAT as EGLint) != egl::SLOW_CONFIG as EGLint }
     }
 
-    #[cfg(not(any(wayland_platform, x11_platform)))]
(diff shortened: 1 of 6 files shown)


MarijnS95 (approved):

MarijnS95:
I also wonder if we need optimizations like:

https://github.com/rust-windowing/glutin/blob/5a7783c3290a498a17a23030d0aa814bb9febd55/glutin/src/api/egl/config.rs#L330-L347

This should compile down to `match ...? { _ => None }` which can only ever return `None` anyway.

kchibisov:
that's to avoid clippy lints and is easier to read.

MarijnS95:
Didn't we typically disable that lint because this "single match arm" occurs in many more places in `glutin`?

I don't find the (negated) `cfg`s particularly easy to read nor modify when new variants ever need to be added.

kchibisov:
We don't display it and it's never triggered., you can grep for it

kchibisov:
Seems like clippy doesn't complain about it, so I guess it's fine.

MarijnS95 (approved):

PR #1710, opened 2024-10-24
Add MacOS (Cgl) make_current_surfaceless() implementation

bschwind:

I have tested this on a complex OpenGL application which uses multiple "surfaceless" contexts on many threads while sharing textures and having those textures displayed on the `main` thread with a windowed OpenGL context, and everything appears to function correctly.

### DIFF
--- a/glutin/src/api/cgl/context.rs
+++ b/glutin/src/api/cgl/context.rs
@@ -255,6 +255,10 @@ impl ContextInner {
             self.update();
             self.raw.makeCurrentContext();
 
+            run_on_main(|_mtm| unsafe {
+                self.raw.setView(None);
+            });
+
             Ok(())
         })
     }
--- a/glutin/src/api/cgl/context.rs
+++ b/glutin/src/api/cgl/context.rs
@@ -74,6 +74,13 @@ pub struct NotCurrentContext {
 }
 
 impl NotCurrentContext {
+    /// Make a [`Self::PossiblyCurrentContext`] indicating that the context
+    /// could be current on the thread.
+    pub fn make_current_surfaceless(self) -> Result<PossiblyCurrentContext> {
+        self.inner.make_current_surfaceless()?;
+        Ok(PossiblyCurrentContext { inner: self.inner, _nosendsync: PhantomData })
+    }
+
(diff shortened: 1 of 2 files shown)


glutin/src/api/cgl/context.rs
```diff
@@ -236,6 +250,15 @@ impl ContextInner {
         })
     }
 
+    fn make_current_surfaceless(&self) -> Result<()> {
+        autoreleasepool(|_| {
+            self.update();
+            self.raw.makeCurrentContext();
```
kchibisov, line 256: Shouldn't you explicitly set `self.view.setView(None)` as well to account for previous `make_current` on this thread?

bschwind: Perhaps so, let me give it a try.

bschwind: I tested and it continues to work well. I'm guessing it starts out as `Nil` on `NSOpenGLContext` but good to be explicit here.

kchibisov: the point was when you can it with `setView(Some)` before at some point, so we should likely clear it.

bschwind: @kchibisov would you consider this comment chain resolved?

kchibisov (approved):
--

next time would suggest to re-request review (ui on the right).

bschwind:
> next time would suggest to re-request review (ui on the right).

Roger that!

kchibisov:
There's also the same APIs for `glx`/`wgl` from what I can tell, so we may have it at top-level at some point.

it's just they both need extensions, so display feature may be needed for users to configure their pipeline beforehand.

bschwind:
> There's also the same APIs for glx/wgl from what I can tell, so we may have it at top-level at some point.

Yep, a colleague of mine has added similar functionality for `glx`, so we'll probably have another PR up for that soon. I currently don't have a windows machine so I can't test `wgl` at the moment, but hopefully I'll get one soon.

kchibisov:
`wgl` should be the same as `glx` since they were designed by the same process.

PR #1711, opened 2024-10-29
glx: Support making a context current without surface

mbernat:
This mirrors the EGL behavior and follows in footsteps of #1710.

This change should be valid according to https://registry.khronos.org/OpenGL/extensions/ARB/GLX_ARB_create_context.txt (emphasis mine)
> "If either `<draw>` or `<read>` are not a valid GLX drawable, a
    GLXBadDrawable error is generated, unless **`draw` and `read` are both
    None** and the OpenGL version supported by `<ctx>` is 3.0 or greater. In
    this case the context is made current without a default framebuffer,
    as defined in chapter 4 of the OpenGL 3.0 Specification."

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Add `PossiblyCurrentContext::make_not_current_in_place(&self)` for when `Send` capability of `NotCurrentContext` is not required.
 - Add `NotCurrentContext::make_current_surfaceless(self)` and `PossiblyCurrentContext::make_current_surfaceless(&self)` in the `Cgl` implementation to allow the use of surfaceless contexts on MacOS.
+- Add `NotCurrentContext::make_current_surfaceless(self)` and `PossiblyCurrentContext::make_current_surfaceless(&self)` in the `Glx` implementation to allow the use of surfaceless contexts with GLX.
 
 # Version 0.32.1
 
--- a/glutin/src/api/glx/context.rs
+++ b/glutin/src/api/glx/context.rs
@@ -230,6 +230,8 @@ pub struct NotCurrentContext {
 impl NotCurrentContext {
     /// Make a [`Self::PossiblyCurrentContext`] indicating that the context
     /// could be current on the thread.
+    ///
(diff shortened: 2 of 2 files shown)


MarijnS95:
I presume this comes from the suggestion at https://github.com/rust-windowing/glutin/pull/1710#issuecomment-2440635928?

mbernat:
@MarijnS95 Yes, indeed. 

Didn't mean to open the PR just yet, hence the stump form of the PR request, but I'd be happy to see it through it properly.

MarijnS95:
@mbernat no worries, just figured we should update the title to include the `without surface` or `surfaceless` keyword :)

Happy that you included the documentation that clarifies that this is allowed, because `glxMakeContextCurrent()` does not (as expected).

Like the rest of `glutin`, how about we add some extra version/extension validation before calling the function like this function, and have appropriate error handling otherwise?

mbernat:
@MarijnS95 Let me know whether the validation I added is along the lines you had in mind.

I also implemented the API for WGL but the validation is harder to do because its `Display` doesn't carry a `version` info, unlike the other backends. Also, I don't really have a windows use case to test this against. But other than that I can post what I have so far as a separate PR and we can take it from there.

mbernat:
I have fixed the validation, the previous commit was using the GLX version stored in `Display` rather than the OpenGL version as determined by context attributes and profiles (oops).

It's a bit hairy now since the version is only computed in the middle of the context creation and I needed to get the information about surfaceless support out from there; it would be better to compute the version much earlier. Let me know and I'd be happy to try refactoring it.

Other than that, this should be ready for review.

@MarijnS95 @kchibisov please take a look whenever it's convenient for you 🙏 
Not urgent but I'm not very experienced with open source and the review process here, so just a gentle reminder 😇

strohel:
I've just successfully tested this one (in a proprietary application).

kchibisov (approved):

kchibisov:
Just for the reference, querying the context version afterwards is not really possible, and it's just the same as it was passed to the API.

PR #1712, opened 2024-10-29
glutin-winit: try to inject screen_id into RawDisplayHandle

mbernat:
- Use tonarino/winit to support creating windows on non-default screens

This is a hack, so not to be upstreamed as is.


kchibisov:
You can write integration for your own case separately.

mbernat:
Terribly sorry, didn't mean to try upstreaming this change 🙏

PR #1714, opened 2024-11-02
wgl: add `make_current_surfaceless`

kchibisov:
This follows the EGL api to make surfaceless platforms.

--

I haven't tested in on windows, but it's the exact same thing as with glx.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,16 @@
 # Unreleased
 
-- Add `PossiblyCurrentContext::make_not_current_in_place(&self)` for when `Send` capability of `NotCurrentContext` is not required.
-- Add `NotCurrentContext::make_current_surfaceless(self)` and `PossiblyCurrentContext::make_current_surfaceless(&self)` in the `Cgl` implementation to allow the use of surfaceless contexts on MacOS.
-- Add `NotCurrentContext::make_current_surfaceless(self)` and `PossiblyCurrentContext::make_current_surfaceless(&self)` in the `Glx` implementation to allow the use of surfaceless contexts with GLX.
+- Added `PossiblyCurrentContext::make_not_current_in_place(&self)` for when `Send` capability of `NotCurrentContext` is not required.
+- Added `NotCurrentContext::make_current_surfaceless(self)` and
+  `PossiblyCurrentContext::make_current_surfaceless(&self)` in the `Cgl`
+  implementation to allow the use of surfaceless contexts on MacOS.
+- Added `NotCurrentContext::make_current_surfaceless(self)` and
+  `PossiblyCurrentContext::make_current_surfaceless(&self)` in the `Glx`
+  implementation to allow the use of surfaceless contexts with GLX.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,16 @@
...
+- Added `PossiblyCurrentContext::make_not_current_in_place(&self)` for when `Send` capability of `NotCurrentContext` is not required.
+- Added `NotCurrentContext::make_current_surfaceless(self)` and
+  `PossiblyCurrentContext::make_current_surfaceless(&self)` in the `Cgl`
+  implementation to allow the use of surfaceless contexts on MacOS.
+- Added `NotCurrentContext::make_current_surfaceless(self)` and
+  `PossiblyCurrentContext::make_current_surfaceless(&self)` in the `Glx`
+  implementation to allow the use of surfaceless contexts with GLX.
+- Added `NotCurrentContext::make_current_surfaceless(self)` and
+`PossiblyCurrentContext::make_current_surfaceless(&self)` in the `Wgl`
+implementation to allow the use of surfaceless contexts with WGL.
```
MarijnS95, line 12:
```suggestion
- Added `NotCurrentContext::make_current_surfaceless(self)` and
  `PossiblyCurrentContext::make_current_surfaceless(&self)` in the `Wgl`
  implementation to allow the use of surfaceless contexts with WGL.
```

MarijnS95, line 12: How about we merge these and say ``Added ... in the `Cgl`, Glx` and `Wgl` implementations``?

glutin/src/api/wgl/context.rs
```diff
@@ -45,27 +45,34 @@ impl Display {
...
 
         let config = config.clone();
         let is_gles = matches!(context_attributes.api, Some(ContextApi::Gles(_)));
-        let inner = ContextInner { display: self.clone(), config, raw: context, is_gles };
+        let inner = ContextInner {
+            display: self.clone(),
+            config,
+            raw: context,
+            is_gles,
+            supports_surfaceless,
```
MarijnS95, line 74: Is there a reason why we don't store the version and ARB vs "legacy" instead?

Otherwise let's do that when more features inevitably show up that require version/extension checks :)

MarijnS95 (approved):
Don't see why this wouldn't work, but I can probably spin up a Windows machine to test this if that would help.

CHANGELOG.md
kchibisov: I'll merge all of them in the follow-up. The rendering is also correct, so you don't really have to indent.

glutin/src/api/wgl/context.rs
kchibisov: I don't see a reason to store more data if we don't need it.

CHANGELOG.md
MarijnS95: Then de-indent the rest for consistency at least? Not all markdown renderers seem to treat it as a continuation of the unordered list item and instead start a new paragraph.

PR #1715, opened 2024-11-12
api: unify surfaceless APIs on context

kchibisov:
The API was implemented by all backend, this patch simply promotes it to the top-level trait.

### DIFF
--- a/glutin/src/api/cgl/context.rs
+++ b/glutin/src/api/cgl/context.rs
@@ -74,13 +74,6 @@ pub struct NotCurrentContext {
 }
 
 impl NotCurrentContext {
-    /// Make a [`Self::PossiblyCurrentContext`] indicating that the context
-    /// could be current on the thread.
-    pub fn make_current_surfaceless(self) -> Result<PossiblyCurrentContext> {
-        self.inner.make_current_surfaceless()?;
-        Ok(PossiblyCurrentContext { inner: self.inner, _nosendsync: PhantomData })
-    }
-
     fn new(inner: ContextInner) -> Self {
         Self { inner, _nosync: PhantomData }
     }
@@ -108,6 +101,11 @@ impl NotCurrentGlContext for NotCurrentContext {
         surface_read: &Self::Surface<T>,
     ) -> Result<Self::PossiblyCurrentContext> {
         Err(self.inner.make_current_draw_read(surface_draw, surface_read).into())
+    }
+
+    fn make_current_surfaceless(self) -> Result<PossiblyCurrentContext> {
+        self.inner.make_current_surfaceless()?;
+        Ok(PossiblyCurrentContext { inner: self.inner, _nosendsync: PhantomData })
     }
 }
 
@@ -150,13 +148,6 @@ pub struct PossiblyCurrentContext {
     _nosendsync: PhantomData<*mut ()>,
 }
 
-impl PossiblyCurrentContext {
-    /// Make this context current on the calling thread.
-    pub fn make_current_surfaceless(&self) -> Result<()> {
-        self.inner.make_current_surfaceless()
-    }
-}
-
 impl PossiblyCurrentGlContext for PossiblyCurrentContext {
     type NotCurrentContext = NotCurrentContext;
     type Surface<T: SurfaceTypeTrait> = Surface<T>;
@@ -188,6 +179,10 @@ impl PossiblyCurrentGlContext for PossiblyCurrentContext {
         surface_read: &Self::Surface<T>,
     ) -> Result<()> {
         Err(self.inner.make_current_draw_read(surface_draw, surface_read).into())
+    }
+
+    fn make_current_surfaceless(&self) -> Result<()> {
+        self.inner.make_current_surfaceless()
     }
 }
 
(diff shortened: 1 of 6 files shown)


MarijnS95 (approved):

glutin/src/context.rs
```diff
@@ -522,6 +547,10 @@ impl PossiblyCurrentGlContext for PossiblyCurrentContext {
         )
     }
 
+    fn make_current_surfaceless(&self) -> Result<()> {
+        gl_api_dispatch!(self; Self(context) => context.make_current_surfaceless())
+    }
+
     fn make_not_current_in_place(&self) -> Result<()> {
         Ok(gl_api_dispatch!(self; Self(context) => context.make_not_current_in_place()?))
```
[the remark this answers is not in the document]
kchibisov: yeah, here it can be dropped.

glutin/src/api/cgl/context.rs
```diff
@@ -189,6 +180,10 @@ impl PossiblyCurrentGlContext for PossiblyCurrentContext {
     ) -> Result<()> {
         Err(self.inner.make_current_draw_read(surface_draw, surface_read).into())
```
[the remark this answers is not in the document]
kchibisov: well, it's needed, since we just build an error here with.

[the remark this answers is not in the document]
MarijnS95: Right there's no `?`, just an `ErrorKind` that this function directly returns.

glutin/src/context.rs
```diff
@@ -49,9 +49,9 @@ pub trait NotCurrentGlContext: Sealed {
     /// guaranteed to be current.
     fn treat_as_possibly_current(self) -> Self::PossiblyCurrentContext;
 
-    /// Make [`Self::Surface`] on the calling thread producing the
-    /// [`Self::PossiblyCurrentContext`] indicating that the context could
-    /// be current on the thread.
+    /// Make context current on the calling thread producing the
+    /// [`Self::PossiblyCurrentContext`]. The `surface` is used as a target for
+    /// default framebuffer.
```
MarijnS95, line 54: Maybe:
```suggestion
    /// Make context current on the calling thread and produce the
    /// [`Self::PossiblyCurrentContext`]. The `surface` is used as the
    /// default framebuffer.
```

(Otherwise `as a target for **the** default framebuffer`)

glutin/src/context.rs
```diff
@@ -89,37 +98,46 @@ pub trait PossiblyCurrentGlContext: Sealed {
...
 
-    /// Make the context not current to the current thread. If you need to
+    /// Make the context not current on the current thread. If you need to
     /// send the context to another thread, use [`Self::make_not_current`]
     /// instead.
     fn make_not_current_in_place(&self) -> Result<()>;
 
-    /// Make [`Self::Surface`] current on the calling thread.
+    /// Make context current on the calling. The `surface` is used as a target
+    /// for default framebuffer.
```
MarijnS95, line 114:
```suggestion
    /// Make context current on the calling thread. The `surface` is used as
    /// the default framebuffer.
```

glutin/src/context.rs
```diff
@@ -64,17 +64,26 @@ pub trait NotCurrentGlContext: Sealed {
...
     ///
     /// - **WGL/CGL:** not supported.
     fn make_current_draw_read<T: SurfaceTypeTrait>(
         self,
         surface_draw: &Self::Surface<T>,
         surface_read: &Self::Surface<T>,
     ) -> Result<Self::PossiblyCurrentContext>;
+
+    /// Make context current on the calling thread without a default
+    /// framebuffer producing the [`Self::PossiblyCurrentContext`].
```
MarijnS95, line 80: Just noting here, but this applies to all comments: I initially read this as "it is being made current on the calling thread which is also_ producing the PossiblyCurrentContext". Rather than that `PossiblyCurrentContext` is the _product of_ such "make current" operation. Perhaps we can reword this and do it across all similar comments, though I'm out of ideas what it should be.

```suggestion
    /// Make context current on the calling thread without a default
    /// framebuffer, producing the [`Self::PossiblyCurrentContext`].
```

Or:


```suggestion
    /// Make context current on the calling thread without a default
    /// framebuffer, and produce the [`Self::PossiblyCurrentContext`].
```

glutin/src/context.rs
kchibisov: maybe just drop the `producing` part at this point? I'd probably use the `and change it to [Self...`.

surface is not a default framebuffer, it's a target where framebuffer goes, you can change surfaces and have the same buffer iirc.

MarijnS95: Yeah, let's drop `producing` and use `change it to` or just `and return PossiblyCurrentContext`.

---

Surface isn't a framebuffer, but the word "target" isn't mentioned anywhere in https://registry.khronos.org/EGL/sdk/docs/man/html/eglMakeCurrent.xhtml nor https://www.khronos.org/opengl/wiki/Default_Framebuffer, so I'd refrain from using it.

The word `target` is often used as parameter in the binding slot for various `glBind*(GLenum target, ...)` calls however.

PR #1716, opened 2024-11-12
egl/display: workaround gbm platforms on EGL 1.4

### LINKED ISSUE #1708: `EGL_KHR_platform_gbm` (but not `EGL_MESA_platform_gbm`) client extension present on EGL 1.4
In my project, get_platform_display_ext get failed because egl client extensions mismatch。

```
egl client extensions: 
EGL_EXT_client_extensions EGL_EXT_platform_base EGL_KHR_client_get_all_proc_addresses EGL_KHR_platform_gbm EGL_KHR_platform_wayland EGL_EXT_platform_wayland
```

Screenshot_select-area_20241010115958

kchibisov:
Some EGL drivers report the KHR extension with EGL 1.4, which doesn't make any sense, however given that the constant for MESA and KHR is the same, we can check for KHR in MESA branch, but still use the Ext functions.

Fixes #1708.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 - Added `NotCurrentContext::make_current_surfaceless(self)` and
   `PossiblyCurrentContext::make_current_surfaceless(&self)` in the `Wgl`
   implementation to allow the use of surfaceless contexts with WGL.
-- Added workaround for EGL drivers reporting `EGL_KHR_platform_gbm` with EGL 1.4.
+- Added workaround for EGL drivers reporting `EGL_KHR_platform_gbm` without EGL 1.5 client.
 
 
 # Version 0.32.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 - Added `NotCurrentContext::make_current_surfaceless(self)` and
   `PossiblyCurrentContext::make_current_surfaceless(&self)` in the `Wgl`
   implementation to allow the use of surfaceless contexts with WGL.
-- Added workaround for EGL drivers reporting KHR_platform_gbm with EGL 1.4.
+- Added workaround for EGL drivers reporting `EGL_KHR_platform_gbm` with EGL 1.4.
 
 
 # Version 0.32.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Added `NotCurrentContext::make_current_surfaceless(self)` and
   `PossiblyCurrentContext::make_current_surfaceless(&self)` in the `Wgl`
   implementation to allow the use of surfaceless contexts with WGL.
+- Added workaround for EGL drivers reporting KHR_platform_gbm with EGL 1.4.
 
 
 # Version 0.32.1
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -363,8 +363,8 @@ impl Display {
                 )
             },
             RawDisplayHandle::Gbm(handle)
-                // NOTE: Some drivers report that they support KHR extension with 1.4 EGL display, so
-                // workaround here by checking the KHR gbm display as well. The MESA and KHR has
+                // NOTE: Some drivers report that they support the KHR GBM extension without EGL 1.5 client, so
+                // work around that here by checking the KHR GBM extension as well. The MESA and KHR extensions have
                 // the same constant values, thus it'll work regardless.
                 //
                 // They do require EXT during the runtime as well, so we don't change the display
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -362,7 +362,18 @@ impl Display {
                     handle.connection.map_or(egl::DEFAULT_DISPLAY as *mut _, |c| c.as_ptr()),
                 )
             },
-            RawDisplayHandle::Gbm(handle) if extensions.contains("EGL_MESA_platform_gbm") => {
+            RawDisplayHandle::Gbm(handle)
+                // NOTE: Some drivers report that they support KHR extension with 1.4 EGL display, so
+                // workaround here by checking the KHR gbm display as well. The MESA and KHR has
+                // the same constant values, thus it'll work regardless.
+                //
+                // They do require EXT during the runtime as well, so we don't change the display
+                // to Khr here.
+                //
+                // See https://github.com/rust-windowing/glutin/issues/1708.
+                if extensions.contains("EGL_MESA_platform_gbm")
+                    || extensions.contains("EGL_KHR_platform_gbm") =>
+            {
                 (egl::PLATFORM_GBM_MESA, handle.gbm_device.as_ptr())
             },
             RawDisplayHandle::Drm(_) => {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Added `NotCurrentContext::make_current_surfaceless(self)` and
   `PossiblyCurrentContext::make_current_surfaceless(&self)` in the `Wgl`
   implementation to allow the use of surfaceless contexts with WGL.
+- Added workaround for EGL drivers reporting KHR_platform_gbm with EGL 1.4.
```
MarijnS95, line 13:
```suggestion
- Added workaround for EGL drivers reporting `EGL_KHR_platform_gbm` with EGL 1.4.
```

glutin/src/api/egl/display.rs
```diff
@@ -362,7 +362,18 @@ impl Display {
                     handle.connection.map_or(egl::DEFAULT_DISPLAY as *mut _, |c| c.as_ptr()),
                 )
             },
-            RawDisplayHandle::Gbm(handle) if extensions.contains("EGL_MESA_platform_gbm") => {
+            RawDisplayHandle::Gbm(handle)
+                // NOTE: Some drivers report that they support KHR extension with 1.4 EGL display, so
+                // workaround here by checking the KHR gbm display as well. The MESA and KHR has
+                // the same constant values, thus it'll work regardless.
```
MarijnS95, line 368: Should we make it clearer that "they support KHR extension with 1.4 EGL" is the thing that's "not allowed / not supported by the spec"?

```suggestion
                // NOTE: Some drivers report that they support the KHR GBM extension without EGL 1.5 client, so
                // work around that here by checking the KHR GBM extension as well. The MESA and KHR extensions have
                // the same constant values, thus it'll work regardless.
```

Also, I think you meant to refer to the client version, not the display version? The client lacks the [`GetPlatformDisplay` function, implying EGL < 1.5](https://registry.khronos.org/EGL/sdk/docs/man/html/eglGetPlatformDisplay.xhtml), yet [EGL 1.5 is required by `EGL_KHR_platform_gbm`](https://registry.khronos.org/EGL/extensions/KHR/EGL_KHR_platform_gbm.txt).

(And to be extremely picky: the fact that the constants are the same value "doesn't really matter" when the spec doesn't define that the combination of function and constant value are valid. But the driver is already broken anyway, that's why it works)

glutin/src/api/egl/display.rs
```diff
@@ -362,7 +362,18 @@ impl Display {
...
+                //
+                // They do require EXT during the runtime as well, so we don't change the display
+                // to Khr here.
```
MarijnS95, line 371: Following https://gitlab.freedesktop.org/glvnd/libglvnd/-/issues/251#note_2553614, there's apparently no such thing as an "Ext" display or a "Khr" display, at least not related to the `GetPlatformDisplay*` kind of function that was used. Instead, this information should "solely" come from the display version returned by `eglInitialize()`? Perhaps we should skip this comment.

kchibisov: The issue is which function in surface you'll end up using, since those are different. We generally mark it as `Ext` to use the `Ext` function to create surface, since `Khr` has different prototype.

MarijnS95 (approved):
I don't like workarounds for drivers, but I guess that's why we have Vulkan nowadays :)

glutin/src/api/egl/display.rs
kchibisov: If you want to reword, go ahead.

MarijnS95: I don't specifically need to reword this as much as I first want to understand if we should base surface creation etc off of the API that the display was created with, or the version that came out of `eglInitialize()` and subsequently supported extensions _on the device_ (not on the client).

kchibisov: > if we should base surface creation etc off of the API that the display was created with, or the version that came out of eglInitialize() and subsequently supported extensions on the device (not on the client)

we use the version from `eglInitialize` already and I think we should continue doing so.

MarijnS95: Yeah we do in this fallback: https://github.com/rust-windowing/glutin/blob/1d39edd990140683a16187d0ad198933036bdf24/glutin/src/api/egl/display.rs#L524-L542

But if my hunch is correct that only the display version is what matters, I expected a more plain `match version { Version(1, 5) => Khr, ... }` etc?

Not something for this PR though.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Added `NotCurrentContext::make_current_surfaceless(self)` and
   `PossiblyCurrentContext::make_current_surfaceless(&self)` in the `Wgl`
   implementation to allow the use of surfaceless contexts with WGL.
+- Added workaround for EGL drivers reporting `EGL_KHR_platform_gbm` with EGL 1.4.
```
MarijnS95, line 13: Forgot to mention, same request as the comment itself.

```suggestion
- Added workaround for EGL drivers reporting `EGL_KHR_platform_gbm` without EGL 1.5 client.
```

MarijnS95 (approved):
Still looks great, thanks for driving this!

glutin/src/api/egl/display.rs
kchibisov: it's already correct though, since the display is `Khr`, and if it's not we fallback. You can rewrite it, but the result will be exactly same.

MarijnS95 (approved):

glutin/src/api/egl/display.rs
MarijnS95: I meant the extra wrapping logic of making it `Khr` when we used `eglGetPlatformDisplay()`, `Ext` when we used `eglGetPlatformDisplayExt()` and `Legacy` when we used `eglGetDisplay()`. If I understood the linked GitLab thread correctly, that can all solely be based on the _display_ `version` instead like the workaround we did?

kchibisov: That is true, but you still have to group later on, and you have to call some display function, so you can pre-group them into something, since you know version only after creating the display.

The point is that it's still all the same display under the hood, and if we got a missmatch we should correct and not error out, which we all do now.

MarijnS95: Sure, it's true that the used `Get*Display*()` function is defined by the same _extension or EGL 1.5 base_ that provides the `Surface` functions like `eglCreate*{Window,Pixmap}Surface*()`, as well as `eglQueryDisplayAttrib{KHR,EGL}()`. Which means I really no longer understand why https://github.com/rust-windowing/glutin/pull/1690 is failing, if surface creation functions are a feature _of the client and not of the display_ :sweat_smile:

kchibisov: @MarijnS95 all these functions are a part of the extension we end up picking. So if the display is EXT, it means that we should use the EXT functions, since Khr might not be loaded at all.

It also means that you can pick older functions though, but not _newer_ and the reason it failed because we were picking _newer_ functions. The failure was likely due to pointer not being loaded at all for khr function to create the surface, and the display stuff was from the `libglvnd`.

Like `libglvnd` does forward stuff in some cases, and if e.g. underlying driver misses function it'll fail as well.

PR #1721, opened 2025-01-10
wgl/surface: hide WglSurface

kchibisov:
--

not used anywhere in public api.


MarijnS95 (approved):
Good catch. That is nice and confusing with:

https://github.com/rust-windowing/glutin/blob/2c09d897760205b961264de00d0d401ccfb51dd9/glutin/src/surface.rs#L20-L21

PR #1726, opened 2025-01-24
Use `objc2` v0.6.1

madsmtm:
Update to `objc2` v0.6.1, and use the new `objc2-core-foundation` crate.

This requires an MSRV bump to 1.71 to get access to `extern "C-unwind"`.


kchibisov (changes requested):
Make it so your first commit is an MSRV bump and the following updates the objc stuff.

madsmtm:
Done

kchibisov (approved):

PR #1728, opened 2025-02-24
deps: avoid duplicate crates

rex4539:
Also bumped MSRV in order to fix compilation.


rex4539:
Closing due to no interest.

MarijnS95:
How did this "avoid duplicate crates"? As far as I can tell there were no conflicting version bounds. You just converted the repository to utilize workspace dependencies (and should be annotated that way) to make it easier to see and modify shared versions which could potentially avoid such a mishap _in the future_?

PR #1735, opened 2025-06-19
make `GlutinEventLoop` public

LoganDark:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - **Breaking:** Added `make_current_surfaceless(self)` for `{Possibly,Not}CurrentGlContext`.
+- Made `GlutinEventLoop` trait public
 
 # Version 0.32.3
 


kchibisov:
It was not intended to be public and I'm not sure what the use is. Though, sooner we'd be updating to new winit traits and majority of that will just go away.

kchibisov:
>The use is to trivially accept either EventLoop or ActiveEventLoop to my own functions that internally call glutin's. I figured reusing your preexisting sealed trait would be easier than making my own.

Can't the `&dyn ActiveEventLoop` be accepted? It's also a trait.

LoganDark:
> Can't the `&dyn ActiveEventLoop` be accepted? It's also a trait.

Is it?

image

It *seems* to exist [in `winit-core`](https://github.com/rust-windowing/winit/blob/50c0180af309d297562703d1221328f8d8351011/winit-core/src/event_loop/mod.rs#L18), but `winit-core` and `winit_core` are not on crates.io and the trait is not re-exported by `winit`. So I'm not sure how to access it :/

PR #1740, opened 2025-12-27
fix: CGL: try to init pixel format without explicit profile before falling back to profile selection

kalzoo:
- [ ] ~~Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior~~
- [ ] ~~Created or updated an example program if it would help users understand this functionality~~

*Note*: I am not experienced with OpenGL. This contribution fixes a bug "on my machine" but I can't say for certain that it is correct in general.

----

# Problem

On macOS, the current implementation always adds `NSOpenGLPFAOpenGLProfile` when creating pixel formats, then iterates through profiles (4.1 Core, 3.2 Core, Legacy) until one succeeds.

This differs from GLFW's behavior. When GLFW receives no explicit `GLFW_CONTEXT_VERSION_*` hints, it does not add `NSOpenGLPFAOpenGLProfile` at all ([see nsgl_context.m here](https://github.com/glfw/glfw/blob/dbadda26835ec5089ef922e6c290bcf58cf12056/src/nsgl_context.m#L221-L245) and [here](https://github.com/glfw/glfw/blob/dbadda26835ec5089ef922e6c290bcf58cf12056/src/nsgl_context.m#L310) )

The difference matters because:

* With explicit profile: macOS returns a strict context matching that profile
* Without profile attribute: macOS selects a "best compatible" context that may expose both legacy fixed-function GL features AND modern extensions like ARB_framebuffer_object

This causes compatibility issues for applications that need both fixed-function OpenGL (only available in Legacy profile) and ARB_framebuffer_object (not exposed as an extension in strict Legacy mode, only in 3.0+ or unspecified contexts).

In my case, that was `mujoco-rs`, [issue here](https://github.com/davidhozic/mujoco-rs/issues/100).

(Red herring - same error message, but not caused by or using `glutin`: https://github.com/google-deepmind/mujoco/issues/1778 )

# Solution

So, I changed that section to try creating the pixel format without any `NSOpenGLPFAOpenGLProfile` first. If that fails, fall back to the existing/previous explicit profile iteration.

# Testing

Tested on macOS with applications using MuJoCo physics engine rendering, which requires both fixed-function GL and ARB_framebuffer_object. Previously failed with `OpenGL ARB_framebuffer_object required` error; now works correctly. Also, `cargo test` passed. No other testing performed.

### DIFF
--- a/glutin/src/api/cgl/config.rs
+++ b/glutin/src/api/cgl/config.rs
@@ -95,32 +95,40 @@ impl Display {
             attrs.push(NSOpenGLPFAStereo);
         }
 
-        attrs.push(NSOpenGLPFAOpenGLProfile);
+        // First, try WITHOUT NSOpenGLPFAOpenGLProfile to match GLFW's default
+        // behavior when no version hint is set.
+        attrs.push(0); // Terminate attrs
 
-        // Stash profile pos for latter insert.
-        let profile_attr_pos = attrs.len();
-        // Add place holder for the GL profile.
-        attrs.push(NSOpenGLProfileVersion4_1Core);
+        let raw = unsafe {
+            NSOpenGLPixelFormat::initWithAttributes(
+                <NSOpenGLPixelFormat as AllocAnyThread>::alloc(),
+                NonNull::new(attrs.as_ptr().cast_mut()).unwrap(),
+            )
+        }
+        .or_else(|| {
+            // If that fails, fall back to trying explicit profiles.
+            // Remove the terminating 0 and add profile attribute.
+            attrs.pop();
+            attrs.push(NSOpenGLPFAOpenGLProfile);
+            let profile_attr_pos = attrs.len();
+            attrs.push(NSOpenGLProfileVersion4_1Core);
+            attrs.push(0);
 
-        // Terminate attrs with zero.
-        attrs.push(0);
-
-        // Automatically pick the latest profile.
-        let raw = [
-            NSOpenGLProfileVersion4_1Core,
-            NSOpenGLProfileVersion3_2Core,
-            NSOpenGLProfileVersionLegacy,
-        ]
-        .into_iter()
-        .find_map(|profile| {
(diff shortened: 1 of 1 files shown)


glutin/src/api/cgl/config.rs
```diff
@@ -95,32 +95,40 @@ impl Display {
...
-        let profile_attr_pos = attrs.len();
-        // Add place holder for the GL profile.
-        attrs.push(NSOpenGLProfileVersion4_1Core);
-
-        // Terminate attrs with zero.
-        attrs.push(0);
-
-        // Automatically pick the latest profile.
-        let raw = [
-            NSOpenGLProfileVersion4_1Core,
```
kchibisov, line 110: Can you make it into e.g. `[(0, 0), (NSOpenGLProfileVersion4_1Core, NSOpenGLPFAOpenGLProfile), ...]` so we don't have this back and forth. It doesn't really matter how many elements we have as long as its terminated with `len`.

kchibisov (changes requested):
Need changelog entry for that.

glutin/src/api/cgl/config.rs
davidhozic: Not sure if the PR author is subscribed to the PR thread. Perhaps you can modify it manually @kchibisov? This would be a nice to have fix since it completely blocks macOS visualization from working in my crate.

davidhozic:
Any news on this @kchibisov ?

kchibisov:
you can apply suggested changes, test and open a PR? I don't have macOS to test, so leave it to someone affected with said problem

davidhozic:
> you can apply suggested changes, test and open a PR? I don't have macOS to test, so leave it to someone affected with said problem

I'll apply some changes and test them tomorrow. Looking at this PR again, it seems to kinda break some stuff for existing users as macOS. Specifically, it will select an older OpenGL version.

I'll try to add some sorts of configuration, with the default staying the same.

PR #1741, opened 2026-01-07
egl: support grabbing current context and display

Supreeeme:

Open questions:
- I'm not sure about the error handling when grabbing the current context.
- I'm also not sure about the error handling when parsing the Display version, though it seems like it'd be a driver bug for that to fail.

### DIFF
--- a/glutin/src/api/egl/display.rs
+++ b/glutin/src/api/egl/display.rs
@@ -179,6 +179,41 @@ impl Display {
         Self::initialize_display(egl, platform_display, None)
     }
 
+    /// Loads the currently active EGL display, if present
+    /// Returns a `ManuallyDrop` display to avoid terminating the display
+    /// by default, since you may not be in control of it.
+    pub fn current() -> Result<std::mem::ManuallyDrop<Self>> {
+        let egl = match EGL.as_ref() {
+            Some(egl) => egl,
+            None => return Err(ErrorKind::NotFound.into()),
+        };
+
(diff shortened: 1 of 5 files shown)


glutin/src/api/egl/display.rs
```diff
@@ -179,6 +179,41 @@ impl Display {
...
+
+        let display_extensions = get_extensions(egl, display);
+        let features = Self::extract_display_features(&display_extensions, version);
+        let inner = Arc::new(DisplayInner {
+            egl,
+            raw: EglDisplay::Ext(display),
+            _native_display: None,
+            version,
+            display_extensions,
+            features,
```
kchibisov, line 212: This should follow display creation logic to some extent, I think.

glutin/src/context.rs
```diff
@@ -149,6 +149,12 @@ pub trait AsRawContext {
     fn raw_context(&self) -> RawContext;
 }
 
+impl AsRawContext for RawContext {
+    fn raw_context(&self) -> RawContext {
+        *self
+    }
+}
```
kchibisov, line 156: It was missing intentionally.

kchibisov (changes requested):
In general, if you want to add raw contexts we need to have `borrow` and `owned` ways of taking it, and for dropping part, we must just track on creation + in `drop` how we got the object, in case it was borrowed, we should just ignore dropping. 

I'm fine with having it just for `EGL`.

PR #1746, opened 2026-04-18
fix(wgl): handle unreliable WGL_TRANSPARENT_ARB by returning None

### LINKED ISSUE #1717: WGL backend supports_transparency is false
Hello!

I attempted to create a transparent window using egui, but it was not transparent. After investigating further, I discovered that the issue stems from the finalize_window (https://github.com/rust-windowing/glutin/blob/master/glutin-winit/src/lib.rs#L194) function. The supports_transparency method returns false for me. Upon additional research, it appears that the wgl_extra::TRANSPARENT_ARB attribute is consistently false on my setup, which seems odd given that I have a modern discrete GPU with up-to-date drivers. And this attribute is false on windows, compiling and running same code under  WSL ubuntu makes this attribute True.

Moreover, the Glutin window example is transparent since the finalize_window function is never called for it. I'm wondering if it's safe to remove this check entirely and proceed without it. According to the Winit code, it creates transparent windows without performing any compatibility checks. So, I'm confused about why Glutin modifies this behavior.

Could someone clarify this for me or suggest a solution?

ManiaciaChao:

Fixes #1717 

### Why

`WGL_TRANSPARENT_ARB` is unreliable and can return false negatives, leading to incorrect false results.
e.g. https://github.com/GPUOpen-Drivers/AMD-Gfx-Drivers/issues/29

### How

Return `None` instead of `Some(false)` when transparency can’t be determined reliably, providing consistent semantics and avoiding misleading results.

### DIFF
--- a/glutin/src/api/wgl/config.rs
+++ b/glutin/src/api/wgl/config.rs
@@ -448,9 +448,10 @@ impl GlConfig for Config {
             // WGL_TRANSPARENT_ARB is not reliable and can sometimes produce false
             // negatives. return None to provide consistent semantics, since we
             // don't have any better way to detect
-            let transparent =
-                unsafe { self.raw_attribute(wgl_extra::TRANSPARENT_ARB as c_int) != 0 };
-            transparent.then_some(true)
+            match unsafe { self.raw_attribute(wgl_extra::TRANSPARENT_ARB as c_int) } {
+                1 => Some(true),
+                _ => None,
+            }
         }
     }
 
--- a/glutin/src/api/wgl/config.rs
+++ b/glutin/src/api/wgl/config.rs
@@ -445,7 +445,12 @@ impl GlConfig for Config {
         if self.inner.descriptor.as_ref().is_some() {
             None
         } else {
-            unsafe { Some(self.raw_attribute(wgl_extra::TRANSPARENT_ARB as c_int) != 0) }
+            // WGL_TRANSPARENT_ARB is not reliable and can sometimes produce false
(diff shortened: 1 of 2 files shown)


glutin/src/api/wgl/config.rs
```diff
@@ -445,7 +445,12 @@ impl GlConfig for Config {
         if self.inner.descriptor.as_ref().is_some() {
             None
         } else {
-            unsafe { Some(self.raw_attribute(wgl_extra::TRANSPARENT_ARB as c_int) != 0) }
+            // WGL_TRANSPARENT_ARB is not reliable and can sometimes produce false
+            // negatives. return None to provide consistent semantics, since we
+            // don't have any better way to detect
+            let transparent =
+                unsafe { self.raw_attribute(wgl_extra::TRANSPARENT_ARB as c_int) != 0 };
+            transparent.then_some(true)
```
kchibisov: Use `filter` and make `< 0` into None then, and then do a `map?

kchibisov: if the goal was to make only `1` work, you should explicitly do so by using `== 1` and returning `None` otherwise.

kchibisov (changes requested):

glutin/src/api/wgl/config.rs
ManiaciaChao: Good catch! I'll revise this.

ManiaciaChao:
Looks like the CI is being blocked by issues in existing code outside of this PR.

kchibisov (approved):

PR #1747, opened 2026-06-07
Allow skipping pixel-format OpenGL profile attributes

davidhozic:
- [ ] ~~Created or updated an example program if it would help users understand this functionality~~

This is essentially the same as #1740 , with the exception of allowing configurable behavior and in being more compact. By default, nothing is changed to the previous behavior of `glutin` on macOS (CGL) except the added no-profile fallback when all prior profile selections fail.

Essentially, it addresses the problem of certain OpenGL extensions not working on macOS (`ARB_framebuffer_object`).

All the formatting errors are pre-existing.

### DIFF
--- a/glutin/src/config.rs
+++ b/glutin/src/config.rs
@@ -266,9 +266,9 @@ impl ConfigTemplateBuilder {
     /// Request that the `NSOpenGLProfile` selection is skipped on `CGL/macOS`.
     /// This allows better compatibility with older openGL profiles
     /// that rely on certain OpenGL extensions (from newer versions).
-    /// 
+    ///
     /// # Api-specific
-    /// 
+    ///
     /// Only supported with `CGL` (`macOS`)
     #[inline]
     pub fn skip_cgl_profile(mut self, skip: bool) -> Self {
@@ -377,7 +377,7 @@ impl Default for ConfigTemplate {
 
             api: None,
 
-            skip_cgl_profile: false
+            skip_cgl_profile: false,
         }
     }
 }
--- a/glutin/src/config.rs
+++ b/glutin/src/config.rs
@@ -263,6 +263,19 @@ impl ConfigTemplateBuilder {
         self
     }
 
+    /// Request that the `NSOpenGLProfile` selection is skipped on `CGL/macOS`.
+    /// This allows better compatibility with older openGL profiles
+    /// that rely on certain OpenGL extensions (from newer versions).
+    /// 
+    /// # Api-specific
+    /// 
+    /// Only supported with `CGL` (`macOS`)
+    #[inline]
+    pub fn skip_cgl_profile(mut self, skip: bool) -> Self {
+        self.template.skip_cgl_profile = skip;
+        self
+    }
+
     /// Build the template to match the configs against.
     #[must_use]
     pub fn build(self) -> ConfigTemplate {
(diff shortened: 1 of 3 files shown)


davidhozic:
@kchibisov This makes minimal changes to the behavor and preserves the original behavior of when nothing is specified.

Maybe just a thought for the future. From what I can tell, there is no reason for the profile selection not being inside the `Display::create_context` method; before the context creation:

https://github.com/rust-windowing/glutin/blob/e2f87adba6ca6dfdf2aa9791204243114d6f6f02/glutin/src/api/cgl/context.rs#L53

Is there a reason that it is currently inside the `find_configs` method? On other platforms, the profile selection is set via `ContextAttributesBuilder::with_context_api` and `ContextAttributesBuilder::with_profile`.

davidhozic:
I tested it and it now works on macOS, in addition to the existing unchanged platforms.

kchibisov:
I think there was some macOS specific reason for that, but I can not recall it. Moving it and align with other platforms would have been better, in my opinion.

Maybe something related to send/sync or some OpenGL _quirks_ macOS have given that apple considers it legacy. Would suggest to just browse git history, could be a reason for, since I think it was like that with even original glutin design.

davidhozic:
> I think there was some macOS specific reason for that, but I can not recall it. Moving it and align with other platforms would have been better, in my opinion.
> 
> Maybe something related to send/sync or some OpenGL _quirks_ macOS have given that apple considers it legacy. Would suggest to just browse git history, could be a reason for, since I think it was like that with even original glutin design.

From what I can tell it should work. The exception will be that errors will not be caught in find_configs. Perhaps it is best to leave this configuration inside find_configs for now and change it sometime in the future if it turns problematic.

glutin/src/config.rs
```diff
@@ -263,6 +263,19 @@ impl ConfigTemplateBuilder {
...
 
+    /// Request that the `NSOpenGLProfile` selection is skipped on `CGL/macOS`.
+    /// This allows better compatibility with older openGL profiles
+    /// that rely on certain OpenGL extensions (from newer versions).
+    ///
+    /// # Api-specific
+    ///
+    /// Only supported with `CGL` (`macOS`)
+    #[inline]
+    pub fn skip_cgl_profile(mut self, skip: bool) -> Self {
```
kchibisov, line 274: Can we infer it from `with_api`/`Api` instead?

kchibisov (changes requested):

glutin/src/config.rs
davidhozic: By skipping profile selection when `ConfigTemplate::api` is `None`? We can do that if you're okay with profile selection being skipped by default.

kchibisov: I think so, though, it's technically a behavior change, but probably it's fine? What it picks on recent enough macs?

davidhozic: It will switch to the legacy OpenGL (OpenGL 2.1) + extensions. ~~I'm fine with making it check the `ConfigTemplate::api`, however it doesn't feel very intuitive that you need to select OpenGL explicitly in the `Api`~~ Actually it is fine; users will just have to assume they can't rely on OpenGL being selected without `with_api` on macOS, and it will look to them like something else may be selected.

davidhozic:
This should be fine now @kchibisov . Sorry it took so long, I sort of forgot about it.

davidhozic:
I've just applied the `cargo fmt` formatting to the files I've modified. The failing CI about formatting is related to the master branch

PR #1757, opened 2026-07-13
Use `wayland-egl` crate instead of `wayland-sys`

ids1024:
This is a draft since it is currently based on the upcoming release of wayland-rs.


kchibisov:
But do we need to? I think the goal was to use less dependencies.

ids1024:
It is an additional dependency, but it's a tiny crate that with `default-features = false` doesn't depend on anything other than wayland-sys.

One of the changes in the upcoming release is to make wayland-sys a private dependency of wayland-backend and wayland-egl. Ideally nothing other than those two crates should depend on it directly.

